PR #873, opened 2014-09-05
[WIP] Change all-fields pf to pf2 (instead of p2 which does not exist).

jkeck:
Don't merge yet.  This is a change from current production and what's been tested.  It's possible that we were falling back on the correct behavior previously, so we should evaluate if this change is unexpectedly breaking anything else.


thatbudakguy:
closing, since it's currently `pf2`.

PR #999, opened 2015-01-08
fixes the missing browse nearby mystery

mejackreed:
Removed no longer necessary `collapse` class

This commit introduced `visibility: hidden` to `collapse` class: https://github.com/twbs/bootstrap/commit/6cfd176ac3d0eedc5b9b50fd88bf0774af0f1222

### DIFF
--- a/app/views/catalog/record/_callnumber_browse.html.erb
+++ b/app/views/catalog/record/_callnumber_browse.html.erb
@@ -20,7 +20,7 @@
   </div>
 
   <% document.holdings.browsable_callnumbers.each_with_index do |callnumber, index| %>
-    <div id="callnumber-<%=index%>" class="embed-callnumber-browse-container">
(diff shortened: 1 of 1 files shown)


jkeck:
@mejackreed thanks for this.  It looks like it unfortunately breaks in a different way for records that have multiple call numbers (see 1517101 as an example).

#### Initial page load
initial

#### After clicking first call number
click-first-callnumber

jkeck:
### 1517101 Before
1517101-before

### 1517101 After
1517101-after

mejackreed:
I'm not sure if this is related to this bug or another, but on default when the page loads the browse bar is not appearing centered on the correct itme.

:+1: for merging, but shall we add this as another ticket?

jkeck:
Yes, please add that as another ticket.  Good catch.

mejackreed:
Ok added in #1000

PR #1003, opened 2015-01-29
Temporarily declare explicit z-index on .modal-dialog.

jkeck:


jchristo4:
@jkeck Tested it locally with `cite`, `Topic` facets and `Find it` modal overlays. Looks good  :+1:

PR #1005, opened 2015-02-04
Add Whenever/Capistrano integration so our cron jobs are written.

jkeck:
Also making schedule.rb a shared file so that different servers
can have different cron schedules w/o using rails_config.


mejackreed:
:+1:

PR #1010, opened 2015-03-12
Don't try to link to a barcoded LANE item when there are only MHLDS.

jkeck:
This is currently causing an error for any LANE record that has no barcoded items and only has MHLD data.


mejackreed:
:+1:

PR #1016, opened 2015-03-26
Only set the thumbnail URL if it's not already set.

### LINKED ISSUE #963: Set precedence to ISBN when retrieving Google book covers
Refer SW-1306

jkeck:
Closes #963 

@jchristo4 does this seem reasonable to you?  I think Google will return the numbers in the order we send them in and since we'll only try to set the image for the first match we have in the response this should address the issue where OCLC/LCCN matches are being used when an ISBN is returned first.

### DIFF
--- a/app/assets/javascripts/jquery.plug-google-content.js
+++ b/app/assets/javascripts/jquery.plug-google-content.js
@@ -38,7 +38,6 @@
         $.ajax({
           type: 'GET',
           url: batchBooksApiUrl,
-          async: false,
           contentType: "application/json",
           dataType: 'jsonp',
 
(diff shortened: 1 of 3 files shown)


jchristo4:
@jkeck Testing PR with objects having ISBN & OCLC/LCCN

jkeck:
Thanks, I'll fix up the tests if you sign-off on the approach.

app/assets/javascripts/jquery.plug-google-content.js
```diff
@@ -38,7 +38,7 @@
         $.ajax({
           type: 'GET',
           url: batchBooksApiUrl,
-          async: false,
+          async: true,
```
mejackreed: This setting is true by default, any reason we are calling this out here?

jkeck: Not really, @jchristo4 added that change (although I'm not sure he's available to explain why).

jkeck: We can see what happens if we remove it (assuming it doesn't have unintended affects).

mejackreed: Not a big deal, just if we can cut code... :haircut:

jkeck:
@mejackreed updated and squashed.

PR #1021, opened 2015-05-29
Add temporary LIBRARY_INSTRUCTIONS for ART.

jkeck:
art-lib-instructions

### DIFF
--- a/lib/constants.rb
+++ b/lib/constants.rb
@@ -1090,6 +1090,10 @@ module Constants
    }
    BROWSABLE_CALLNUMBERS = ["LC", "DEWEY", "ALPHANUM"]
    LIBRARY_INSTRUCTIONS = {
+     'ART' => {
+       heading: 'Library closed June25 - August23',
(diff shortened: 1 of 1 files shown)


lib/constants.rb
```diff
@@ -1090,6 +1090,10 @@ module Constants
    }
    BROWSABLE_CALLNUMBERS = ["LC", "DEWEY", "ALPHANUM"]
    LIBRARY_INSTRUCTIONS = {
+     'ART' => {
+       heading: 'Library closed June25-August23',
```
mejackreed: Should there be spaces here?

```
heading: 'Library closed June 25 - August 23',
```

jkeck: @jvine?  I had the same thought, but just followed the image you sent me.

jvine: Original text is what PB provided, but yeah, it's better w spaces. Let's do that.

jkeck: done

PR #1045, opened 2015-07-22
Add MEDIA-MTXT is a library.

jkeck:
Also fixes up the constants file to match fix quotes and indentation.

media-mtxt


jkeck:
:shipit:

PR #1046, opened 2015-07-30
Update email regular expression and guard against duplicate matches

cbeer:


cbeer:
Stashing this here so we don't forget, but it might benefit from tests or rethinking the method..

jkeck:
Do you think it would be worth simply adding red-carpet or some other markdown parser that we can limit to just linking behavior?

cbeer:
:+1: to making someone else do the hard work

PR #1055, opened 2015-09-08
Add a pagination threshold config setting and don't allow search…

jkeck:
…results that surpass the threshold.

threshold-message


mejackreed:
Did we also consider a robots.txt change too?

jkeck:
The particular case that we're trying to solve for here is not a robots.txt obeying crawler. I agree that we should probably do something in the robots.txt as well but we should probably think a bit more about our strategy there, and this is a stop-gap measure to stop any scripts from doing this in the immediate future.

PR #1068, opened 2015-10-22
App work to facilitate integrate new requests application

jkeck:
`¯\_(ツ)_/¯`


config/initializers/devise.rb
```diff
@@ -1,3 +1,18 @@
...
+  config.env_key = lambda do |env|
+    if env['REMOTE_USER']
+      "#{env['REMOTE_USER']}@stanford.edu"
+    elsif Rails.env.development? && ENV['REMOTE_USER']
+      ENV['REMOTE_USER']
+    end
+  end
+  config.auto_create = true
+  config.auto_update = true
+  config.attribute_map = { webauth_groups: 'WEBAUTH_LDAPPRIVGROUP' }
```
cbeer: Are these machines blessed for directory access, or is this just here optimistically?

cbeer:
`¯\_(ツ)_/¯`. Looks good enough to me.

config/initializers/devise.rb
jkeck: Ah, you know what, I think this added for Triannon work (where we were on a blessed server).

I can certainly take it out, but it's just no-op if we're not blessed, so maybe we could :shipit:.

PR #1073, opened 2015-10-27
Add feature to handle many managed PURLs in one record

jkeck:
@mejackreed I've added some tests here.  I think it's ready to :ship: for deploy this week.

many-purls


mejackreed:
Looks good. :shipit: !

PR #1074, opened 2015-10-28
Render digital bookplate information when present

### LINKED ISSUE #1062: Bookplates
Assuming that SearchWorks will have the following information in the index:
* bookplate title
* druid
* filename

## On the record view

The Bookplate section appears as the last section in the "record-sections" chunk of metadata. It does not need to appear in the record-side-nav.

Most bookplated items will have one bookplate, but more are possible.

Each bookplate is represented by a media block that includes:
* bookplate image (from the filename)
* bookplate title
  * links to masthead search result of all items w this bookplate
* external link to the Spotlight digital bookplates exhibit

screen shot 2015-10-09 at 2 12 07 pm

The blocks are two-columns on wide screen, collapsing to one column on mobile. Like col-sm-6. 

screen shot 2015-10-09 at 2 12 32 pm

screen shot 2015-10-09 at 12 29 13 pm

## On the masthead

When the user clicks on a bookplate name on the record view, the search result has a masthead. The Bookplate facet is not otherwise exposed.

* bookplate image
* masthead title = bookplate title
* generic text
* external link to the Spotlight digital bookplates exhibit
* breadcrumb shows facet value

screen shot 2015-10-09 at 4 51 39 pm

Formatting fun:
* The image and text are in two columns - so the text should NOT wrap around the image when the screen narrows. 
* On small devices, it would be great if both the bookplate title and the generic text would truncate with ellipsis at 3 lines 

screen shot 2015-10-09 at 5 02 34 pm

## Image sizing

The image dimensions are really variable. Is it possible to constrain differently for portrait vs. landscape? Like 100px on the short edge?

jkeck:
Closes #1062 

There will probably be some styling updates that we'll want to make. Since we don't have a live index we can't deploy this anywhere, but @jvine I'm more than happy to show you what it looks like in different screen resolutions, etc using the fixture data.

@aeschylus Please leave any comments w/ questions you have about these changes.  We can also discuss over chat or in person next week depending on your availability.

### Record View
bookplate-record

### Search Results
bookplate-search


app/models/concerns/stacks_images.rb
```diff
@@ -0,0 +1,25 @@
...
+#  Mixin to add methods for crafting Stacks image urls
+###
+module StacksImages
+  ##
+  # Generates a Stacks image url using given arguments
+  # @param [String] druid
+  # @param [String] image_id
+  # @param [Symbol] size
+  # @return [String]
+  def craft_image_url(druid, image_id, size)
```
cbeer: We should update this to use IIIF sometime. If you visit the generated urls, it'll redirect you to the IIIF equivalents.

jkeck: Yeah, probably.  Just curious, would the correct thing be here (for the 400px long-edge) to construct the druid + image id + `/full/!400,400/0/default.jpg` for this kind of thing?  Or is there another url pattern we should be using?

cbeer: :+1: perfect.

jkeck: :+1: good call.  I've made that change.

PR #1078, opened 2015-10-30
Add feed icon to atom feed for results sorted by date_cataloged

### LINKED ISSUE #1070: RSS feed
Add RSS icon in front of the search results count - but not part of the h2.

Something like this - i.e. icon is hidden from screen reader and text is invisible on screen.

`<a href="..."><i class="..." aria-hidden="true"></i><span class="sr-only">RSS feed for this result</span></a>`

minimum - rss on search results

Icon is Streamline rss-feed-box - needs to be added to sul_styles.
rss-feed-box

Color is $color-pantone-144 #e98300;

Bonus: add date-cataloged as an option to the sort dropdown.

sort by acquisition date

jkeck:
Closes #1070 

rss-feed

### DIFF
--- a/app/assets/stylesheets/modules/results-documents.css.scss
+++ b/app/assets/stylesheets/modules/results-documents.css.scss
@@ -84,9 +84,20 @@
 .search_num_of_results {
   text-align: right;
 
-  h2 {
+  .results-heading {
     border-bottom: 1px solid $sul-h2-border-color;
     color: $sul-h2-font-color;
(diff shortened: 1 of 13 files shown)


jkeck:
See #1079 re feed-con.

jkeck:
@jvine is `sort=new-to-libs` okay as a url param? (or do you particularly care?)

jvine:
@jkeck Fine w me.

PR #1080, opened 2015-10-30
Add Government documents access point and featured resource block

jkeck:
(requires #1074)

### Home Page
feature-block

### Search Results
masthead

**Note:** There is a config option which sets the block on the home page off by default. This will allow us to ship this code to production whenever we want and turn it on once the data is present.


app/views/catalog/mastheads/_government_documents.html.erb
```diff
@@ -0,0 +1,14 @@
+<div id="masthead-container">
+  <div id="masthead" class="government-documents-masthead">
+    <h1>Government documents</h1>
+    <div>
+      <p>Stanford is a depository library for state, federal, UN, and EU government documents.</p>
+      <p>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
```
cbeer: Should we get real language before merging this, or is there a ticket to make sure we do this before deploying?

jkeck: Yeah, probably. The link to it on the home page if "off" by default so if we did push this to prod there would be no way to get here to see this text. 

@jvine do you have text in mind for this?  I could also just kill the `<p>` and add it later if needed (although it looks a little silly w/o it).

jvine: @jkeck I don't have text yet. If you haven't already removed it, maybe leave the lorem ipsum in place for now.

PR #1149, opened 2016-01-21
fixed buggy sidebar by deleteing javascript and using CSS instead of …

aeschylus:
…expand/collapse behaviour


aeschylus:
I ended up doing this by just deleting code and using CSS, so there's less to test! There may be more work to do on the sidebar, but this fixes the bug and adds a nice hover style in addition to the selection style.

jkeck:
:+1: LGTM! Well done.  Just waiting for travis and will :shipit:

PR #1154, opened 2016-01-27
Address a few issues around GBS book covers

### LINKED ISSUE #720: Strange browser-specific cover-display behaviour
This record:
http://searchworks-new.stanford.edu/view/2018222

In Firefox, it shows the wrong cover.
screen shot 2014-08-13 at 12 53 32 pm

In Chrome, it shows the wrong cover briefly, then updates it with the correct cover.
wrong-cover

jkeck:
Closes #720

1. Normalize standard numbers to remove whitespace and useless characters before sending to into the Google API (this was causing some LCCNs to be viewed as ISBNs).
2. Instead of looping through the indeterminately sorted JSON response we are looping through the images, and selecting the piece from the JSON response based on the best key available.

### DIFF
--- a/app/assets/javascripts/jquery.plug-google-content.js
+++ b/app/assets/javascripts/jquery.plug-google-content.js
@@ -159,8 +159,9 @@
         bibkeys += [isbn, oclc, lccn].join(',') + ',';
       });
 
-      bibkeys = bibkeys.replace(/,,/, '');
-      bibkeys = bibkeys.replace(/,$/, '');
+      bibkeys = bibkeys.replace(/,{2,}/, ','); // Replace 2 or more commas with a single
+      bibkeys = bibkeys.replace(/^,/, ''); // Remove leading comma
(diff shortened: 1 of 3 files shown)


shelleydoljack:
When searching by ckey, the book cover in the search result differs from the record page: http://searchworks-test.stanford.edu/covers/?utf8=%E2%9C%93&search_field=search&q=989611. The record page displays the right cover. But when searching by title, "Round the world in a motor-car", the search results page and the record page display the right cover: http://searchworks-test.stanford.edu/covers/?utf8=%E2%9C%93&search_field=search&q=Round+the+world+in+a+motor-car.

jkeck:
Addressed issue noted by @shelleydoljack with ckey 989611 in certain search results.

shelleydoljack:
Wrong book cover is retrieved for ckey 1734195. I believe this is caused by the LCCN having a space, which we strip when passing to Google. Shouldn't we URL-escape spaces in LCCNs instead of stripping them?

jkeck:
Should be good to merge once CI passes.

PR #1159, opened 2016-01-27
Update to Rails 4.2

cbeer:
Includes #1157 and #1158.

Fixes #997, too.


jkeck:
A few things come to mind WRT the deprecation warning coming from Roadie.

Either we just try to use `*_url` all the time in our metadata.  It may not be the worst thing however we may forget at some point when adding a new piece of metadata (which may just have to be a risk that we'll take).  I would be happy to just make this a ticket and :ship: this PR.

I tried to rebase this but it got a little nasty.  I can try to take a crack at it again but might not get around to it before I head off for vacation.

cbeer:
Rebased.

PR #1163, opened 2016-01-29
[WIP?] Closes #1098, adds sidenavminimap to search results page (but not on …

### LINKED ISSUE #1098: Add sidebar top/bottom nav buttons to search results
Same as on the record view, but only top/bottom links. They should be in the DOM somewhere around the pagination.

screen shot 2015-11-06 at 11 49 37 am

aeschylus:
…the 'zero results' page), changes variable names and selectors to work on arbitrary pages in Searchworks.

It would also be good to change filenames for the relevant views/components. Not sure if that should go in a resparate PR or not - also, it would be good to test some of this (for instance, the zero results page), but I could use some navigation help with that.


aeschylus:
As for changing filenames (which I know can be treacherous in rails, because magic), I would need to update the application.js manifest, the record_side_nav.js file, and the _record_side_nav partial, as well as the two files that call this partial. Are there any other places that would be affected? (Trying this out now that it's committed.)

aeschylus:
Hm, all checks have failed. Withdrawing. Also, noticed I left some logging in.

PR #1164, opened 2016-01-29
Issue#1098 abstracted menu

aeschylus:


app/views/catalog/_index.html.erb
```diff
@@ -63,7 +63,10 @@
       <%= render 'additional_results_block' %>
     <%- end %>
 
-    <%= render 'results_pagination' %>
+    <%- if !@response.empty? %>
```
jkeck: It might be more readable to use Ruby's `unless` instead of `if !{Boolean}`.

(also, and I know this is above, but just an FYI the `<%-` can be written simply as `<%`.  The `-` is a hold-over from previous versions of ERB and I try to remove whenever possible)

app/views/catalog/_index.html.erb
```diff
@@ -63,7 +63,10 @@
...
+      <%= render 'shared/side_nav_minimap' %>
```
jkeck: Is this particular position in the DOM of this element important for keyboard accessibility purposes?  Might be a question for @jvine  (or perhaps you all have already worked this out).

jkeck:
:+1: Looks good!

One minor comment inline. Also, would it be possible to squash this into fewer commits?

aeschylus:
Sure, I can take care of the comment and squash.

app/views/catalog/_index.html.erb
aeschylus: Cool, thanks for the tip. I'll make these changes.

jkeck:
Closing as was re-submitted as ##1181

aeschylus:
:+1:

PR #1167, opened 2016-01-30
Fix wrapping of library names in access panel

jkeck:
## Before
before

## After
after

### DIFF
--- a/app/assets/stylesheets/modules/access-panel.css.scss
+++ b/app/assets/stylesheets/modules/access-panel.css.scss
@@ -6,19 +6,23 @@
   font-size: 1.125em;
   font-weight: 400;
 }
+
 .access-panel-heading {
   &.library-location-heading h3 {
-    border-bottom: 1px dotted white;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/access-panel.css.scss
```diff
@@ -6,19 +6,23 @@
...
 }
+
 .access-panel-heading {
   &.library-location-heading h3 {
-    border-bottom: 1px dotted white;
-    display: inline-block; /* so the border won't extend past the text */
-    margin-right: 10px;
-    &:hover,&:active {
-      border-bottom: 1px solid white;
+    border-bottom: 1px dotted $white;
```
aeschylus: This is a way. When you need the border to be included in the specified width, you can adjust the box model to `border-box`: https://css-tricks.com/box-sizing/, https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

PR #1169, opened 2016-02-01
style find-it link as a button; link location text to same url (excep…

peetucket:
…t when not available or when an external library)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ group :development, :test do
   # We use database cleaner to empty out the database between tests (see spec_helper for usage)
   gem 'database_cleaner'
   gem 'poltergeist', '>= 1.8.1'
+  gem 'launchy' # useful for debugging rspec/capybara integration tests -- put "save_and_open_page" in your test to debug
 end
 
 group :sqlite do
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -59,6 +59,7 @@ group :development, :test do
   # We use database cleaner to empty out the database between tests (see spec_helper for usage)
   gem 'database_cleaner'
   gem 'poltergeist', '>= 1.8.1'
+  gem 'launchy' # useful for debugging rspec/capybara integration tests -- put "save_and_open_page" in your test to debug
```
jkeck: :+1: 
Can we separate the addition of this dependency into a separate commit?  That way if one or the other needs to be reverted it won't inadvertently remove something else.

app/assets/stylesheets/searchworks-mixins.css.scss
```diff
@@ -95,13 +95,18 @@
 }
 
 @mixin stackmap-find-it {
-  margin-left: 10px;
-
+  background-color: #007cbc;
+  border-color: #006ba3;
+  margin-right: 5px;
+  color: #ffffff;
```
jkeck: Can we move these colors into variables (or use existing ones when available)?  I believe that is the preferred practice in the scss world and will help us a lot when we implement SULStyles.

(FWIW `#ffffff` is already defined as `$white`)

jkeck:
:+1: LGTM! Few super-minor comments.

Gemfile
peetucket: :+1: sounds good

PR #1170, opened 2016-02-02
Only show bound-with note in holdings when 590$c is present

### LINKED ISSUE #1092: "Bound with" note should appear only if 590 has subfield c.
SW-1478, SW-1416

https://searchworks.stanford.edu/view/351225
https://searchworks.stanford.edu/view/495733
https://searchworks.stanford.edu/view/10164799

The bound-with note in yellow should not appear if the 590 does not have a subfield c.

jkeck:
Closes #1092 

Note 6d7eadc adds a Base MARC Field parsing class that handles most,if not all of the universal parsing code.

1ac3d40 build on top of that class to allow us to easily pull out the bound-with note.

My hope is that we can use (and enhance) this new MARC field parsing class for the upcoming MARC issues when possible, instead of trying to coerce the existing parsing code.

## 3942974 (has $c)
3942974-holdings

## 351225 Before (no $c)
351225-before

## 351225 After (no $c)
351225-after

### DIFF
--- a/app/models/marc_field.rb
+++ b/app/models/marc_field.rb
@@ -1,0 +1,144 @@
+###
+#  MarcField is the base class to provide common
+#  behavior for other classes to inherit from
+###
+class MarcField
+  attr_reader :document, :tags
+  delegate :present?, to: :relevant_fields
+
+  def initialize(solr_document, tags = [])
+    @document = solr_document
+    @tags = tags
+    run_preprocessors
+  end
+
+  def inspect
+    "#<#{self.class.name}:0x#{object_id} @document=#<SolrDocument:#{document.id}>>"
+  end
+
+  def label
+    I18n.t(
+      "searchworks.marc_fields.#{i18n_key}.label",
+      default: :"searchworks.marc_fields.default.label"
+    )
+  end
+
+  def values
+    relevant_fields.map do |field|
+      field.subfields.map(&:value).join(subfield_delimeter)
+    end
+  end
+
+  def to_partial_path
+    'marc_fields/marc_field'
+  end
+
+  private
+
+  def marc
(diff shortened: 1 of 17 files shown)


app/models/marc_field.rb
```diff
@@ -0,0 +1,138 @@
...
+  def append_unmatched_vernacular_fields
+    marc.fields(['880']).each do |vernacular_field|
+      next unless field_has_vernacular_matcher?(vernacular_field)
+      tag, matcher = vernacular_field_tag(vernacular_field)
+      next unless tags.include?(tag) && matcher == '00'
+      relevant_fields.insert(tag_indexes[tag], vernacular_field)
+    end
+  end
+
+  def tag_indexes
```
ndushay: skimming this class, everything makes sense to me except this method.  Perhaps a method comment would be useful?

ndushay:
this looks good to me except for me not understanding the function of the tag_indexes method.  Oh ... and should any of the model methods be protected or anything?

app/models/marc_field.rb
jkeck: Will add a comment.

jkeck:
@ndushay I don't understand what you mean by "should any of the model methods be protected or anything?".

The only public methods in `MarcField` are `inspect`, `label`, 'values`, and `to_partial_path`.  Everything else is private.

ndushay:
d'oh - missed the private.  Yay me.  So just confused by the one method - perhaps not so crucial.

jkeck:
:+1: Added a comment, rebased, and pushed.  Let me know if the comment doesn't makes sense or if you see some other things I should add.

Thanks for the :eyes:

ndushay:
:+1:   when it's ready, go for it.

Issue #1141, opened 2016-01-13
Display 260/264 $3


jvine:
This applies to MARC tags 260 and 264. Here is an example of a record with multiple 264 tags that need subfield 3 (|3) displayed to make sense: 

https://searchworks.stanford.edu/view/11527026

This should just be a matter of "turning on" display of subfield 3. 

Also, the subfields in 260 and 264 should display in the order in which they are present in the MARC record. That is, subfield 3 --> subfield a --> subfield b --> subfield c and so on. 

SW-1574

PR #1172, opened 2016-02-02
Replace MARC Imprint parsing with new subclass of MarcField.

jkeck:
This completes 1/2 of #1141  Will submit a separate PR for the 264.

Note that this PR includes #1170.  That one should be merged first.

## 11513964 (before)
11513964-before

## 11513964 (after)
11513964-after


ndushay:
aside from line note, this looks good.    What a drag that 264 and 260 have to be treated so differently.  Yay RDA.  Yay old data.

ndushay:
thanks for explanation.  This is also :+1: for merging when it's ready.   I really like this change.

spec/lib/search_works_marc_spec.rb
```diff
@@ -57,14 +57,4 @@
       end
     end
   end
-  describe 'matched vernacular' do
```
jkeck: I think we lost the line-note because it was on a commit that got rebased.

This particular test is being removed because it's testing the `MarcImprint` class which no longer exists.

In addition to that, this particular test is really attempting to test the vernacular field matching.  There are more generalizable tests against vernacular field matching in the new `MarcField` base class (which we'll hopefully slowly begin migrating everything to).

PR #1177, opened 2016-02-03
Update existing Instrumentation class to use new MarcField parsing.

### LINKED ISSUE #1133: Users don't notice/understand the library name link in access panel
Make the whole bar clickable
- [x] move the `<a>` up to surround `<div class="library-location-heading access-panel-heading">`
- [x] add hover state to the div: `background-color: #565347; /* $color-gray-90  */`

Make the library name `.access-panel-heading h3`  look like other links
- [x] remove the line-height setting (let it use default)
- [x] add:
```
border-bottom: 1px dotted white;
display: inline-block; /* so the border won't extend past the text */
margin-right: 10px; 
```
- [x] give it a hover/active state of `border-bottom: 1px dotted white;`

Default state:
clickable loc bar
Hover state:
clickable loc bar hover

(Click test results - users clicked the image more than the name; and adding an explicit "directions" or map icon caused distraction/confusion w stackmap.)

jkeck:
Closes #1133

This PR also includes the addition of several other subfields (identified in the original spec) that were not implemented initially in the instrumentation class.

## 7620108 (before)
7620108-before

## 7620108 (after)
7620108-after

## 10246985 (before)
10246985-before

## 10246985 (after)
10246985-after


ndushay:
This also looks good to go when  rebased / tests pass, etc.  :+1:

PR #1183, opened 2016-02-04
Split MARC 690 into separate local subject and remove…

### LINKED ISSUE #1114: 2nd 690 field not displayed
SW-1547

(I feel like dropping the "Local subject" logic altogether might be an acceptable solution.)

jkeck:
…includes 'collection' logic.

Closes #1114 

## 10771135 (before)
10771135-before

## 10771135 (after)
10771135-after


ndushay:
actually, given the original ticket, should the test ensure that more than 1 690 displays?

PR #1184, opened 2016-02-04
add language facet to home page; collapse resource facet by default

### LINKED ISSUE #1112: Add language facet to home page
Add language facet below Library facet
Close Resource type facet

SW-1555

peetucket:
fixes #1112

### DIFF
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -4,7 +4,7 @@ module FacetsHelper
 
 
   def collapse_home_page_facet?(facet)
-    facet.field == 'building_facet'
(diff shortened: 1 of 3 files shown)


app/helpers/facets_helper.rb
```diff
@@ -4,7 +4,7 @@ module FacetsHelper
 
 
   def collapse_home_page_facet?(facet)
-    facet.field == 'building_facet'
+    ['language','building_facet','format_main_ssim'].include? facet.field
```
ndushay: is format facet supposed to be collapsed now?

peetucket: that is how i read the ticket #1112

Issue #1140, opened 2016-01-06
Fix MODS date display in search results


jvine:
Applies to BCE dates, and AD dates earlier than 1000. Dates are displaying correctly in the record view, but not in results. 

Affects Roman Coins collection (BCE), as well as Walters Manuscripts (AD).

SW-1543

The date in the Title[Data]  will be completed when:
- [x] solrconfig change deployed to prod (sul-dlss/solrmarc-sw#123)
- [x] MODS data reindexed with sul-dlss/sw-indexer-service#52
- [x] #1186 is deployed

The imprint_display update requires:
- [x] PR for sul-dlss/stanford-mods#52
- [x] new release of stanford-mods gem
- [x] PR for sul-dlss/sw-indexer-service  to utilize new code in stanford-mods to populate imprint_display field
- [x] MODS data reindexed

ndushay:
This is related to:

sul-dlss/stanford-mods#56  early date display
sul-dlss/sw-indexer-service#43 dates for roman coins
sul-dlss/stanford-mods#53  u and uu display

ndushay:
And also related to sul-dlss/stanford-mods#52

ndushay:
@jvine  - close this if you agree.    Where the search result image went, I don't know -- separate issue.

AFAICT, the MODS records with early dates look great — the search results imprint statement and Title[Date] are better now.  

## Before
![walters before](https://cloud.githubusercontent.com/assets/96775/13198017/694f7bbe-d7b2-11e5-99a9-88bdc0c3b91f.png)

## After
![walters after](https://cloud.githubusercontent.com/assets/96775/13198027/9b513670-d7b2-11e5-96ac-f5fed322aca5.png)

## Roman Coins no imprint  but yes, Title[Date]
![more roman coins](https://cloud.githubusercontent.com/assets/96775/13198044/faf40c1a-d7b2-11e5-8e48-7f07a24f1653.png)

## Roman Coins with imprint
![more roman coins with imprint](https://cloud.githubusercontent.com/assets/96775/13198054/4cb65f9e-d7b3-11e5-97c3-6f7a17b35650.png)

jvine:
:+1:

PR #1186, opened 2016-02-06
Fix MODS date display in search result title

ndushay:
Connects to #1140  
which will be finished when:
- solrconfig change deployed to prod (sul-dlss/solrmarc-sw#123)
- MODS data reindexed with sul-dlss/sw-indexer-service#52
- This PR is deployed)

Also:
- connects to sul-dlss/sw-indexer-service#43 (search results piece of index BCE dates for coin collection)
- connects to sul-dlss/solrmarc-sw#111 (BCE dates for coin collection)
- connects to sul-dlss/stanford-mods#56:  display part of adds B.C. to dates before year 0, adds A.D. to dates
- connects to sul-dlss/stanford-mods#53:  display part of nice string for decades (1950s for 195u); nice string for centuries (18th century for 17uu)

## Roman Coins (before)
1140 before

## Roman Coins (after)
1140 after

## Papyri (before)
papyri dates before

## Papyri (after)
papyri dates after

### DIFF
--- a/app/helpers/results_document_helper.rb
+++ b/app/helpers/results_document_helper.rb
@@ -1,11 +1,15 @@
 module ResultsDocumentHelper
 
-  def get_main_title document
+  def get_main_title(document)
     (document['title_display'] || "").html_safe
   end
 
+  def get_main_title_date(document)
+    # MODS data is indexed with the display date in Solr field pub_year_ss
+    return "[#{document["pub_year_ss"]}]" if document["pub_year_ss"].present?
 
-  def get_main_title_date document
+    # the code below is for MARC data
+    # TODO:  have solrmarc-sw index the date display string into pub_year_ss for MARC data
     publication_year    = document["publication_year_isi"].to_s
     beginning_year      = document["beginning_year_isi"].to_s
     earliest_year       = document["earliest_year_isi"].to_s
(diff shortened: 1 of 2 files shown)


ndushay:
@jvine  this isn't deployed yet, but do the pix above give you a sense that this is what you had in mind?  This PR is only about the date in the title in the search result.

ndushay:
Here's another image, showing B.C. and A.D. in Roman Coins:

more roman coins

jvine:
Yes, looks good.

app/helpers/results_document_helper.rb
```diff
@@ -1,11 +1,16 @@
...
 
-  def get_main_title document
+  def get_main_title(document)
     (document['title_display'] || "").html_safe
   end
 
+  def get_main_title_date(document)
+    # MODS data is indexed with the display date in Solr field pub_year_ss
+    date = document["pub_year_ss"].to_s
+    return "[#{date}]" unless date.empty?
```
jkeck: Instead of casting `document["pub_year_ss"]` to a string and then checking if the string is `empty?` we can simply just check if `date` is `present?` (and drop the `to_s`).

This avoids the needless string casting which might be useful given that this code is executed a minimum of 20 times for every search result.

jkeck:
Very minor comment added. :+1:

ndushay:
@jkeck approved the revision out-of-band.  when build passes, I'm merging.

Issue #1165, opened 2016-01-30
Simple No results page from Advanced search


jvine:
- [ ] remove the orphan "Your search:" `<li>` 
- [ ] add a link to the previous page in the browser history - not a direct link to Advanced search. Goal is to return to the form with their values still present.

```
<ul class="list-unstyled zero-results-list">
  <li>
      <i class="fa fa-search-plus"></i>
      <a href="...browser back... "> Return to Advanced search</a> 
   </li>
   <li>Try using fewer terms to start, then limit your results.</li>
 </ul>
```

Before
![screen shot 2016-01-29 at 4 55 19 pm](https://cloud.githubusercontent.com/assets/6945691/12692335/757b6146-c6aa-11e5-85cd-8fe2e07d4124.png)

After
![screen shot 2016-01-29 at 4 54 31 pm](https://cloud.githubusercontent.com/assets/6945691/12692215/02c550ae-c6a9-11e5-95fe-8ef73444d849.png)

peetucket:
So this applies in all cases even when the user gets no results from a search of any kind?  In other words, the return text always says "Return to Advanced search page" even when they didn't come from there?

Also, there is another box below the one shown above that says "Check other sources" and has links to WorldCat and the library website.  This stays, correct?

jvine:
No. This applies only when the user comes to the no results from an Advanced search, which I think is signaled by &search_field=advanced in the URL - but there might be a better way.

The "other sources" box in fact does not appear for the Advanced search, and that's fine.

Here's a no result advanced search:  https://searchworks.stanford.edu/catalog?utf8=%E2%9C%93&op=AND&search=&search_title=aergrareg&search_author=&subject_terms=aergaerg&series_search=&pub_search=&isbn_search=&range%5Bpub_year_tisim%5D%5Bbegin%5D=&range%5Bpub_year_tisim%5D%5Bend%5D=&sort=relevance&search_field=advanced&button=

jvine:
And of course, the link to "back" will not work as expected when you go jump into a search URL like that. Fun.

peetucket:
got it, i'll leave previous behavior for other searches  as is

PR #1188, opened 2016-02-09
[WIP] change message when no search results from advanced search only

peetucket:


app/helpers/application_helper.rb
```diff
@@ -78,4 +78,7 @@ def disabled_class_for_no_selections(count)
       "disabled"
     end
   end
+  def from_advanced_search?
```
jkeck: I know it's a super simple helper method, but would you mind adding a unit test for this in `application_helper_spec.rb`?

peetucket:
fixes #1165

Issue #1185, opened 2016-02-05
Loading of image thumbnails is flickery and distracting; add placeholders and pleasant state transitions.


aeschylus:
A lot of the Searchworks application could be spruced up with some non-invasive css transitions. One of the more jarring effects is the loading of thumbnails. The grey placeholder is simply magically replaced by a flickering image. It would be nicer to show some kind of icon or loading placeholder and smoothly transition the thumbnail pixels in as they become available.

In general this is advisable with asynchronous content. It's position on the page should be held by an existing element that indicates the content's absence and status, and when it arrives it should render with [object constancy](http://bost.ocks.org/mike/constancy/).

mejackreed:
:+1: I like it

PR #1189, opened 2016-02-09
Issue#1122 collection preview styling

aeschylus:


app/assets/stylesheets/modules/image-collection-filmstrip.scss
```diff
@@ -82,6 +54,13 @@
 
       li:last-child {
         margin-right: 0;
+        width:100px;
+        height:100px;
+        padding:27px;
+        text-align: center;
+        span {
+            color:#666;
+        }
```
aeschylus: In the JS the size of the parent is set dynamically by the image size. There may be some opportunity to refactor how that is done to a more css-based approach in the future. Also, see: https://github.com/sul-dlss/SearchWorks/issues/1185

app/helpers/collection_helper.rb
```diff
@@ -3,12 +3,28 @@ def link_to_collection_members(link_text, document, options={})
...
+      if document.collection_members.total == 1
+        "View this item"
+      elsif document.collection_members.total == 2
+        "View both items"
+      else
+        "View all #{document.collection_members.total} items"
+      end
+    end
+  end
+
```
aeschylus: Using pluralise here didn't handle the cases where we need different wording altogether.

jkeck:
Could I bother you to rebase 55a446ee61ff9834656124b210daa85edcfdb225 out of this PR? That commit was already merged into master.

app/helpers/collection_helper.rb
jkeck: Makes sense. Perhaps unit tests for the two new methods that have been added would be appropriate.

aeschylus:
Whoops. Rebased yet again.

spec/features/responsive/home_page_responsive_spec.rb
```diff
@@ -6,20 +6,20 @@
...
-        expect(page).to have_css(".panel-title", text: "Resource type")
-        expect(page).not_to have_css("li a", text: "Book", visible: true)
-        expect(page).not_to have_css("li a", text: "Newspaper", visible: true)
+      within(".blacklight-access_facet") do
+        expect(page).to have_css(".panel-title", text: "Access")
+        expect(page).not_to have_css("li a", text: "Online", visible: true)
+        expect(page).not_to have_css("li a", text: "At the Library", visible: true)
       end
     end
   end
```
jkeck: I think this entire file is from a previous commit?  I don't think it has anything to do w/ this PR, but perhaps I'm wrong.

aeschylus: I'm not sure how to remove this. It has been rolled into the previous squashes and I'm not sure how to find out which real commit created it, or how to extract that one commit out of the now squashed commits.

jkeck: I would recommend doing something like `git reset HEAD~` to unstage the changes in your commit, and then checkout the files that do not have anything to do with your PR.

From a quick glance, it looks like there are unrelated changes in `_home_page.html.erb`, `facets_helper.rb`, and this file `home_page_responsive_spec.rb`.

If that doesn't work right off the bat, let me know and I'll see what I can do to help.

The main reason I'm being such a stickler about this is, if we need to revert this commit down the road we would inadvertently revert adding the language facet to the home page (which would be unexpected to say the least).

aeschylus: No qualms with stickling. Thank you for the direction; that seems to have worked.

ndushay:
@aeschylus  @jkeck  is this ready?  seems like it's been languishing.

aeschylus:
It is ready unless there are further comments.

It's been quite active, it's just been invisible because it gets rebased three times a day. @jkeck: regarding your deleted comment on a file that no longer exists, I'm all for stickling. Thank you for the direction.

I've added tests and did some minor refactoring, and removed some duplicated commits from master that were lingering.

Please merge.

jkeck:
Hi @aeschylus, sorry for the delay on this.  Now that the commit history is correct, I pulled it down and took it for a test drive (so I can see what the end results looked like).

I notice a potential oddity on the record view, and I just want to pass by @jvine to validate if this is the desired behavior or if there are any modifications we should make there.  Check out the `14 items online` and `60 items online` in the screenshots below.

### 6369086


### 6780453

jvine:
Seeing this in context, I actually like the text links on the record view. @aeschylus can you just style them in context, something like `{font-size: 1.2em; font-weight: normal;}` - so the computed font size is (close to) 16px.

aeschylus:
Sure, I will change this. I'm glad you like my unintentional CSS conflicts! I did check first to see if those classes were being used elsewhere, but I guess I missed this one.

db/schema.rb
```diff
@@ -36,19 +36,19 @@
...
     t.datetime "current_sign_in_at"
     t.datetime "last_sign_in_at"
-    t.string   "current_sign_in_ip",     limit: 255
-    t.string   "last_sign_in_ip",        limit: 255
+    t.string   "current_sign_in_ip"
+    t.string   "last_sign_in_ip"
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.boolean  "guest",                              default: false
+    t.boolean  "guest",                  default: false
```
jkeck: :question:

jkeck:
I removed the unrelated schema changes (some sort of database adapter issue?) and pushed a commit fixing the failing spec so we can get this merged in.

PR #1193, opened 2016-02-10
Update capistrano and capistrano-rails

jkeck:


ndushay:
:+1:  when builds pass.

PR #1196, opened 2016-02-11
Unify content type specific partials to mods and marc

### LINKED ISSUE #407: Support for type = Book
This is a feature ticket for SearchWorks to handle GryphonDOR content with type=Book. On the record view page this would include embedding the existing embeddable FlipBook viewer  that is also embedded on PURL pages (see: http://purl.stanford.edu/bb085ws2114).

What would be needed to display type=book in the results view, film-strip and gallery views?  Tagged as design needed in case any design review or input is necessary.  

Another consideration is how "Stanford Only" flipbooks will behave.

### LINKED ISSUE #1146: Disentangle display type
We need to remove the distinction between image and file in rendering items.

Collections will still need to know if they should render a filmstrip or a file list somehow.

jkeck:
Closes #1146 
Closes #407 

Before this PR, there are content type specific partials for marc records, image items, file items, image collections, file collections (and for almost all of those, also a potential merged/MARC variation).

This PR reduces that to simply marc items, mods items, marc colllections, and mods collections (or really, the only distinction being metadata format and if the object is an item or a collection).

The end result is nearly parity with production with the exception of #407 (example below)

## 421708 (before)
421708-before

## 421708 (after)
421708-after


jkeck:
I believe the reduction in coverage is due to a reduction in relevant lines of code.

app/models/concerns/digital_collection.rb
```diff
@@ -3,37 +3,63 @@ def is_a_collection?
...
-          rows: (@options[:rows] || 20)
+          rows: 20
         }
       )['response']
     end
-  end
 
+    def majority_of_documents_have_image_urls?
+      return unless number_of_documents_with_image_urls > 0
+      number_of_documents_with_image_urls > number_of_documents_without_image_urls
```
ndushay: I remember a bit about this discussion;  my recollection is that the choice for showing images vs. files was a bit arbitrary.  I think the images look nicer than the files -- should it be tweaked to show images more of the time?   >= instead of >?  Or if there are at least ... 5 image urls if there are 10 or more items?  @jvine  do you have an opinion?

ndushay:
I looked this over.  As you said, it is a lot of changes in one PR ... but I can't think of an intelligent way to split it up.

I think this change is complex enough that we will want to be sure someone pokes at the app manually.  Did @jvine or anyone else have a chance to do that with the display-type test app?  

I would be okay with any additional changes needed for this work to be submitted as separate PRs, unless the eyeball vetting of the app turns up something egregious quickly.

I'm not confident I really understand all of this code well enough to grok the tests super well, but the test coverage remains high.  Also, this is removing a lot of code, which is usually a good sign.  :-)

So overall, :+1: .   My only real concern is about the design issue:  are we using a satisfactory metric for choosing filmstrip vs. file list?

app/models/concerns/digital_collection.rb
jvine: I'd agree with >= but I don't think we want to weight too heavily in favour of images if it misrepresents the character of the collection.

jkeck: :+1: Updated the algorithm and added a test.

jkeck:
Thanks for the comments @ndushay I believe I have addressed them.  BTW, @jvine did poke around the deployed app (prob. not extensively, but we both have done some smoke tests over the various content types and nothing has come up yet).

ndushay:
@jkeck  feel free to merge after tests pass.  You've got my :+1:

jkeck:
Will merge once I merge #1189 so I can resolve the conflict in my PR instead of requiring the other PR to resolve mine (which may be a bigger pain).

PR #1202, opened 2016-02-12
Dynamically show Location facet when a whitelist of building facet …

### LINKED ISSUE #1174: Show Location facet when Art Library selected
- [ ] Create a whitelist of libraries that will show the Location facet. "Art & Architecture (Bowes)" is currently the only library, but others may follow.
- [ ] When the whitelisted library is selected in the Building facet, display the Location facet:
  - at the top of the facets list
  - open (same behaviour as Databases facet)

jkeck:
…values are selected.

Closes #1174 

art-locked-stacks-facet

### DIFF
--- a/app/controllers/concerns/location_facet.rb
+++ b/app/controllers/concerns/location_facet.rb
@@ -1,0 +1,31 @@
+##
+# A mixin to add dynamic libary specific location facet support
+module LocationFacet
+  SUBLOCATION_LIBRARIES = ['Art & Architecture (Bowes)'].freeze
+
+  extend ActiveSupport::Concern
+
+  included do
+    if respond_to?(:before_action)
+      before_action :add_location_facet, only: [:index, :facet]
+    end
+  end
+
+  protected
+
+  def add_location_facet
+    return unless building_facet_includes_library_with_sublocation?
+    blacklight_config.facet_fields['location_facet'].tap do |facet|
+      facet.show = true
+      facet.if = true
+    end
+  end
+
+  def building_facet_includes_library_with_sublocation?
+    params[:f] &&
+      params[:f][:building_facet] &&
+      params[:f][:building_facet].any? do |facet|
+        SUBLOCATION_LIBRARIES.include?(facet)
+      end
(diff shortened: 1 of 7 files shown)


jkeck:
@lmcglohon or @ndushay can you comment here to confirm once the config change is in production?

Thanks!

ndushay:
@jkeck   just put the config changes on SW solr master;  the changes will propagate to slaves with the next index update (within the hour, I expect).

ndushay:
@jkeck @lmcglohon   config changes are on all 3 SW solr slaves.  Green light.

jkeck:
:+1: thanks! This is ready for review then.

app/controllers/concerns/location_facet.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def building_facet_includes_library_with_sublocation?
+    params[:f] &&
+      params[:f][:building_facet] &&
+      params[:f][:building_facet].any? do |facet|
+        libraries_with_sublocations.include?(facet)
+      end
+  end
+
+  def libraries_with_sublocations
```
ndushay: another constant disguised as a method

ndushay:
What do you think of grabbing the Solr schema from https://github.com/sul-dlss/solrmarc-sw/blob/master/stanford-sw/solr/conf/schema.xml -- either as a one off copy now, or make the Rake task use it from there?  solrconfig-no-repl.xml   would be the solr config you'd want.

app/controllers/concerns/location_facet.rb
jkeck: :+1:

jkeck:
@ndushay I think it would be ideal to automate that in some way (and I'm not sure I have the bandwidth right now to take that on).

I'm not sure having the separate config is causing us any harm at this point, so I'm inclined to keep it as is.  If  you would like, please feel free to create a ticket for us to evaluate doing that in a future sprint.

ndushay:
Created issue  #1212.   No problem postponing it.

PR #1204, opened 2016-02-13
Migrate old serials parsing helpers to new MarcField parsed …

### LINKED ISSUE #1121: Linking behaviour for 76x - 78x serials
Currently: If 785 $t is present we do search_title; if $x we do all fields.

Desired behaviour, for all fields in the 76x - 78x range:
- [ ] If $s or $t is present (but not $a), do search_title 
- [ ] If $s AND $t are present, choose $s
- [ ] If $a is present - plus $s or $t - combine to search author_title
- [ ] If $x (ISSN) or $z (ISBN) is present, link it (separately) to isbn_search

Example with $a and $s
https://searchworks.stanford.edu/view/10750756

Example with linked $t that also has $a - only $t currently linked
https://searchworks.stanford.edu/view/10781763

Example with ISSN - currently doing all-fields search
https://searchworks.stanford.edu/view/358080

SW-1496, SW-494, SW-681, INDEX-217

jkeck:
…LinkedSerials class.

Closes #1121 

## 10750756 (before)
10750756-before

## 10750756 (after)
10750756-after

## 358080 (before)
358080-before

## 358080 (after)
358080-after

## 10781763 (before)
10781763-before

## 10781763 (after)
10781763-after


app/models/marc_fields/linked_serials.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def merged_with?(field)
+    field.tag == '785' && field.indicator2 == '7'
+  end
+
+  def merged_with_785s
+    @merged_with_785s ||= marc.fields('785').select(&method(:merged_with?))
+  end
+
+  def subfield_labels
```
ndushay: are these subfield labels or marc field/tag labels?

app/models/marc_fields/linked_serials.rb
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  private
+
+  def process_subfield(field, subfield)
+    return { text: subfield.value } unless respond_to?("#{subfield.code}_subfield", true)
+    send("#{subfield.code}_subfield", field, subfield.value)
+  end
+
+  def a_subfield(field, value)
```
ndushay: this might be a little less confusing named  subfield_a   instead of a_subfield, but I can certainly live with it the way it is.

app/models/marc_fields/linked_serials.rb
```diff
@@ -0,0 +1,125 @@
...
+      '0' => 'Continues',
+      '1' => 'Continues in part',
+      '2' => 'Supersedes',
+      '3' => 'Supersedes in part',
+      '4' => 'Merged from',
+      '5' => 'Absorbed',
+      '6' => 'Absorbed in part',
+      '7' => 'Separated from'
+    }
+  end
```
ndushay: shouldn't methods like this (and  "subfield_labels"  be constants instead?

```ruby
SUBFIELD_780_LABELS = {
     '0' => 'Continues',
     '1' => 'Continues in part',
     '2' => 'Supersedes',
     '3' => 'Supersedes in part',
     '4' => 'Merged from',
     '5' => 'Absorbed',
     '6' => 'Absorbed in part',
     '7' => 'Separated from'
}.freeze
```

ndushay:
[this comment WIP;  i already lost one version]

I like how this change shrinks the marc_helper and views.

Should the commit message should indicate the functionality is changing as well as the refactoring?  

I can't easily attest to the code's "correctness" in the functionality change -- perhaps @jvine could take a look?  Perhaps the before/after images are enough ... but it was a bit tough for me to assess if the tests were a rewrite of the old tests or testing the new spec.

Wondering about the constants disguised as methods ... I would expect the execution overhead of constants to be lower.

And some minor quibbles on some method names;  I can live with it the way it is.

spec/models/marc_fields/linked_serials_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      end
+    end
+  end
+
+  describe 'values' do
+    context 'when main entry ($a) is present' do
+      context 'when title data ($s or $t) is present' do
+        let(:marc) { main_entry_and_title_serial_fixture }
+
+        it 'included a hash that combines the main entry with the title and indicated author_title search' do
```
ndushay: Is this another way of saying " If $s AND $t are present, choose $s"  from #1121 ?    I confess confusion.

spec/models/marc_fields/linked_serials_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          expect(subject.values.first[:values].first[:link]).to eq 'Serial Main Entry1'
+          expect(subject.values.first[:values].first[:search_field]).to eq 'search'
+        end
+      end
+    end
+
+    context 'for standard number ($x or $z)' do
+      let(:marc) { main_entry_and_title_serial_fixture }
+
+      it 'links to the search_isbn sesarch field' do
```
ndushay: sesarch  typo if you're in here anyway

ndushay:
per the methods vs. constants question -- seems like it could be a religious war.  

Some good thoughts here:

http://stackoverflow.com/questions/15903835/class-method-vs-constant-in-ruby-rails
 https://github.com/bbatsov/ruby-style-guide/issues/328  <-- has some benchmarks
http://devblog.avdi.org/2011/08/18/do-we-need-constants/

if there is a reason to have them as methods, I suspect we want to at least memoize them?  I don't foresee them needing dynamic changes, or the method being overridden by a subclass ...

app/models/marc_fields/linked_serials.rb
jkeck: I'm going to leave this as-is.  For better-or-worse, this is consistent with a pattern that exists in other marc parsing classes.

app/models/marc_fields/linked_serials.rb
jkeck: :+1: Update to constant and renamed per suggestion.

app/models/marc_fields/linked_serials.rb
jkeck: :+1: Updated to constant and renamed per suggestion.

spec/models/marc_fields/linked_serials_spec.rb
jkeck: Yeah, I found the description a bit confusing.  The way this is described is more like the code is written (and is correct as vetted by @jvine).

All three of these cases are tested (now that I've pushed up an additional test).

spec/models/marc_fields/linked_serials_spec.rb
jkeck: :+1: Good :eyes:

jkeck:
I admit I haven't jumped on the constants train for stuff like this, but I can live with it, and it's probably the correct pattern.  FWIW, the execution overhead for those particular methods should have been extremely low.  I wouldn't really think that's a good use-case for memoization, but perhaps that's for a different conversation.

You are right, I was missing a couple tests and pushed them up to test the title search scenario.  Thanks for catching that.

Also update the commit message.

Thanks for the :eyes: @ndushay!

Issue #1201, opened 2016-02-11
Change text of "At the library" button


jvine:
In search results accordion section, change "At the library" to "Check availability". 

Before
![screen shot 2016-02-11 at 3 27 53 pm](https://cloud.githubusercontent.com/assets/6945691/12994275/226496f6-d0d4-11e5-851b-72b73ad8f7a3.png)

After
![screen shot 2016-02-11 at 3 28 15 pm](https://cloud.githubusercontent.com/assets/6945691/12994278/249bfb58-d0d4-11e5-8589-26b15c3bd566.png)

PR #1207, opened 2016-02-17
Fixes #463 and #1201. 

### LINKED ISSUE #463: Accessibility of accordion controls on results
Needs to alert SR to clickability and open/closed state. 

- [ ] Put href on accordion header so it can be tabbed to

aeschylus:
…dded aria labels with toggling of snippet and section body. The former link in now a button, and is tabbable and togglable by means of the space bar or enter key. There is still some strange animation behaviour from bootstrap.


jkeck:
:+1:

PR #1208, opened 2016-02-17
Add virtual object support

### LINKED ISSUE #1051: Virtual objects: link from child to parent
Individually-discoverable child objects will have a special "related item" link to their parent record.

example child
http://searchworks-test.stanford.edu/preview/view/yr145dc7638
and parent
http://searchworks-test.stanford.edu/preview/view/10449093

Create a sidebar block in the child record that links to the parent. Appears above the "Collection" block (if there is one).

Block title is "Appears in"
Body shows the material-type icon, title, and date of the parent. Title is linked directly to the parent record.

screen shot 2015-08-25 at 2 51 56 pm

screen shot 2015-08-25 at 2 54 55 pm

jkeck:
Closes #1051 

_Note because the refactor in 1ef412d requires synchronizing with an index change_

## Collection
_The 6 collection children are 2 virtual objects, each of which have 2 child documents_
collection

## Virtual object
set

## Item
item


jkeck:
FWIW, I believe the `0.04%` drop in coverage is not legitimate.  I think this is just caused by lines-of-code changes.

jkeck:
Just rebased this on-top of master and should be ready :ship:

PR #1209, opened 2016-02-17
Add MARC 351 to the Contents/Summary section.

### LINKED ISSUE #1194: Display MARC tag 351 in Contents/Summary section
SW-1595

Display tag 351, in Contents/Summary section
Subfields $3 $a $b $c in the order they appear in the MARC (which seems to be $3 $c $a $b)
Ignore $6 and $8

No special punctuation; just a space between each subfield. Label for dipslay is "Organization & arrangement"

example record https://searchworks.stanford.edu/view/10604415

jkeck:
Closes #1194

### DIFF
--- a/app/models/marc_fields/organization_and_arrangement.rb
+++ b/app/models/marc_fields/organization_and_arrangement.rb
@@ -1,0 +1,22 @@
+##
+# MarcField class to parse MARC 351s. The MARC spec indicates that the only
+# human readable subfields for this particular field are $3, $a, $b, and $c.
+class OrganizationAndArrangement < MarcField
+  private
+
+  def preprocessors
+    super + [:whitelist_subfields]
+  end
+
+  def whitelist_subfields
+    relevant_fields.each do |field|
+      field.subfields = field.subfields.select do |subfield|
+        %w(3 a b c).include?(subfield.code)
+      end
+    end
+  end
+
+  def tags
+    ['351']
+  end
+end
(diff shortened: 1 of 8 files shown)


jkeck:
@jvine is the placement in the `Contents/Summary` section correct (the first label/value)?

jvine:
@jkeck Can Finding Aid be pinned to the top of that section? since it's out of order anyway. Then everything else in MARC order.

jkeck:
@jvine:

jvine:
:+1:

app/models/marc_fields/organization_and_arrangement.rb
```diff
@@ -0,0 +1,22 @@
...
+    relevant_fields.each do |field|
+      field.subfields = field.subfields.select do |subfield|
+        %w(3 a b c).include?(subfield.code)
+      end
+    end
+  end
+
+  def tags
+    ['351']
+  end
```
ndushay: It's not worth fixing, but another constant hiding as a method

spec/models/solr_document_spec.rb
```diff
@@ -109,4 +109,10 @@
       expect(subject).to be_kind_of MarcSeries
     end
   end
+
+  describe 'MarcOrganizationAndArrangmenet' do
```
ndushay: not worth fixing, but there's a typo here.

ndushay:
:+1: to merge;  ok as is.

app/models/marc_fields/organization_and_arrangement.rb
jkeck: As discussed w/ @ndushay in person, the `tags` method is the API set by the base `MarcField` class so that has farther reaching consequences.

PR #1214, opened 2016-02-18
Update custom font and app code to include Archived website icon.

### LINKED ISSUE #1211: Add Web Archive Icon to Icon Set

jkeck:
Closes #1211 

warc-icon


ndushay:
This looks okay to me, but is definitely outside my comfort zone.  For example, why did 2 lines move to  app/assets/stylesheets/modules/sul-icons.scss?  Perhaps @aeschylus  would be a better reviewer for this?

jkeck:
Ah, good catch @ndushay.

The file `_sul-icons-rails.scss` is a file that is auto-generated by FontCustom.  However; it appears a developer had previously added some modifications to that file in a later commit.  When I re-generated that file w/ FontCustom I noticed that those two lines were removed in the diff (which didn't make sense) and I noticed in the blame that those were added later.  So I moved them to our custom `sul-icons.scss` which is where our icon customizations belong (and that way we don't inadvertently lose it later if we need to re-generate the font again).

jkeck:
That did prompt me to look over the generated `.svg` and realized in auto-generated a copyright statement attributing the font to me.  I just removed that FWIW.

ndushay:
I'm still out of my zone here.  So those 2 icons are diff from other icons?  And belong in the other file?

jkeck:
Yeah, they are totally unrelated to the icon font and should have never been in that file in the first place.

jkeck:
And feel free to punt on this and let somebody else merge it.  I know it's not in what you're most comfortable with.

jkeck:
FWIW for whoever does review this, the only changes that are not automatically generated by FontCustom are in `lib/constants.rb` and `app/assets/stylesheets/modules/sul-icons.scss`.  The rest are auto-generated files that I copied over.

aeschylus:
Looks good to me. Merging.

In general with generated files (js bundles as well)I think there is an opportunity to more fully separate the front-end and back-end, but that is significant technical debt/re-architecting, so I don't think it should really change anything now/here. For example, why are these generated files checked in, while the final javascript and non-font css is regenerated as part of the workflow?

jkeck:
@aeschylus because this will all use SULStyles font in the future but since SULStyles requires ruby 2.2 (and our production servers aren't upgraded to 2.2 yet) we are just regenerating the icon font with the added icon.

In the future, we'll simply use SULStyles and the font will be hosted on our CDN.

PR #1215, opened 2016-02-18
Update formats in fixture data to ones that correspond with icons.

jkeck:
**Note: This only applies to the fixtures in dev/test**
## Before
facet-before

## After
facet-after


jkeck:
:pray:

PR #1218, opened 2016-02-19
Adds accessibility features to browse-nearby

aeschylus:


jkeck:
@aeschylus looks like you have a legitimately failing view spec: `spec/views/catalog/record/_callnumber_browse.html.erb_spec.rb`

The test is testing that the call number is in an `a` and probably just needs to be changed to be a `button`.

jkeck:
@aeschylus it looks like this change breaks some (unfortunately untested) behavior.

The `View full page` button for the call number browse is broken.  I believe the link's `href` was set from the `href` of the call number when that particular call number browse was expanded (https://github.com/sul-dlss/SearchWorks/pull/1218/files#diff-a91c901a2f11b7e3febe85900d3808e1R85).  Now that that's a button, there is no `href` and therefore the `View full page` link never gets updated.

Perhaps we can put what was the `href` of the call number link into a data attribute on the button so the `View full page` link can get updated to point to the appropriate browse based on which call number browse is expanded (or whatever makes the most sense to you).

aeschylus:
Alright. I can fix this.

app/assets/javascripts/embedded-call-number-browse.js
```diff
@@ -22,10 +22,10 @@
       this.startDoc = item.data('start');
       this.embedContainer = this.embedViewport.find('.gallery');
       this.url = item.data('url');
-      this.browseUrl = item.attr('href');
+      this.browseUrl = item.data('index-path');
+      console.log(this.browseUrl);
```
jkeck: Leftover debugging?

PR #1223, opened 2016-02-19
Properly configure database topic facet so Blacklight knows…

### LINKED ISSUE #1089: Database topics are sorted alpha but numerical sort button is active
Related to #845? and SW-1537

Database topics default sort is alpha. Other facets have numerical default sort. All "more" popups show the default sort as Numerical. 

screen shot 2015-11-05 at 1 02 18 pm

jkeck:
… it is alpha/index sort by default.

Fixes #1089 

## Before
db-topic-facet-before

## After
db-topic-facet-after


ndushay:
:+1:  when PR travis completes green

PR #1232, opened 2016-02-29
Do not attempt to link nil link values in MARC series data

jkeck:


ndushay:
Trying to think of a smarmy comment about creating null version of the class and a factory to avoid checking for nil per Sandi Metz's class, but this will have to suffice.

PR #1235, opened 2016-02-29
Move Mouseflow config into settings.yml

jkeck:


config/settings.yml
```diff
@@ -28,3 +28,5 @@ OCLC:
   BASE_URL: http://www.worldcat.org/webservices/catalog/content/citations
   CITATION_FORMATS: ['all']
   DOCUMENT_FIELD: 'oclc'
+MOUSEFLOW: <%= Rails.env.production? %>
+MOUSEFLOW_KEY: '55d6233e-5245-4a7c-854d-f6e64e753af4'
```
jkeck: Just FYI, this key is not a secret (as it is plainly available in the markup).  The traffic needs to be coming from the host we've configured on the Mouseflow end, so there is no worries about exposing this key.

PR #1236, opened 2016-02-29
Remove buttons and related javascript, update styling for browse near…

aeschylus:
…by to enable scrollbar navigation

### DIFF
--- a/app/assets/javascripts/embedded-call-number-browse.js
+++ b/app/assets/javascripts/embedded-call-number-browse.js
@@ -122,12 +122,12 @@
           PreviewContent.append($galleryDoc.url, $galleryDoc.embedContainer)
             .done(function(data){
               $galleryDoc.updateDocs();
-              scrollOver();
               Blacklight.do_bookmark_toggle_behavior();
               $(".gallery-document h3.index_title a").trunk8({ lines: 4 });
               reorderPreviewElements();
               $galleryDoc.embedContainer.find('*[data-behavior="preview-gallery"]').previewEmbedBrowse();
               $galleryDoc.addBrowseLinkDivs();
+              scrollOver();
           });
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/embedded-call-number-browse.js
```diff
@@ -138,48 +138,25 @@
...
       // Scroll to the current document
-      function scrollOver(additional){
+      function scrollOver(){
         var current = $galleryDoc.docById();
-        var left;
+        var scrollAmount;
         if (current.length){
-          left = current.position().left;
+          console.log('I have entered the branch');
+          console.log(current.position());
```
jkeck: Leftover debugging cruft?

app/assets/javascripts/embedded-call-number-browse.js
```diff
@@ -138,48 +138,25 @@
...
+          scrollAmount = current.position().left - current.parent().width()/2 + current.innerWidth()/2;
         }else{
           console.log('could not get current position');
```
jkeck: More debugging

jkeck:
@aeschylus I think you mentioned something about adding a commit in w/ this PR to address the book-cover alignment issues in the embedded call number browse.

Perhaps either I was confused about where that was coming from, or it got bumped from these changes.  It looks like they're still not center-aligned.  Would you like me to take care of that in a separate PR?

aeschylus:
Yep, I put this up since I was not able to run the tests locally to see if anything failed. This is not ready for merge.

aeschylus:
@jkeck Looks like I committed the branch for that but never made a respective pr. 

https://github.com/sul-dlss/SearchWorks/pull/1238

Also removed the debug code from the JS here.

PR #1242, opened 2016-03-02
Add Rumsey Map Center as a library

jkeck:
This data does not currently exist in prod so it is non-op (and should be fine to :ship:). It is deployed to a UAT instance for review, but it should be fine to be merged before review (we can address any issues that come up in subsequent commits).


ndushay:
I agree that this change will not break anything in production.

It seems to treat Rumsey largely like SPEC;   I have no means to evaluate if Rumsey will behave as desired, or to evaluate if tests exist to ensure Rumsey will behave as desired ... unless it's in a github issue that I'm unaware of.

jkeck:
Unfortunately there isn't really a ticket that I'm aware of.  This was pretty much live coded w/ @jvine present at sort of the last minute.

Also, this code is pretty terrible and the cost of adding libraries is pretty high (I guess we aren't doing it often enough to warrant refactoring this code, but I really want to do it in the future if we're going to be doing heavy modifications of the availability logic).

jkeck:
I'm going to :ship: this as is.  We'll address any issues w/ how we have implemented the spec once stakeholders can review.

PR #1244, opened 2016-03-02
Move online label into model method and change algorithm to return…

jkeck:
…labels for non-collections with druids.

## 6314154 (before)
6314154-before

## 6314154 (after)
6314154-after

### DIFF
--- a/app/models/concerns/collection_member.rb
+++ b/app/models/concerns/collection_member.rb
@@ -23,6 +23,12 @@ module CollectionMember
     @index_parent_collections
   end
 
+  def online_label
+    return if is_a_collection?
+    return if druid.blank?
+    "<span class='online-label'>Online</span>".html_safe
+  end
+
   private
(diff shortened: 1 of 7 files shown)


ndushay:
nice job DRYing up the code.  

It's a bit strange on the surface that a concern called collection_members is where code lives to display online label if approp. for non-collection members.  Not sure if a diff concern would be a better home, or if "collection_members" is a slight misnomer, or what.

Not a blocker to merge, but worth a few seconds of thought.

jkeck:
I agree the module name is probably not the greatest, but I think maybe I have been confusing in another way somehow.

This code applies to items that are **not** collections (or in other words, **are** members of a collection), which is why I decided to put it in the `CollectionMember` module.

Perhaps my commit message is confusing. :/

jkeck:
Sound like there are no objections.

PR #1254, opened 2016-04-06
Drop druid from thumbnail of multi-managed-purl switcher.

jkeck:
## 8923346 (before)
8923346-before

## 8923346 (after)
8923346-after


ndushay:
:+1:   Wow - that change makes it work?  Cool.

PR #1269, opened 2016-07-30
Update travis build matrix

cbeer:


jkeck:
I need to look more into this failure on ruby 2.3.1.  It appears that something in the blacklight gallery gem is causing this (note that we are pinned to an old version).

PR #1286, opened 2016-10-18
Add a flag that allows for the availability lookup to omit live data …

jkeck:
…from symphony.

Connected to sul-dlss/sul-requests#547

### DIFF
--- a/lib/holdings.rb
+++ b/lib/holdings.rb
@@ -59,8 +59,9 @@ class Holdings
     end
   end
 
-  def as_json(*)
-    live_data = JSON.parse(LiveLookup.new(@document[:id]).to_json)
+  def as_json(live: true)
+    live_data = {}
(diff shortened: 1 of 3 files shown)


lib/holdings.rb
```diff
@@ -59,8 +59,9 @@ def mhld
     end
   end
 
-  def as_json(*)
```
ndushay: so the * here has always been superfluous?  No other consumers besides requests app?

jkeck: Yes, I think rails might (or may have) passed an implicit parameter when doing something like `render json: myObject` (where that would effectively call `myObject.as_json(root: true)` or something like that).  I think that parameter has an affect on ActiveRecord objects (where it's typically used) but in this case we don't want to do anything different.

PR #1292, opened 2016-10-28
Set a 2.5 second timer for receiving height updates from requests.

jkeck:
If the timer expires we set the height to something that should be large enough to handle the WebAuth form.

Part of sul-dlss/sul-requests#492
## WebAuth Form Before

before
## WebAuth Form After

_Note: The bottom of the page is cutoff by the browser not the modal and is scrollable_
after


jkeck:
There is no way to test this because we cannot load iframe content in our test suite.

PR #1315, opened 2017-05-15
Make updates that allow ruby 2.3.4

jkeck:


camillevilla:
👍 Tried this out locally and things look good

PR #1325, opened 2017-06-06
Add SPEC-INPRO to locations

### LINKED ISSUE #1317: New location for SPEC in RWC
Symphony Current Location = SPEC-INPRO
SearchWorks translation = In process (yes, it should look just like any other item "In process".)

camillevilla:
Closes #1317

Adds SPEC-INPRO to:

- LOCS
- UNAVAILABLE_LOCS
- UNAVAILABLE_CURRENT_LOCS

- [x] I'm still wrapping my head around the notation. Can anyone confirm whether this is a valid fixture?
`"123 -|- SPEC-COLL -|- GUNST -|- SPEC-INPRO -|- -|- -|- -|- -|- ABC 123"` for `:item_display`

I also need some further clarification:

- [x] should users be able to request `SPEC-INPRO` items? => No
- [x] should SPEC-INPRO belong to `UNAVAILABLE_LOCS` and/or `UNAVAILABLE_CURRENT_LOCS`?  => SPEC-INPRO will be a current location

spec-inpro


jkeck:
@camillevilla that looks good to me.  If we're in the `Callnumber` class at some point we should add a comment there about the format of this field.  You can "kind-of" suss it out through looking at the class itself.

https://github.com/sul-dlss/SearchWorks/blob/master/lib/holdings/callnumber.rb

You can see from there (in methods like `type`, `truncated_callnumber`, `shelfkey` etc) that there is code that deals w/ an `item_display` array positionally (e.g. `item_display[2]`) that let you know what each position of the field maps to.

So in that field you have a barcode, library code, home location code, current location code, and a callnumber, which looks 👍 to me.

camillevilla:
@jkeck AH! Thanks for that. I'm bookmarking that one...

I got some further clarification from @seestone about this and updated the fixture. `SPEC-INPRO` should behave similarly to `INPROCESS`, as a current location entered in MARC 999 subfield k.

PR #1326, opened 2017-06-06
[WIP] add X-UA-Compatible to head content

### LINKED ISSUE #1322: Force IE to render the site without compatibility mode
See https://github.com/sul-dlss/purl/pull/127 for details. 

This issue applies to SW and Requests as well as PURL. In IE, if compatibility mode is turned on for stanford.edu, the column layouts break - everything is left aligned and 100% width. Can be replicated with crossbrowsertesting. Managed computers seem to have this issue.

Here's an example of what it looks like:
iesw

drh-stanford:
This PR fixes #1322. Based on https://github.com/sul-dlss/purl/pull/127

Marked **WIP**:

* [ ] Deploy to stage(?) or somewhere where we can get to it from the outside so we can test
* [ ] Test with IE


jkeck:
I think there is a precedent for putting this out in the wild w/o necessarily needing to do manual testing.  We've implemented the same header in other web properties with success, so I'm not sure we need to hold this up (it's fairly standard).

cbeer:
👍 to just shipping it

PR #1327, opened 2017-06-06
Move index/marc language processing to a new Language MarcField…

### LINKED ISSUE #1319: Change label "language" to "notation" for music
Leader/06 = c or d 
and 
546 contains only subfield |b 
the Searchworks display label for the field should be Notation. 

(It would also be cool to get rid of the ". " that precedes the subfield 
|b data when there is no subfield |a present)

[SW-1077](https://jirasul.stanford.edu/jira/browse/SW-1077)
 
Examples
http://searchworks.stanford.edu/view/10093706 
http://searchworks.stanford.edu/view/9217198 
http://searchworks.stanford.edu/view/10108264

jkeck:
…subclass.

Closes #1319 

For the screenshots below, look at the Language/Notation field for the changes.

## 11412204 (before)
11412204-before

## 11412204 (after)
11412204-after

## 10093706 (before)
10093706-before

## 10093706 (after)
10093706-after

### DIFF
--- a/app/models/concerns/marc_metadata.rb
+++ b/app/models/concerns/marc_metadata.rb
@@ -1,0 +1,7 @@
+##
+# A mixin to include MarcField subclasses into the SolrDocument
+module MarcMetadata
+  def language
+    @language ||= Language.new(self)
+  end
+end
--- a/app/models/marc_fields/language.rb
+++ b/app/models/marc_fields/language.rb
@@ -1,0 +1,42 @@
+##
+# A class to merge content from the indexed
+# language data and the MARC 546 field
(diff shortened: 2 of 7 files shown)


drh-stanford:
The PR looks good, but I don't see a difference between the screenshots for 11412204.

jkeck:
@drh-stanford yeah, that was just intended to show that there was no change (although by happenstance we're not adding an extra space character in there somewhere).

In this https://github.com/sul-dlss/SearchWorks/issues/1319#issuecomment-304341666 that record is given as an example of something that **should not** get the `Notation` label due to the fact that it contains more that the subfield $b.

PR #1333, opened 2017-06-07
add new LinkedAuthor class for rendering linked "authors" and 711 $j suppression

### LINKED ISSUE #1321: "Role" should not be part of name link: 110 $e, 111 $j, 711 $j not working correctly
Basic rule:
- for 100, 110, 700, 710, subfield e should be displayed but should not be part of the linked search.
- for 111, 711 subfield j should be displayed but should not be part of the linked search.

This is working for most cases, but still incorrect for these tags:

## 110 $e
https://searchworks.stanford.edu/view/10159310
Correct link should be:
[Ecuador. Procuraduría General del Estado,] author, issuing body.
(exclude author, issuing body)

https://searchworks.stanford.edu/view/9197584
[Augustinians. Prior General (1570-1581 : Guidelli),] author.
(exclude author)

## 111 $j
https://searchworks.stanford.edu/view/10165011
[Technical Workshop on Organic Agriculture (1st : 2010 : Ogbomoso, Nigeria)] creator.
(exclude creator)
 
## 711 $j 
https://searchworks.stanford.edu/view/9952016
[Biennial Forum of Government Auditors (19th : 2012 : Alexandria, Va.).] author.
(exclude author)

Special case: For 711 that includes $t (title), we need to suppress the $j from the search, but include it in the link text, because it falls between two elements (name + title) that are in the same search. We did this previously with 700 and 710, examples here:
700 $e when $t is present: 10168859, 9615329 
710 $e when $t is present: 9197584, 9685915

Lots of prior discussion here:
[SW-993](https://jirasul.stanford.edu/jira/browse/SW-993)

drh-stanford:
This PR fixes #1321. It introduces a new `MarcField` class to handled linked "authors" so that we can implement 110 and 111 records. Currently, we're just using data from Solr fields to display and link the record. The linking semantics are more nuanced as described in #1321 so we need to generate them directly from MARC. This PR also ports over the 100 records previously handled by `link_to_author_from_marc`. The changes appear in both search results and show pages.

### Before 110$e

screen shot 2017-06-07 at 2 16 24 pm

### After 110$e

screen shot 2017-06-07 at 2 16 56 pm

### Before 111$j

screen shot 2017-06-07 at 2 19 46 pm

### After 111$j

screen shot 2017-06-07 at 2 20 25 pm

### Before 711$j

screen shot 2017-06-07 at 2 21 14 pm

### After 711$j

screen shot 2017-06-07 at 2 25 57 pm

### Before in search results

screen shot 2017-06-08 at 11 28 58 am

### After in search results

screen shot 2017-06-08 at 11 29 20 am

### DIFF
--- a/app/models/marc_fields/linked_author.rb
+++ b/app/models/marc_fields/linked_author.rb
@@ -1,0 +1,71 @@
+##
+# A class to parse MARC "author" fields that have hyperlinks for searching/display
+# (i.e., href vs link/anchor) and post text for display (context).
+#
+# Initially, ported over from solrmarc-sw `author_corp_display`, `author_meeting_display`, and
+# based in part on SearchWorks' `link_to_author_from_marc`
+#
+class LinkedAuthor < MarcField
+  TAGS = {
+    creator: %w[100],
+    corporate_author: %w[110],
+    meeting: %w[111]
+  }.freeze
+
+  attr_reader :target
+
+  def initialize(document, target)
+    @target = target.to_sym
+    super(document, TAGS[@target] || [])
+  end
+
+  def to_partial_path
+    'marc_fields/linked_author'
+  end
+
+  def i18n_key
+    super + '.' + target.to_s
+  end
+
+  def values
+    relevant_fields.map do |field|
+      field.subfields.each_with_object({}) do |subfield, hash|
+        output_for(field, subfield).each do |key|
+          hash[key] ||= ''
+          hash[key] << "#{subfield.value} "
+        end
+      end
+    end
+  end
+
+  private
+
+  # @return [Array<Symbol>] `:link`, `:search`, or `:post_text`
+  def output_for(field, subfield)
+    keys = []
+    keys << :link if linkable?(field, subfield)
+    keys << :search if searchable?(field, subfield)
+    keys << :post_text if post_text?(field, subfield)
+    keys
+  end
+
+  def linkable?(_field, subfield)
+    case target
+    when :creator, :corporate_author
+      !%w[e i 4].include?(subfield.code) # exclude 100/110 $e $i $4
+    when :meeting
+      !%w[j 4].include?(subfield.code) # exclude 111 $j $4
+    end
+  end
+
+  def searchable?(field, subfield)
+    return false unless linkable?(field, subfield) # must first be linkable
+    !%w[t].include?(subfield.code) # exclude $t
+  end
+
+  def post_text?(field, subfield)
+    return true if %w[e 4].include?(subfield.code) # always $e $4
+    return false if linkable?(field, subfield) || searchable?(field, subfield) # cannot be linkable or searchable
+    !%w[i].include?(subfield.code) # exclude $i
+  end
+end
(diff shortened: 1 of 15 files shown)


app/views/marc_fields/_linked_author_index.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if linked_author.present? %>
+  <% linked_author.values.each do |author| %>
+    <%= link_to(author[:link].strip, catalog_index_path(q: '"' + author[:search] + '"', search_field: 'search_author')) %>
+    <%= author[:post_text] if author[:post_text] %>
+    <br/>
```
jkeck: Is this line-break necessary if this is being wrapped in an `<li>`?

app/views/catalog/_marc_search_results_document_fields.html.erb
```diff
@@ -17,23 +17,15 @@
...
-    <% author_corp = link_to_data_with_label(document, "Corporate Author", 'author_corp_display', {controller: 'catalog', action: 'index', search_field: 'search_author'}) %>
-    <% unless author_corp.blank? %>
-      <%= render "field_list_from_index", :fields => author_corp %>
-    <% end %>
-
-    <% author_meeting = link_to_data_with_label(document, "Meeting", 'author_meeting_display', {controller: 'catalog', action: 'index', search_field: 'search_author'}) %>
-    <% unless author_meeting.nil? %>
-      <%= render "field_list_from_index", :fields => author_meeting %>
+    <% %i[creator corporate_author meeting].each do |target| %>
+      <% linked_author = LinkedAuthor.new(document, target) %>
```
jkeck: What would you think about adding accessors for the different author/creator flavors in https://github.com/sul-dlss/SearchWorks/blob/master/app/models/concerns/marc_metadata.rb?

This would allow us to avoid referencing classes directly in our views.

app/models/marc_fields/linked_author.rb
```diff
@@ -0,0 +1,77 @@
...
+    keys = []
+    keys << :link if linkable?(field, subfield)
+    keys << :search if searchable?(field, subfield)
+    keys << :post_text if post_text?(field, subfield)
+    keys
+  end
+
+  def linkable?(_field, subfield)
+    (creator? || corporate_author?) && !%w[e i 4].include?(subfield.code) ||
+      meeting? && !%w[j 4].include?(subfield.code)
```
jkeck: This method is a bit confusing.  If it doesn't make sense to break this down into smaller pieces, can you add a comment explaining a bit about what's going on here?

app/models/marc_fields/linked_author.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def searchable?(field, subfield)
+    !%w[t].include?(subfield.code) && linkable?(field, subfield)
+  end
+
+  def post_text?(field, subfield)
+    %w[e 4].include?(subfield.code) ||
+      (!linkable?(field, subfield) && !searchable?(field, subfield) && !%w[i].include?(subfield.code))
```
jkeck: Same with this method.  A comment might help clear up the logic of this method.

jkeck:
Few _hopefully_ minor comments.  👏 looks good!

app/views/marc_fields/_linked_author_index.html.erb
drh-stanford: i can move the `<li>` tags into this view partial. i ported this over from the field marc partial and it was to break up the venacular.

drh-stanford:
ok, i tried to address your comments. thanks!

PR #1343, opened 2017-06-09
Update to Blacklight 6 and Rails 5.1

cbeer:
There's some (hopefully minor) style differences that should probably be identified and addressed before this gets deployed to -prod.


jkeck:
Looks like we'll need to roll out the ruby 2.3.4 upgrade in order to get this deployed for testing.

cbeer:
👌. We can probably do that everywhere now that we have blacklight5 deployed, right?

PR #1347, opened 2017-06-09
add new LinkedRelatedWorks MarcField to handle 730 related works

### LINKED ISSUE #1340: Include more subfields in 730 related work linked search
SW-1640
Example records
https://searchworks.stanford.edu/view/9375544 
https://searchworks.stanford.edu/view/8447575

730 "Related work" link currently includes $i $a
Link should include:  a d f k l m n o p r s t
and display but not link:  h i x 0 3 5 6 8

drh-stanford:
This PR fixes #1340. This code extracts the logic from the marc_helper methods which is non-trivial, and I deleted the (newly) unused code from marc_helper.

### Before

screen shot 2017-06-12 at 3 55 21 pm

### After

screen shot 2017-06-12 at 3 55 03 pm

### Before

screen shot 2017-06-12 at 4 10 47 pm

### After

screen shot 2017-06-12 at 4 10 24 pm

### DIFF
--- a/app/models/marc_fields/linked_related_works.rb
+++ b/app/models/marc_fields/linked_related_works.rb
@@ -1,0 +1,95 @@
+##
+# A class to parse MARC related works (730 and others) and link them
+# to a search. Some subfields are not include in the search
+# but they do show up inside the link text, hence the logic in
+# the map/join methods.
+#
+class LinkedRelatedWorks < MarcField
+  LINK_CODES = %w[a d f k l m n o p r s t].freeze
+  TEXT_CODES = %w[h i x 3].freeze
+
+  def to_partial_path
+    'marc_fields/linked_related_works'
+  end
+
+  def tags
+    %w[700 710 711 720 730]
+  end
+
+  # @return [Array<Hash<String>>] output in `:pre_text`, `:link`, `:search`, and `:post_text`
+  def values
+    relevant_fields.map do |field|
+      join_to_view(map_to_sections(field))
+    end
+  end
+
+  def preprocessors
+    super + %i[filter_out_indicator2 filter_out_missing_titles]
+  end
+
+  private
+
+  def filter_out_indicator2
+    relevant_fields.reject! do |field|
+      field.canonical_tag != '730' && field.indicator2 == '2' # are not "included works"
+    end
+  end
+
+  def filter_out_missing_titles
+    relevant_fields.reject! do |field|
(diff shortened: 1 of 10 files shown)


app/models/marc_fields/linked_related_works.rb
```diff
@@ -0,0 +1,94 @@
...
+  }.freeze
+
+  # Organizes the `map` output into output suitable to render in a view and returned by `values`
+  # @return [Hash<String>] output in `:pre_text`, `:link`, `:search`, and `:post_text`
+  def join_to_view(subfields, result = {})
+    SECTION_TO_VIEW.each_pair do |from, destinations|
+      destinations.each do |to|
+        subfields[from].each do |subfield|
+          next if to == :search && TEXT_CODES.include?(subfield.code) # omit text subfields from searches
+          result[to] = '' unless result[to]
```
jkeck: This might be a bit more idiomatic written as `result[to] ||= ''`.

app/models/marc_fields/linked_related_works.rb
```diff
@@ -0,0 +1,94 @@
...
+class LinkedRelatedWorks < MarcField
+  LINK_CODES = %w[a d f k l m n o p r s t].freeze
+  TEXT_CODES = %w[h i x 3].freeze
+
+  def to_partial_path
+    'marc_fields/linked_related_works'
+  end
+
+  def tags
+    %w[700 710 711 720 730]
```
drh-stanford: yeah, it's complicated. `contributors_and_works_from_marc` parses those same tags ('700', '710', '711', '720') BUT only outputs results when `indicator2 == '2'`.  So, the LinkedRelatedWorks class filters out those with the `filter_out_indicator2` preprocessor.

I wasn't really sure how far to go in refactoring. For example, we could have a LinkedField class and then derive a LinkedRelatedWorks, LinkedIncludedWorks, and LinkedContributor and get rid of the `contributors_and_works_from_marc` function entirely. As it is in the PR now, I've only extracted the `LinkedRelatedWorks` (related_works) result from `contributors_and_works_from_marc`.

drh-stanford:
I've addressed the minor comments and force pushed those changes. Still pending is the discussion here https://github.com/sul-dlss/SearchWorks/pull/1347#discussion_r122009001

app/models/marc_fields/linked_related_works.rb
jkeck: That's totally fine.  Thanks for the explanation.

PR #1350, opened 2017-06-13
use openseadragon:install generator

### LINKED ISSUE #1349: StackMap broken
Possibly due to the Openseadragon update that happened?

Example ckey: 8551658  (click the Find-it link)

We're getting an `Uncaught ReferenceError: OpenSeadragon is not defined`

drh-stanford:
This PR fixes #1349. The latest version of the OpenSeadragon gem uses a generator for the install, and we were missing the files that it created/updated.

Note I'm not sure this code is needed? https://github.com/sul-dlss/SearchWorks/blob/master/config/initializers/assets.rb#L1. Also, I'm not sure how to test this... maybe a feature test that brings up the FindIt window?

### Before

screen shot 2017-06-13 at 10 47 06 am

### After

screen shot 2017-06-13 at 10 45 26 am


drh-stanford:
The coverage decreased but I'm not sure why.

jkeck:
I think it's due to the blacklight 5 upgrade.  Should be fine.

PR #1351, opened 2017-06-13
Chance OCLC api host back to http since https returns a 404.

jkeck:
Friggin' OCLC 🤦‍♂️


drh-stanford:
I'm getting an empty cite page like this. Do you have a catkey that's showing the citation?

screen shot 2017-06-13 at 11 52 32 am

jkeck:
@drh-stanford you need the API key.  Check out our [shared configs](https://github.com/sul-dlss/shared_configs/blob/searchworks-prod/config/settings/production.yml#L16-L17).

drh-stanford:
Yes, that was the issue. It's working for me now 👍 

screen shot 2017-06-13 at 2 37 21 pm

PR #1355, opened 2017-06-14
add support for stackmap blacklists and add ART as a new stackmapable library

### LINKED ISSUE #1304: Add StackMap to Art
Waiting to hear back if they have any location exclusions.

drh-stanford:
This PR fixes #1304.

Marked **WIP** until we can verify that the stackmap.com API is working for the ART library.

### After

screen shot 2017-06-14 at 3 13 31 pm

### DIFF
--- a/lib/holdings/callnumber.rb
+++ b/lib/holdings/callnumber.rb
@@ -144,12 +144,16 @@ class Holdings
       home_location == 'INTERNET'
     end
 
+    # supports whitelist for library
     def stackmapable_library?
       Constants::STACKMAP_LIBS.include?(library)
     end
 
+    # supports a global blacklist and local blacklist for home_location
     def stackmapable_location?
-      !Constants::HIDE_STACKMAP_LOCS.include?(home_location)
+      return false if Constants::STACKMAP_BLACKLIST[:global].include?(home_location)
+      return true if Constants::STACKMAP_BLACKLIST[library].blank? # no local blacklist registered, so we're done
+      !Constants::STACKMAP_BLACKLIST[library].include?(home_location)
     end
 
     def standard_or_zombie_library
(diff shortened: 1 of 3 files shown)


jkeck:
@drh-stanford how would you like to verify that the StackMap API works for the ART library?

jkeck:
@drh-stanford after following up w/ StackMap, it turns out they had an incorrect library code for the Art library.  They've updated this on their end and appears to be returning the correct maps/API responses for ART.

https://searchworks.stanford.edu/view/12062765/stackmap?callnumber=NB621+.M6+A13+2017&library=ART&location=STACKS&title=100+anni+%3A+scultura+a+Milano%2C+1815-1915

drh-stanford:
Looks good (screenshot --> catkey: 11958147)

screen shot 2017-06-22 at 10 45 51 am

lib/holdings/callnumber.rb
```diff
@@ -149,7 +149,12 @@ def stackmapable_library?
     end
 
     def stackmapable_location?
-      !Constants::HIDE_STACKMAP_LOCS.include?(home_location)
+      !Constants::STACKMAP_BLACKLIST[:global].include?(home_location)
```
jkeck: One suggestion here, which might simplify a bit.

What if we add a `blacklisted_stackmap_locations` method (or something like that) that did something along the lines of 

```ruby
Constants::STACKMAP_BLACKLIST[library] || Constants::STACKMAP_BLACKLIST[:global]
```

Then `stackmapable_location?` could basically be `blacklisted_stackmap_locations.include?(home_location)` and we could get rid of the `stackmapable_blacklisted?` method all-together (which also could simplify `stackmapable?` above).

Just a suggestion, if you still like this approach better, then I'm 👍

drh-stanford:
Thanks. I changed the implementation to modify only `stackmapable_location?` -- I think that simplifies the readability of the code. Let me know what you think.

PR #1356, opened 2017-06-14
run blacklight:marc:install rails generator for blacklight-marc 6.x

drh-stanford:
Results from `rails generate blacklight:marc:install`. In particular, please review the additions it made to SolrDocument.

### DIFF
--- a/app/assets/stylesheets/blacklight.scss
+++ b/app/assets/stylesheets/blacklight.scss
@@ -9,5 +9,7 @@
 @import 'bootstrap-overrides';
 
 @import 'blacklight/blacklight';
+@import 'blacklight_marc';
+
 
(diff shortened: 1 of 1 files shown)


app/models/marc_indexer.rb
```diff
@@ -0,0 +1,161 @@
+$:.unshift './config'
+class MarcIndexer < Blacklight::Marc::Indexer
```
cbeer: I don't think we're going to switch to traject any time soon. This and those `config/` additions should probably be removed.

drh-stanford:
Ok, thanks. It looks like that brings it down to just the css stuff after I removed the code duplication in SolrDocument

Issue #1359, opened 2017-06-15
Select-by-letter for Databases is not working


drh-stanford:
![screen shot 2017-06-15 at 9 50 41 am](https://user-images.githubusercontent.com/1861171/27192879-c91b22e4-51b1-11e7-8a81-c759d143f12f.png)

the select-by-letter thing for Databases isn't working for me -- see https://searchworks-stage.stanford.edu/?f%5Bformat_main_ssim%5D%5B%5D=Database&prefix=g&sort=title

cbeer:
I hope #1360 will fix this.

PR #1360, opened 2017-06-15
Switch to using the SearchBuilder pattern

cbeer:


drh-stanford:
The Database access point passes the filter-by-letter test in this PR:

```
Databases Access Point
  should be present when selecting the databases facet (PENDING: Needs format/resource-type facet to settle down)
  should sort by title by default (FAILED - 78)
  should display the database specific facet (PENDING: Needs format/resource-type facet to settle down)
  should filter by letter
```

so do all the CJK tests

the `Stanford::NearbyOnShelf` tests all fail which brings the failure count way up

drh-stanford:
Fixes #1359

PR #1365, opened 2017-06-15
fix missing public/500 template for bad requests

drh-stanford:
This PR fixes missing template 500 errors for bad requests to json apis.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -361,7 +361,7 @@ class CatalogController < ApplicationController
                                                    facets_from_request,
                                                    blacklight_config)
       end
-      format.html { render 'public/500', layout: false, status: 400 }
+      format.html { render status: :bad_request, layout: false, file: Rails.root.join('public', '500.html') }
     end
   end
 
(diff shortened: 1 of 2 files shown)


jkeck:
@drh-stanford I wonder if we can solve the same thing w/o the `Rails.root.join` in the controller by changing this: https://github.com/sul-dlss/SearchWorks/blob/blacklight6x/config/environments/production.rb#L19

The default for rails 5.1 is to use an ENV var for this setting (which we would have to have to set via puppet).  We can do that, or just set the value directly in our `production.rb`.

drh-stanford:
Yeah, I'm not sure what the best solution would be. The comment indicates that Apache already serves out the files which seems to be the case on stage (e.g., the `/robots.txt` is working with that ENV var missing). On my laptop I wasn't able to fix the Missing Template exception using `'public/500'` so the `Rails.root.join` was the approach I took -- the wrench is that a simple `render status: 400` doesn't work because we want the 500.html response -- maybe we could just create a 400.html?

jkeck:
I would be okay w/ adding a `400.html` if it solves the issue (although I'm not 100% sure that'll be the case).  I suppose it's ultimately not super important as these are error conditions for api responses, so we can clean this up later if we seed the need.

PR #1366, opened 2017-06-15
update to phantomjs 2.1.1 on Travis builds

drh-stanford:
I'm seeing spurious errors in the minimap test in the `Emailing Records` feature test (line 49). I noticed that we're not using phantomjs 2.x so this PR upgrades the Travis build to use 2.1.1 (based on ArcLight code) and increases the wait time to a minute on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,23 @@ env:
 
 sudo: false
 language: ruby
-cache: bundler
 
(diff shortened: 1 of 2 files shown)


drh-stanford:
nevermind. neither of these changes fixed the spurious errors :|

Issue #1378, opened 2017-06-20
Add drop-down to switch between search endpoints


jkeck:
Acceptance Criteria
- [x] Dropdown has 2 links: Catalog and Articles
- [x] Catalog link goes to home page
- [x] Articles link goes to new article landing page
- [x] Link in banner needs to reflect current context (i.e. articles when in article search)
- [x] Search box needs to search current context (i.e. articles when in article search)

Out of scope
* All link (will go to bento eventually)
* Links pass on current search context (in search form)

![screen shot 2017-06-20 at 11 22 25 am](https://user-images.githubusercontent.com/96776/27349015-cbbc3266-55aa-11e7-909c-ea8feec88170.png)

jvine:
Markup:

```
<div id="search-dropdown">
    <button type="button" aria-expanded="false" [etc...]>
       <span class="sr-only">search</span> all
    </button>
   <ul class="dropdown-menu" role="menu">
        <li class="active">
           <a href="[searchurl]">
            all <span class="help-block">catalog, articles, website, and more in one search</span>
            </a></li>
        <li>
           <a href="[searchurl]">
            catalog <span class="help-block">books & media in the Stanford Libraries'collections</span>
            </a></li>
        <li>
           <a href="[searchurl]">
            articles <span class="help-block">journal articles, e-books, and other licensed e-resources</span>
            </a></li>
    </ul>
</div>
```

PR #1384, opened 2017-06-21
adds a new Article search landing page at /article/home

### LINKED ISSUE #1379: Create article search landing page
Acceptance Criteria
* Has a 3 column layout with headers

Out of scope
* Facet in the first column

screen shot 2017-06-20 at 11 29 41 am

drh-stanford:
This PR fixes #1379.

Notes:
- I kept the empty helper/javascript/css that the controller generator -- I'm assuming that we'll need them later
- I wasn't sure if we wanted the "RELATED" links to be configurable or not? They are hardcoded into the view template like the current home page has. If so, I suggest we make a different ticket for that.
- The banner still says "SearchWorks catalog" but that's addressed in another ticket: #1378 
- The styling is not exactly like the screenshot in the #1379 ticket because I'm using the existing styling for the SearchWorks home page, i.e., the underlined headers, etc.
- The rails route is `/article/home` is an alias to a restful route `/article/new` as I'm assuming that we'll want more restful routes

### `/article/home`

screen shot 2017-06-22 at 10 32 40 am

### DIFF
--- a/app/assets/javascripts/article.js
+++ b/app/assets/javascripts/article.js
@@ -1,0 +1,1 @@
+// TODO: intentionally left blank
--- a/app/assets/stylesheets/modules/article.scss
+++ b/app/assets/stylesheets/modules/article.scss
@@ -1,0 +1,1 @@
+// TODO: intentionally left blank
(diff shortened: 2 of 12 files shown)


jvine:
This is totally fine for now. We aren't solid on any of the content below the h1 header at this point. There are some styling details etc. that should be addressed later. Agree on using the current SW styling at this point - I don't want to introduce inconsistency between catalog and articles.

drh-stanford:
Ok, I've removed the fake facet and moved all the styling into the css file.

PR #1388, opened 2017-06-23
Add a dropdown to switch between Catalog and Article Search

### LINKED ISSUE #1378: Add drop-down to switch between search endpoints

jkeck:
Closes #1378 

There is some left to be desired w/ the styling, but knowing that this search bar area is going to go through some significant changes for article search we figured we shouldn't spend too much time on it.

search-dropdown

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,10 @@
 module ApplicationHelper
+  # This is a path helper that routes the main search
+  # form based on the current search context.
+  def searchworks_search_action_path(opts = {})
+    return article_home_path(opts) if article_search?
+    search_catalog_path(opts)
+  end
 
   def render_search_bar_advanced_widget
     render partial:'catalog/search_bar_advanced_widget'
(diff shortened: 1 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,10 @@
 module ApplicationHelper
+  # This is a path helper that routes the main search
+  # form based on the current search context.
+  def searchworks_search_action_path(opts = {})
+    return article_home_path(opts) if article_search?
```
drh-stanford: I think this will need to change to `article_index_path` when that route becomes available.

PR #1390, opened 2017-06-23
Fix local broken tests

jkeck:
@camillevilla let me know if these fixes those failing tests you were seeing too.  You'll need to re-index your test fixtures, just a heads up.


camillevilla:
@jkeck wahoo! these passed locally for me.

PR #1395, opened 2017-06-23
ArticleController routes that does a simple article search and show page (no paging, facets, etc.)

### LINKED ISSUE #1380: Start migrating code from the EDS example app
This may be better off in a gem/engine, but for now, we will migrate the code into the SearchWorks codebase to be accessed by the article search code/functionality.

- [ebsco/eds_blacklight_search_app](https://github.com/ebsco/eds_blacklight_search_app)
- https://github.com/ebsco/edsapi-ruby

- [x] Migrate `Eds::Repository` and `Eds::SearchService`
- [x] Add blacklight configuration to ArticleController
- [x] Migrate `index` action
- [x] Migrate `show` action
- [x] Fix broken interaction between /catalog and /article re: `url_for_document` helper method
- [x] Finish pending specs
- [ ] Migrate paging support (or maybe put in a different ticket?)

drh-stanford:
This PR fixes #1380. It implements a basic article search and show page without Blacklight features like paging, facets, etc. To use, you need to add the EDS_* settings to your `config/settings/development.local.yml` (see Slack).

Notes:

- I marked the `search` test for the `Eds::Repository` as pending since the mocking for this is a little detailed. Similarly, I marked the feature spec for article search as pending -- we need a way to mock EDS API results basically. The views are also not spec'd as they are vanilla blacklight partials right now.
- Not all of the EDS metadata is shown -- they are working on providing us with a complete list of the metadata available. Notably, the links to full text, etc. are not in this PR.
- Not all the functionality in `Eds::Repository.send_and_receive` is being used right now -- specifically, the next/prev and multiple id fetching support.

### Search results
screen shot 2017-06-23 at 3 28 43 pm

### Show page

screen shot 2017-06-23 at 3 29 02 pm


app/controllers/article_controller.rb
```diff
@@ -2,5 +2,80 @@
...
 class ArticleController < ApplicationController
+  include Blacklight::Configurable
+
+  before_action :eds_init, only: %i[index show]
+  # TODO: probably need to move this into an Eds::SearchService initializer
+  def eds_init
+    session['guest'] = false # TODO: hardcoded to bypass auth
+    session['eds_session_token'] =
+      EBSCO::EDS::Session.new(
+        guest: false, # TODO: hardcoded to bypass auth
```
jkeck: Wondering if we should make this true by default, even for our initial implementation.  Currently, anybody on campus could theoretically access this site if they knew about it, and could potentially access metadata if items they are technically not supposed to be able to.

app/helpers/article_helper.rb
```diff
@@ -2,4 +2,19 @@ module ArticleHelper
...
+
+  def current_search_session
+    {} # TODO: placeholder
+  end
+
+  def doi_link(options = {})
+    doi = options[:value].try(:first)
+    return if doi.blank?
+    url = 'https://doi.org/' + doi.to_s
+    link_to(doi, url, target: '_blank')
```
jkeck: We typically try to avoid `_blank` targets for usability purposes.  @jvine any reason to keep this as a `target="_blank"`?

config/initializers/eds.rb
```diff
@@ -0,0 +1,3 @@
+require 'ebsco/eds'
+
+fail unless Settings.EDS_USER && Settings.EDS_PASS && Settings.EDS_PROFILE
```
jkeck: Can we provide some sort of message about why we're failing here?

spec/controllers/article_controller_spec.rb
```diff
@@ -7,4 +7,27 @@
       expect(response).to render_template('new')
     end
   end
+
+  it 'handles authentication'
+  it 'handles configuration'
```
jkeck: Are these `TODO` tests?  (I also see a few others in other places, just not familiar w/ this pattern)

drh-stanford: yes these are TODO/pending tests

jkeck:
I _think_ you can fix that failing test by rebasing on-top of master.  That test should have been fixed in #1390.

app/helpers/article_helper.rb
drh-stanford: I can remove it -- the only reason it's there is because that's how the EDS example helper had it.

app/controllers/article_controller.rb
drh-stanford: Ah, good point. I'll make it guest-only.

drh-stanford:
Ok, I think I've addressed all the feedback. Note that the coverage went down due to pending specs, in particular the ones where we need to mock the EDS API.

PR #1403, opened 2017-06-26
Add breadcrumbs to article search results

### LINKED ISSUE #1393: Add search breadcrumb to the results page
- add search breadcrumb to upper left of results page
- "Start over" returns to the Article search home page
- search parameters are represented in the same way as catalog search

screen shot 2017-06-23 at 11 04 51 am

camillevilla:
Closes #1393

- Start over button redirects to articles home page
- Added a feature test that should be skipped until we have an EDS stub
- only one breadcrumb displays until we have facets implemented

## Screenshot
screen shot 2017-06-26 at 4 46 07 pm


camillevilla:
Removing the last pill currently redirects to a blank page at:
`http://localhost:3000/article?utf8=✓`

In Searchworks, removing the last pill redirects to:
`https://searchworks.stanford.edu/?search_field=search&utf8=✓`

screen shot 2017-06-26 at 5 00 06 pm

I have two questions that we could maybe split into another ticket:

1. Should removing the last pill redirect to the articles home page? Or some other general search result? e.g. maybe only Stanford things? A blank articles search currently raises an error: `EBSCO::EDS::BadRequest`
2. Where do we configure that behavior?

camillevilla:
RE: 
1. default behavior will be a redirect to the articles home page. EBSCO does not currently allow a blank search.

config/routes.rb
```diff
@@ -69,5 +69,6 @@
   resources :course_reserves, only: :index, path: "reserves"
 
   get 'article/home' => 'article#new' # alias for article search home page
-  resources :article, only: %i[index new show], defaults: { format: :html }
```
camillevilla: @jkeck @drh-stanford I ran into errors when trying to use article_home_path in tests. I noticed our other routes don't have `format:html` specified. Is it ok to remove this?

spec/features/article_searching_spec.rb
```diff
@@ -41,4 +41,38 @@
 
     expect(current_url).to match(%r{/article\?.*&q=Kittens})
   end
+  describe 'breadcrumbs', js:true do
+   scenario 'start over button returns users to articles home page' do
+     skip('need to stub out Eds::SearchService') if ENV['CI']
+     visit article_home_path
```
camillevilla: @jkeck this pattern is starting to get a little redundant. Maybe I could DRY this out and add a helper?

jkeck: @camillevilla which pattern is starting to get redundant?  `visit article_home_path`?

config/routes.rb
jkeck: 👌 by me

spec/features/article_searching_spec.rb
camillevilla: @jkeck schwops, should have put this comment a few lines down. I meant:
```ruby
visit article_home_path

within '.search-form' do
  fill_in 'q', with: 'kittens'
  find('#search').trigger('click')
end
```

jkeck: Hmm, I suppose a `fill_in_search_form` (or something along those lines) could be appropriate.  I'll leave it up to you.

app/helpers/render_constraints_helper_behavior.rb
```diff
@@ -0,0 +1,17 @@
...
+  include Blacklight::RenderConstraintsHelperBehavior
+
+  def remove_constraint_url(localized_params)
+    scope = localized_params.delete(:route_set) || self
+
+    unless localized_params.is_a? ActionController::Parameters
+      localized_params = ActionController::Parameters.new(localized_params)
+    end
+
+    options = localized_params.merge(q: nil, utf8: nil, action: article_search? ? 'new' : 'index')
```
jkeck: Hmm, is the only thing that we're overriding this method is to add the `action` ternary?

I'm wondering this is an opportunity to look into updating how we handle the article search home page to be more in-line w/ how Blacklight works (which could prevent the need to override any helpers at all).

Blacklight uses the index action to render both the search results  and the home page (using a helper to determine if there are query params that should be applied and a result rendered.

If we kill `article#new` and use the same helper, we may be able to render a `home_page` partial, and be more in-line w/ how the Blacklight helpers are expecting the UI to be organized.

(perhaps you already went down this path, but this gave me pause).

camillevilla: @jkeck that sounds promising. I just rebased and the only remaining override is the `action` ternary. Article search's current index action returns this with Blacklight's default remove url query params (`?search_field=search&utf8=✓`):
screen shot 2017-06-30 at 12 16 46 pm

I think I found the helper you're talking about in [catalog/index.html.erb](https://github.com/sul-dlss/SearchWorks/blob/master/app/views/catalog/index.html.erb). I'll give it a shot and ping you when I have something.

camillevilla:
alrighty @jkeck this one's ready for review. Things to note:

- #1440 (error when using a blank search) is not resolved by this PR. 
- We get a no results page when the only constraint is a facet (see example below)

article_breadcrumbs_facet

jkeck:
I would like to look into a possible pattern for stubbing out the results before merging this (so we can unskip the tests).  If I can't get something useful by tomorrow let's 🚢 this.

spec/features/article_searching_spec.rb
```diff
@@ -67,4 +68,40 @@
...
+       fill_in 'q', with: 'kittens'
+       find('#search').trigger('click')
+     end
+
+     expect(page).to have_css('.appliedFilter', text: /kittens/)
+
+     within '#appliedParams' do
+       find('a.btn', text: /Start over/).trigger('click')
+       expect(page).to have_current_path(article_index_path)
+       expect(page).not_to have_css('.appliedFilter', text: /kittens/)
```
jkeck: You may want to do this outside of the `within` (or even in a separate `within` block below it).   Clicking on `Start over` has changed the page so I have found that the tests can complain about the element going way when that happens.

camillevilla: @jkeck thanks for the tip. It worked!

spec/features/article_searching_spec.rb
```diff
@@ -67,4 +68,38 @@
...
+     end
+
+     expect(page).to have_css('.appliedFilter', text: /kittens/)
+
+     find(:css, 'a.remove').trigger('click')
+     expect(page).not_to have_css('.appliedFilter', text: /kittens/)
+     expect(current_url).not_to match(%r{/article\?.*&q=kittens})
+   end
+ end
 end
```
jkeck: There seems to be some odd indentation going on w/ this `scenario`.

spec/views/article/index.html.erb_spec.rb
```diff
@@ -2,5 +2,28 @@
 
 RSpec.describe 'article/index.html.erb' do
   before { render }
-  it 'runs a search'
+
+  xit 'home page' do
```
jkeck: Are we able to unskip all these tests now?

camillevilla:
@jkeck I had to add a copy of  `has_search_parameters?` to `ArticleHelper` to get the view tests to pass. But hopefully this one is good to go 🤞

app/controllers/article_controller.rb
```diff
@@ -186,4 +186,8 @@ def setup_eds_session(session)
       debug: Settings.EDS_DEBUG
     ).session_token
   end
+
+  def has_search_parameters?
```
jkeck: This method is added both here in the controller and in the helper.

camillevilla: @jkeck I'm wasn't really sure where to put this one. In Blacklight, the method lives in a concern [catalog.rb](https://github.com/projectblacklight/blacklight/blob/master/app/controllers/concerns/blacklight/catalog.rb#L112). The app functions when the method is only in the controller. But when I run the view tests, I get the following error:

```
  1) article/index.html.erb home page
     Failure/Error: <% if has_search_parameters? %>

     ActionView::Template::Error:
       super: no superclass method `has_search_parameters?' for #<#<Class:0x007fd2ff512488>:0x007fd3001dfaa8>
       Did you mean?  has_range_limit_parameters?
     # ./app/views/article/index.html.erb:1:in `_app_views_article_index_html_erb___3943846516671069103_70272113930040'
     # ./spec/views/article/index.html.erb_spec.rb:4:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   super: no superclass method `has_search_parameters?' for #<#<Class:0x007fd2ff512488>:0x007fd3001dfaa8>
     #   Did you mean?  has_range_limit_parameters?
     #   ./app/views/article/index.html.erb:1:in `_app_views_article_index_html_erb___3943846516671069103_70272113930040'
```

jkeck: @camillevilla hmm, what if you add `helper_method :has_search_parameters?` beneath the method definition in the controller and remove the helper?  That might get it working.

camillevilla: @jkeck it's working! weirdly enough, I had to add it towards the top of the controller. Adding it below the method didn't work for some reason 🤔

Issue #1409, opened 2017-06-28
Add 2-column layout to record view


jvine:
Basic layout is the same as SearchWorks catalog.

- [x] record nav bar is above the title
- [x] title is full width
- [x] below the title, two columns are split col-md-4 and col-md-8
- [x] left column has the Online panel
- [x] styling of Online panel should be same as SW catalog 
- [x] right column has metadata
- [x] Online panel lists the full-text links (see #1399)

![screen shot 2017-06-28 at 11 22 14 am](https://user-images.githubusercontent.com/6945691/27653712-cc48f746-5bf4-11e7-8fcd-ef9f40b434c8.png)

PR #1411, opened 2017-06-28
Show fulltext links on results and show pages

### LINKED ISSUE #1399: Link to full-text from EDS results
Results view:
List all the returned full text links with label as link text. 
No icons. 
If there's no label, use the URL as link text.
Markup = `<ul><li>` for each link
Last element within in the document

Record view:
We will want the same list on the record view, in the "Online" box in the left sidebar. Since this has bigger layout implications for the record view, this might break out into a separate ticket.

drh-stanford:
This PR fixes #1399.

### On shows page
screen shot 2017-06-28 at 1 19 18 pm

### On results page
screen shot 2017-06-28 at 1 19 32 pm

### DIFF
--- a/app/assets/stylesheets/modules/article.scss
+++ b/app/assets/stylesheets/modules/article.scss
@@ -9,3 +9,7 @@
     padding-top: 10px;
   }
 }
+
+li.article-fulltext-link {
+  list-style-type: none;
+}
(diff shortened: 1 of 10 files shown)


drh-stanford:
@jkeck Thanks. I think I've addressed your feedback. I'm not sure I'm using the Presenter pattern correctly. Is there a way to link an `add_show_field` to a presenter method? I didn't see one so I changed the show view to call it.

jkeck:
@drh-stanford unfortunately there is not.  I believe there is an `accessor` option, but that's about an accessor on the document not a presenter.

I think this will ultimately be okay though since we most likely won't be using `add_show_field` when we complete work for #1409.

PR #1418, opened 2017-06-29
persist the EDS session token between Blacklight sessions

### LINKED ISSUE #1412: Persist the EDS session token between searches
The EDS API will fail when asking for the 251st result on a new session. We're currently issuing all searches as a new-session. Persisting the session token should improve performance too.

drh-stanford:
This PR fixes #1412.

Verified that it's working correctly with Bill and Eric on Slack. Basically you can look at the Faraday logs to see the API traffic (set EDS_DEBUG=true in your setteings). It should not issue a CreateSession call unless you clear your browser cookies.

### DIFF
--- a/app/controllers/article_controller.rb
+++ b/app/controllers/article_controller.rb
@@ -10,7 +10,6 @@ class ArticleController < ApplicationController
   before_action :eds_init, only: %i[index show]
   # TODO: probably need to move this into an Eds::SearchService initializer
   def eds_init
-    session['guest'] = true # TODO: hardcoded to non-authenticated
     setup_eds_session(session)
   end
 
@@ -74,7 +73,7 @@ class ArticleController < ApplicationController
 
   def search_service
     eds_params = {
-      'guest' => session['guest'],
+      'guest' => true, # TODO: hardcoded to non-authenticated
       'session_token' => session['eds_session_token']
     }
     Eds::SearchService.new(blacklight_config, search_state.to_h, eds_params)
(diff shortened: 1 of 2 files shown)


app/controllers/article_controller.rb
```diff
@@ -91,4 +83,18 @@ def search_service
   def set_search_query_modifier
     @search_modifier ||= SearchQueryModifier.new(params, blacklight_config)
   end
+
+  # Reuse the EDS session token if available in the user's session data,
+  # otherwise establish a session
+  def setup_eds_session(session)
+    return if session['eds_session_token'].present?
+    session['eds_session_token'] = EBSCO::EDS::Session.new(
+      guest: session['guest'],
```
jkeck: Hmm, this has me thinking a bit.  The only time that the `guest` status is going to change is when the user logs in or out.  At log-in time, we'll need to determine if the user should be in guest mode or not.  I guess I'm just a bit worried about storing the `guest` status in the session w/o a good handle on how that session variable might get set/unset.

What would you think about just hard-coding this to be `true`, removing the `guest` session variable, and hammering out the implementation details when we need to implement non-guest sessions?

cbeer: There's also some callback in `devise-guests` when a guest user logs in you might use to update the session. `logging_in_user`, maybe?

drh-stanford: This should be addressed in 69d2357.

Issue #1441, opened 2017-07-05
Audit configured metadata from EDS


jkeck:
Are we showing all available metadata on the record view? Check the code vs. what's configured in our controller.

jkeck:
Here are the available fields as of this comment (`0.1.8.pre` version of `ebsco-eds`):

I will submit a PR shortly to add additional fields to the record display as requested by @jvine.

## Already present (or being added as part of #1414):
* id
* eds_fulltext_links
* eds_document_doi
* eds_publication_date
* eds_physical_description
* eds_publication_type
* eds_languages
* eds_subjects
* eds_abstract
* eds_authors
* eds_title
* eds_source_title
* eds_database_name
* eds_publication_info
* eds_issns
* eds_isbns
* eds_series
* eds_volume
* eds_issue
* eds_page_count
* eds_page_start

## Currently missing
* eds_accession_number
* eds_database_id
* eds_access_level
* eds_relevancy_score
* eds_composed_title
* eds_other_titles
* eds_author_affiliations
* eds_subjects_geographic
* eds_subjects_person
* eds_author_supplied_keywords
* eds_notes
* eds_publication_type_id
* eds_publication_year
* eds_document_type
* eds_document_oclc
* eds_issn_print
* eds_isbn_print
* eds_isbns_related
* eds_isbn_electronic
* eds_covers
* eds_cover_thumb_url
* eds_cover_medium_url
* eds_fulltext_word_count
* eds_result_id
* eds_plink
* eds_html_fulltext
* eds_images
* eds_all_links
* eds_non_fulltext_links
* eds_publication_id
* eds_publication_is_searchable
* eds_publication_scope_note

PR #1450, opened 2017-07-06
reorganize show record metadata into sections

### LINKED ISSUE #1414: Break metadata into sections on the record view page
Goal is to break metadata into predictable sections that can be easily navigated to. (Will make separate ticket for jump menu.)

- [x] remove the duplicate title from the metadata list
- [x] break multiple values into separate lines (multiple `<dd>`s per `<dt>`)
- [x] with the exception of the top section (authors first) the order of the information in each section is not super important
- [x] headings are `<h2>`

Top section (no visible heading)
- [x] authors
- [x] source info (journal title, publisher etc.)
- [x] date
- [x] language

Abstract/Summary
- [x] abstract
- [ ] summary
- [ ] description

Subjects
- [x] subjects
- [x] geographic subjects
- [x] author keywords (no data as yet)
- [ ] descriptors (here's an example record: https://stanford.idm.oclc.org/login?url=http://search.ebscohost.com/login.aspx?direct=true&db=agr&AN=IND605263032&site=eds-live)
- [ ] categories (eample: https://stanford.idm.oclc.org/login?url=http://search.ebscohost.com/login.aspx?direct=true&db=nlebk&AN=1491977&site=eds-live)

Details
- [x] everything else
- [x] note that Database and Format have been moved down to this section from the top

### Before
screen shot 2017-06-29 at 10 34 44 am

### After
screen shot 2017-06-29 at 11 03 04 am

drh-stanford:
This PR fixes #1414 and is connected to #1441. It uses a section-based configuration (see `config.show.sections` which keeps all the configuration in ArticleController) and rewrites the `_show_default` partial.

Also upgrades to EDS 0.1.9pre for a dependency on multivalued authors.

Note there's no unit tests (pending EDS fixtures). It's a bit difficult to test this manually as finding example records in the wild isn't particularly effective.

### Typical record view

screen shot 2017-07-07 at 9 51 32 am

### Record view with missing Subjects section (as expected due to insufficient metadata)

screen shot 2017-07-07 at 9 51 59 am


app/views/article/record/_section_fields.html.erb
```diff
@@ -0,0 +1,14 @@
...
+fields.each do |field_name, label|
+  if doc_presenter.respond_to?(field_name)
+    value = doc_presenter.send(field_name)
+  else
+    value = document[field_name.to_s]
+  end
+ -%>
+  <% if value.present? %>
+    <dt><%= label -%></dt>
+    <dd><%= safe_join(Array.wrap(value)) -%></dd>
```
jkeck: This is already being `safe_joined` in the presenter I think?  Do we need it both places?

app/controllers/article_controller.rb
```diff
@@ -70,30 +70,43 @@ def eds_init
...
+        eds_issue: 'Issue',
+        eds_page_start: 'Page Start',
+        eds_page_count: 'Page Count',
+        eds_isbns: 'ISBN',
+        eds_issns: 'ISSN',
+        eds_publication_info: 'Published',
+        eds_document_oclc: 'OCLC',
+        eds_document_type: 'Document Type',
+        eds_other_titles: 'Other Titles',
+        eds_physical_description: 'Physical Description'
```
jkeck: One general concern I have about this approach is that we no longer have the ability to use configurations that we might want/need down the road (e.g. the `helper_method` that was removed, linking to searches, etc).

It's tough to say now if we'll need to break completely out of the BL configurations in order to derive our customized display, but as-is, this might make it harder to do something along the lines of link authors or subjects to searches (unless I'm missing something about this approach that would allow that).

One alternate approach might be to still configure all the show fields via `add_show_field` (allowing us to configure `heper_methods`, etc) but loop through a sections configuration in the UI can call `doc_presenter.field_value('field_name')` to continue to take advantage of the BL rendering pipeline.  This will do things like handle the `render_multivalued` that we've added to the presenter, and we can configure a helper to deal w/ the fields that we know to contain html (which is currently being handled by the `render_text_from_html` presenter method).

Happy to chat more about potential alternate approaches.

drh-stanford: Thanks for the feedback. I think I've got something working now that keeps the `add_show_field` and feeds through the default presenter. The only concern I have is that rather than multiple values showing up on different lines, they are "and"ed together like so:

screen shot 2017-07-06 at 2 56 02 pm

maybe there's a way to fix that in the rendering pipeline? Previously I was using the presenter to do a join with `'</dd><dd>'`...

jvine:
@drh-stanford now that I see more metadata - the Composed title is preferable to the Journal - can we switch their positions?

drh-stanford:
@jkeck I just did one last bit which was to add line breaks to the Authors and Subjects. I updated the screenshots and I removed the WIP -- it's ready to ship.

PR #1451, opened 2017-07-06
Add a helper for stubbing article search responses

jkeck:
@drh-stanford && @camillevilla interested on your input on this approach.

### DIFF
--- a/spec/features/article_searching_spec.rb
+++ b/spec/features/article_searching_spec.rb
@@ -30,7 +30,7 @@ feature 'Article Searching' do
   end
 
   scenario 'when searching w/i the article context we stay in the articles controller' do
-    skip 'need to stub out Eds::SearchService'
+    stub_article_service(docs: [SolrDocument.new(id: 'abc123')])
     visit article_home_path
 
     within '.search-form' do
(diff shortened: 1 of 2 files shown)


spec/support/stub_article_service.rb
```diff
@@ -0,0 +1,59 @@
...
+      raise "Single document response requsted but #{docs.length} provided." if docs.many?
+      expect_any_instance_of(Eds::SearchService).to receive(:fetch).and_return(
+        [StubAticleResponse.new(docs.first), docs.first]
+      )
+    else
+      raise "Unknown article stub type #{type} provided."
+    end
+  end
+
+  class StubAticleResponse < Blacklight::Solr::Response
```
jkeck: This class is more verbose than I was hoping initially.  I figure we can extend/modify this moving forward as the need presents itself.

drh-stanford: Typo here: `StubAticleResponse` should be `StubArticleResponse`? You have it spelled that way in various places.

spec/features/article_searching_spec.rb
```diff
@@ -39,6 +39,28 @@
...
     end
 
+    expect(page).to have_css('h2', text: /\d+ results?/)
+
     expect(current_url).to match(%r{/article\?.*&q=Kittens})
   end
+
+  scenario 'article records are navigable from search results' do
+    stub_article_service(docs: [SolrDocument.new(id: 'abc123', eds_title: 'The title of the document')])
+    stub_article_service(type: :single, docs: [SolrDocument.new(id: 'abc123', eds_title: 'The title of the document')])
```
drh-stanford: I'm not sure I follow this. Why are you calling `stub_article_service` twice?

jkeck: The first is for the search results and the second is for the document view (hence the `type: :single`) because the search service either received `search_results` or `fetch`.

I'll update the SolrDocuments that I'm using to stub so that is more obvious.

spec/support/stub_article_service.rb
jkeck: `facepalm.gif`

PR #1454, opened 2017-07-08
[BLOCKED] Attempt at getting article search facet limiting working.

jkeck:
_**[BLOCKED]** by ebsco/edsapi-ruby#10_

## No limit set
_This is the end of a very long list of facets_

---

end-of-unlimited-lang-facet

## Limit set to 10
_This is the end of a shorter, but still pretty long list of facets_

---

end-of-liited-lang-facet


jkeck:
This isn't possible via EDS at this time, we'll need to pick this back up once it is.

PR #1455, opened 2017-07-08
Update PhantomJS version on Travis

### LINKED ISSUE #1402: Warning message on Travis about using old PhantomJS version
Our Travis builds are warning that we're using an old version of PhantomJS:

```
You're running an old version of PhantomJS, update to >= 2.1.1 for a better experience.
```

camillevilla:
Closes #1402 and updates Travis version to 2.1.1

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -12,7 +12,7 @@ Capybara.register_driver :poltergeist do |app|
 end
 Capybara.javascript_driver = :poltergeist
 
-Capybara.default_max_wait_time = 10
+Capybara.default_max_wait_time = ENV["CI"] ? 30 : 10
 
 if ENV["COVERAGE"] or ENV["CI"]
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -12,7 +12,7 @@
 end
 Capybara.javascript_driver = :poltergeist
 
-Capybara.default_max_wait_time = 10
+Capybara.default_max_wait_time = 30
```
jkeck: Were you seeing issues locally w/ this wait time?  Just wondering if we should only set it to this under CI (so that it fails faster locally).

Thoughts?

camillevilla: I'm still getting the occasional timeouts and Poltergeist problems locally, but it's not as bad as under CI. I added a ternary operator for now and will keep an eye on tests failing locally.

PR #1457, opened 2017-07-10
Restyling of record toolbar and add results navigation to the show page

### LINKED ISSUE #1417: Add record navigation and tools to the top of the record view
All behaviour on this toolbar should be the same as SW catalog. Any behaviour and styling changes here are intended to affect catalog as well.

- if there is an active search result, show the left-side nav controls
screen shot 2017-06-29 at 12 36 10 pm

- if no search result, show only the right-side tools
screen shot 2017-06-29 at 12 38 31 pm

- if it's the first record in the result, hide the "previous" control (same for "next" on last record)
screen shot 2017-06-29 at 12 40 43 pm

- "Cite" should be present only if EDS can provide a formatted citation

## Styling
Heads-up: This is going to be tweaky, as there is some inconsistency and hacked styling in the existing controls. Note that the results nav/toolbar is intended to have the same styling, as there are some shared settings. 

- background color:  #797268
- drop shadow: same as grey navbar
- border: 0
- text color: white
- font size: 15px

- hover same as grey navbar
screen shot 2017-06-29 at 12 52 22 pm

- focus same as grey navbar
screen shot 2017-06-29 at 12 58 28 pm

drh-stanford:
This PR fixes #1417.  Note that it updates the styling for both catalog and article search.

Note these screenshots are from catalog search.

### Default show page (no search)

screen shot 2017-07-10 at 10 42 13 am

### Record is the first hit of the search

screen shot 2017-07-10 at 10 43 04 am

### Record is the last hit of the search

screen shot 2017-07-10 at 10 42 50 am

### Record is in the middle of the search

screen shot 2017-07-10 at 2 19 21 pm


drh-stanford:
@jkeck Here's a video of the strange behavior I'm seeing between sessions. It looks like the article session searches in the Search table aren't working correctly.

In the video, I search for `bear` then look at the second hit, then in a new window I search for `boar` then look at the second hit. Then if I refresh the first window, I get popped into the `boar` search results.

bearboar mov

drh-stanford:
So that refresh issue in the video above is an existing bug in production, so I'm going to break it out into its own ticket for later.

jkeck:
@drh-stanford I believe this isn't necessarily a "bug" per-se, but a side-effect of how the Blacklight feature works (in other words, I'm not sure we're likely to do anything about this in SearchWorks).

drh-stanford:
rebased onto newer article-search

drh-stanford:
I've fixed that border-bottom in d7f56bf. It was getting set by Blacklight.

drh-stanford:
Ack... it's happening in 2 places -- search results and record show. One is Blacklight the other is Bootstrap. So, it's still need a fix.

PR #1477, opened 2017-07-13
Unlink subject terms in article search results

### LINKED ISSUE #1473: Unlink subject terms on search results

camillevilla:
Closes #1473

This PR
- adds `eds_subjects` to our first stubbed article
- only unlinks subject terms in the search results
- reorganizes tests in `article_searching_spec.rb`

## Before

linked_subjects

## After

unlinked_subjects

### DIFF
--- a/app/controllers/article_controller.rb
+++ b/app/controllers/article_controller.rb
@@ -30,22 +30,10 @@ class ArticleController < ApplicationController
     config.index.display_type_field = 'eds_publication_type'
     config.index.fulltext_links_field = 'eds_fulltext_links'
 
-    # Configured index fields not used
-    # config.add_index_field 'author_display', label: 'Author'
-    # config.add_index_field 'id'
-
-    # Summary
     config.add_index_field "eds_authors", label: 'Authors'
-    config.add_index_field "eds_author_affiliations", label: 'Author Affiliations'
     config.add_index_field "eds_composed_title", label: 'Composed Title', helper_method: :strip_html_from_solr_field
-    config.add_index_field "eds_publication_date", label: 'Publication Date'
-    config.add_index_field "eds_languages", label: 'Language'
-
-    # Subjects
     config.add_index_field "eds_subjects", label: 'Subjects'
-    config.add_index_field "eds_subjects_geographic", label: 'Geography'
-    config.add_index_field "eds_subjects_person", label: 'Person Subjects'
-    config.add_index_field "eds_author_supplied_keywords", label: 'Author Supplied Keywords'
+    config.add_index_field "eds_abstract", label: 'Abstract', helper_method: :strip_html_from_solr_field
 
     config.add_search_field('search') do |field|
       field.label = 'All fields'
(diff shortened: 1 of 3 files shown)


app/helpers/article_helper.rb
```diff
@@ -15,10 +15,14 @@ def link_subjects(options = {})
     return unless options[:value]
     separators = options.dig(:config, :separator_options) || {}
     values = render_text_from_html(options[:value])
-    values.collect do |value|
+    if params[:action] == 'index'
```
jkeck: Hmm, this feels a bit wrong.  Perhaps we can discuss alternate solutions at some point today.

drh-stanford: I think you can implement this within ArticleController by omitting the helper method `link_subjects` for the index pages. If you look in there, check out how we ignore the `separator_options` for index pages -- see https://github.com/sul-dlss/SearchWorks/blob/article-search/app/controllers/article_controller.rb#L121

camillevilla:
@drh-stanford thanks for the tip. @jkeck here's another stab. It still doesn't feel quite right to except ALL possible helper methods for Subjects.

app/controllers/article_controller.rb
```diff
@@ -118,7 +118,9 @@ def eds_init
     config.show.sections.each do |_section, fields|
       fields.each do |field, options|
         config.add_show_field field.to_s, options
-        config.add_index_field field.to_s, options.except(:separator_options)
+        exceptions = :separator_options
+        if _section == 'Subjects' then exceptions = %i[separator_options helper_method] end
```
drh-stanford: This looks good to me. The only comment here is that `"_"` as a prefix to a variable name means that the variable is unused (i.e., rubocop will insist that you use underscores to mark unused variables). So we can rename `_section` to `section`.

camillevilla:
Thanks @drh! I renamed the variable and squashed my commits. @jkeck let me know if you have any additional feedback.

jkeck:
I guess my concern lies in why are we making an exception in the method at all?  We only display 5 fields in the results, why don't we just configure those using the standard Blacklight method for each of the fields?

It feels like when this was originally done it was trying to DRY up the controller configurations prematurely.  Now that we have the same field displayed in both places, but the desire to configure them differently, perhaps it would be better to un-DRY this as opposed to pushing the differences into logic/code.

drh-stanford:
We can explicitly do the `add_index_field` calls to match what's in `_index_default.html.erb` and `_index_brief_default.html.erb`, instead of putting that `add_index_field` calls in that loop. That's fine with me.

camillevilla:
🤞 @jkeck @drh-stanford here's a version that uses `add_index_field`.

app/controllers/article_controller.rb
```diff
@@ -114,14 +127,6 @@ def eds_init
       }
     }
 
-    # Register section fields with show/index presenters to leverage rendering pipeline
-    config.show.sections.each do |_section, fields|
```
jkeck: We'll want to keep this loop, but just remove the `config.add_index_field` on line 121.

camillevilla:
@jkeck the loop has been (partially) restored.

PR #1478, opened 2017-07-13
Use ArticleSearchBuilder to pass queries onto EDS API

### LINKED ISSUE #1471: Previous / Next buttons not working with EDS gem
The Blacklight way of defining the Previous/Next buttons isn't working with the EDS gem. See slack for a screencast and behavior. Blocked on EDS gem debugging.

drh-stanford:
This PR is connected to #1471.

I refactored Eds::SearchService/Repository which allows us to use our own `ArticleSearchBuilder` rather than the existing Catalog's `SearchBuilder` (which has a processing chain that we shouldn't use). 

The core issue is that the Previous/Next buttons weren't working on the show page (and I also believe on the index page too), so this PR mainly cleans up the code path to reduce the potential for problems. I've added support for the `track` action on Articles in the routing, and updated the `id` constraints to match the EDS document id rules.

The refactoring also allows us to use the Blacklight `index`, `show`, and `track` methods as-is in `ArticleController` rather than define them ourselves (and lose features like the `.json` API).

For `link_to_previous_document` and `link_to_next_document`, we don't need the `.disabled` class anymore, so I've re-copied the Blacklight version of these methods and just changed the CSS class we use (it'd be nice to have that be a parameter).

Finally, I added some more configuration values for the EDS gem (timeouts, cache, logging).

Note that there's still a problem with the Prev/Next buttons as reported in #1471 which is why this PR does not close it -- that ticket is blocked on an EDS ticket right now. But this PR is good to go.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,7 +20,6 @@
 # Ignore all logfiles and tempfiles.
 /log/*.log
 /tmp
-faraday.log
 
 config/settings.local.yml
 config/settings/*.local.yml
--- a/app/controllers/article_controller.rb
+++ b/app/controllers/article_controller.rb
@@ -22,6 +22,7 @@ class ArticleController < ApplicationController
   configure_blacklight do |config|
     # Class for sending and receiving requests from a search index
     config.repository_class = Eds::Repository
+    config.search_builder_class = ArticleSearchBuilder
 
     # solr field configuration for search results/index views
     config.index.document_presenter_class = IndexDocumentPresenter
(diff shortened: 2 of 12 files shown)


drh-stanford:
@jkeck any idea why these changes cause the tracking URL in the document links on the search results page to have a `per_page=3` in it? 

`/catalog/bth__102714049/track?counter=3&amp;per_page=3&amp;search_id=1`

This isn't happening on the sandbox site so it's something in this set of changes :(.

PR #1497, opened 2017-07-19
prioritize and relabel full text links

### LINKED ISSUE #1476: Linking to full text
There are 4 possible link types:

1. HTML full text (we're not getting these yet)
   - reword as "View full text"
   - goes to record view, which will include full text 
2. PDF full text
   - "View/download full text PDF"
   - default browser behaviour for PDF
3. Check SFX for full text
   - "View full text on content provider's site" 
   - add an "external link" icon
   - SFX will use direct linking to bypass the menu and send the user to the best source
4. View request options
   - "Find it in print or via interlibrary services"
   - user ends up at the SFX menu

Link types we won't use:

- "Access URL" and "Availability" - these appear to be duplicative of SFX links 
- custom links, such as "View Record from Oxford University Press" - we will turn off custom linking in the EDS Admin, which will turn "View Record from.." links into "Check SFX.." links.

Link types are prioritized as follows:

- 1 and 2 are preferred, and can coexist
- show 3 only if there's no 1 or 2
- show 4 only if there's no 1-3

Before:
screen shot 2017-07-12 at 4 24 46 pm

After:
screen shot 2017-07-12 at 4 22 39 pm

Before:
screen shot 2017-07-12 at 4 25 52 pm

After:
screen shot 2017-07-12 at 4 23 26 pm

Fancier styling will be added later.

drh-stanford:
This PR is connected to #1476. It rewrites the `PresenterFulltextLinks.fulltext_links` to support re-labeling and prioritization.

### Example index page

screen shot 2017-07-19 at 11 54 18 am

### Example show page
screen shot 2017-07-19 at 11 53 25 am

### DIFF
--- a/app/presenters/concerns/presenter_fulltext_links.rb
+++ b/app/presenters/concerns/presenter_fulltext_links.rb
@@ -1,8 +1,58 @@
 module PresenterFulltextLinks
   def fulltext_links
-    field_key = configuration.index.fulltext_links_field
-    document[field_key].collect do |link|
-      "<li class=\"article-fulltext-link\">#{view_context.link_to(link['label'] || link['url'], link['url'])}</li>".html_safe
+    links = document[configuration.index.fulltext_links_field]
+    links.collect do |link|
+      if show_link?(links, link)
+        link = relabel_link(link)
+        "<li class=\"article-fulltext-link\">#{view_context.link_to(link['label'], link['url'])}</li>".html_safe
+      end
+    end.compact
+  end
+
+  private
+
+  def relabel_link(link)
+    link = link.dup
+    map = to_link_mapping(link)
+    link['label'] = map[:label] if map.present?
+    link
+  end
+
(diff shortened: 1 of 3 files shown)


app/presenters/concerns/presenter_fulltext_links.rb
```diff
@@ -1,8 +1,60 @@
...
+    return false if map.blank? # drop links that don't have a mapping
+
+    case map[:priority]
+    when 1, 2
+      true
+    when 3
+      !link_within_priority?(links, 2)
+    when 4
+      !link_within_priority?(links, 3)
     end
```
camillevilla: @drh not sure if this is nitpicky, but could we do something like:
```
when 3, 4
!link_within_priority?(links, map[:priority] - 1) 
```

camillevilla:
@drh I have a small comment on the case statement, but otherwise this looks good to me.

PR #1517, opened 2017-07-25
Add article HTML full text hide/show in record view

### LINKED ISSUE #1509: Add hide / show toggle for article full text
Split off from #1482 / PR #1505

When there is HTML full text for an article, display it on the record page in a show/hide section. 

- When user clicks the article title in the results, show the metadata, with the full text hidden. 
- When user clicks the "full text" link in the results, show the full text expanded.

28193564-9cc8eff8-67f2-11e7-80c6-43912b91c882

camillevilla:
Closes #1509 

This PR adds hide/show functionality for HTML fulltext in the article record view:
- Hides full text by default
- Clicking on the section header (green bar) toggles fulltext
- `View full text / Show full text` does not appear in records where HTML full text is unavailable
- adds a concern for `EdsDocument`

## Demo
hide-show-fulltext

### DIFF
--- a/app/assets/stylesheets/modules/record-metadata-section.scss
+++ b/app/assets/stylesheets/modules/record-metadata-section.scss
@@ -1,9 +1,27 @@
 .record-sections {
+  .section {
+    padding-bottom: 0;
+  }
+}
+
+.record-metadata, .upper-record-metadata {
+  @include show-page-row-layout;
+  dt {
+    width: auto;
+  }
+}
+
+.article-record-sections {
+  .fulltext-toggle-bar {
+    border: 0;
+    text-align: left;
+    width: 100%;
+  }
+
   .section-container-heading {
-    background-color: #009B76;
-    border-bottom: 1px solid $sul-panel-border-bottom;
-    margin-top: 10px;
+    background-color: $pantone-334;
     margin-bottom: 15px;
+    margin-top: 10px;
     padding: 10px 15px;
 
     h2 {
@@ -14,27 +32,6 @@
       font-weight: 400;
     }
   }
-  .section {
-    padding-bottom: 0;
-
-    .section-heading h2 {
-      i.fa {
-        font-size: 15px;
-        margin-right: 5px;
-      }
-    }
-  }
-}
-
-.record-metadata, .upper-record-metadata {
-  @include show-page-row-layout;
-
-  dt {
-    width: auto;
-  }
-}
-
-.article-record-sections {
   dt {
     text-align: right;
   }
--- a/app/views/article/record/_html_fulltext.erb
+++ b/app/views/article/record/_html_fulltext.erb
@@ -1,0 +1,13 @@
+<% doc_presenter = presenter(document) -%>
+<button class="section-container-heading fulltext-toggle-bar" id="fulltextToggleBar" data-toggle="collapse" href="#toggleFulltext" aria-expanded="false" aria-controls="toggleFulltext">
+  <h2>Full text</h2>
+</button>
+<div class='section-body collapse' id="toggleFulltext">
+<% fields.each do |field_name| -%>
+  <% field_name = field_name.to_s -%>
+  <% value = doc_presenter.field_value field_name -%>
+  <% if value.present? %>
+    <div class="blacklight-<%= field_name.parameterize %>"><%= value %></div>
+  <% end -%>
+<% end -%>
+</div>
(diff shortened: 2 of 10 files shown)


jvine:
@camillevilla I like the clickable bar! Nice. The animation looks good but I'd like to see the HTML in action - I can't quite tell from the code what's happening w ARIA etc. Screenhero at some point? 

Couple of clean-up bits for the display:

- Change Full Text -> Full text  
(SUL style is sentence case)
- remove the `<dl>` formatting for the full text section, so it doesn't have the redundant label to the left.

Current
screen shot 2017-07-25 at 10 22 42 am

Mockup
screen shot 2017-07-25 at 10 25 34 am

camillevilla:
hi @jvine! I've made the following updates:
- moved fulltext out of `<dl>`
- updated section header to sentence case
- renamed our div ID / `aria-control` to `toggleFulltext`

camillevilla:
@jkeck or @drh-stanford just did a walkthrough with @jvine and got the thumbs up. I had to

app/helpers/article_helper.rb
```diff
@@ -11,6 +11,10 @@ def article_restricted?(document)
     document['eds_title'] =~ /^This title is unavailable for guests, please login to see more information./
   end
 
+  def html_fulltext_available?(document)
+    !document['eds_html_fulltext'].blank?
```
jkeck: What do you think about this a method on the `SolrDocument` class instead of a helper that gets passed an instance of a document?

app/views/article/record/_html_fulltext.erb
```diff
@@ -0,0 +1,13 @@
+<% doc_presenter = presenter(document) -%>
+<div class="section-container-heading" data-toggle="collapse" href="#toggleFulltext" aria-expanded="false" aria-controls="toggleFulltext">
+  <h2>Full text</h2>
+</div>
```
jkeck: I haven't checked this branch out yet (in the middle of a bunch of changes in my working branch at the moment), but a couple questions I have:

Is this `div` keyboard accessible (e.g. can you tab to it and hit enter to open/close)?  Does the cursor indicate that the heading element is clickable (it didn't look like it from the gif, but can be hard to tell).

camillevilla: @jvine and I talked about this a little bit yesterday. The `div` itself is not tab-able, but the second link (Hide / View full text) is.

The cursor does not indicate the `div` is clickable. Maybe I could turn it into a giant button?

jvine:
Just to note that turning the toggle bar into a button will turn it into a duplicate control for keyboard and SR users, which is not necessarily the goal. I'm feeling like it might be better to turn the toggle into the sole control, and make it explicit. That would eliminate the sidebar box when there's no PDF version, but that's probably fine.

app/views/article/record/_html_fulltext.erb
```diff
@@ -0,0 +1,13 @@
+<% doc_presenter = presenter(document) -%>
+<button class="section-container-heading fulltext-toggle-bar" id="fulltextToggleBar" data-toggle="collapse" href="#toggleFulltext" aria-expanded="false" aria-controls="toggleFulltext">
+  <h2>Full text</h2>
+</button>
+<div class='section-body collapse' id="toggleFulltext">
+<% fields.each do |field_name| -%>
+  <% field_name = field_name.to_s -%>
+  <% value = doc_presenter.field_value field_name -%>
+  <% if value.present? %>
+    <dd class="blacklight-<%= field_name.parameterize %>"><%= value %></dd>
```
jkeck: It looks like this is putting the content in a `dd` w/o any of the other markup (parent `dl` and an accompanying `dt`).

Check out the [Permitted parents section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd) of this page.

I think we can just wrap this in a `div` and call it a day.

app/assets/stylesheets/modules/record-metadata-section.scss
```diff
@@ -1,40 +1,37 @@
...
 
 .article-record-sections {
+  .fulltext-toggle-bar {
+    border: 0;
+    text-align: left;
+    width: 100%;
+  }
+
+  .section-container-heading {
+    background-color: #009B76;
```
jkeck: Can we either use an existing scss variable or turn this into one?

camillevilla:
alrighty, @jkeck! Here are the requested changes.

PR #1527, opened 2017-07-27
add Research Starters to page 1 of results and as a show page

### LINKED ISSUE #1504: Show Research Starters in article results
screen shot 2017-07-20 at 3 47 20 pm

If there is a research starter returned for a search, show it on the first page of the search results, before the numbered results begin. 

If there is more than one, show only the first one.

- box style
     `border: 1px solid #8c1515`
     `box-shadow: 0 4px 8px -2px rgba(0,0,0,0.2);`
- close "x" positioned in upper right corner
   - add screen-reader text to the button: `<span class="sr-only">hide research starter</span>`
- Research Starter label is in css:
    ```
    .research-starter:before {
    position: absolute;
    top: 10px;
    left: 10px;
    font-size: 16px;
    font-weight: 400;
    color: #8c1515;
    text-transform: uppercase;
    letter-spacing: 1px;
    content: "Research Starter";
    ```
- `h3` is composed of
   - `<span class="sr-only">Research Starter: </span>
   - `<a>` with article title
- snippet of text
   - **need to determine where this comes from** - it should be the first ~3 lines of meaningful content in the article - i.e. not including the date updated. 
- image 
   - first image in the article, or whatever EDS provides as thumbnail
   - max-height 100px; width: auto; float: right; 

When the Research Starter box is closed:
screen shot 2017-07-20 at 3 55 03 pm

- hide all content
- rotate the "close" icon 45 degrees
- change its "sr-only" text to "show research starter"

### LINKED ISSUE #1518: Research Starter show page
Basic layout is the same as HTML full text, with these modifications:

1. "Research starter" label and composed title below the article title
2. additional sidebar panel below the "Available online" panel, with links to related research starters
3. if there are images in the full text, display them where they occur, at their default size (?) with their captions, styled in a box (like the media node on the library website, example http://library.stanford.edu/areas)
   ```
    padding: 12px;
    background-color: #fff;
    border: 1px solid #e3dfd5;
   ```
4. render the embedded composed links as actual links (to other research starters)

screen shot 2017-07-25 at 3 21 03 pm

drh-stanford:
This PR fixes #1504 and fixes #1518. 

- [x] need a bug fix from EDS for the search results Research Starter's abstract metadata (see https://github.com/ebsco/edsapi-ruby/issues/55) (fixed in 0.2.8.pre)
- [x] rebase onto article-search for new full text display changes
- [x] fix hardcoding in blacklight_helper.rb
- [x] addressed comments

Notes:
- If the user is not authenticated, then the show page will not have any full text content. We may need to include some indication for the user to login to see the research starter full text.

### On search page
screen shot 2017-07-28 at 10 41 24 am

### On show page (when user is authenticated)
screen shot 2017-07-28 at 10 41 45 am

### On show page (when user is not authenticated)
screen shot 2017-07-28 at 10 52 01 am

### DIFF
--- a/app/assets/stylesheets/modules/research-starter.scss
+++ b/app/assets/stylesheets/modules/research-starter.scss
@@ -4,14 +4,14 @@
   left: 10px;
   font-size: 16px;
   font-weight: 400;
-  color: #8c1515;
+  color: $sul-research-starter-color;
   text-transform: uppercase;
   letter-spacing: 1px;
   content: "Research Starter";
 }
 
 .research-starter {
-  border: 1px solid #8c1515;
+  border: 1px solid $sul-research-starter-color;
   box-shadow: 0 4px 8px -2px rgba(0,0,0,0.2);
 
   h3, p {
@@ -27,7 +27,7 @@
     margin-bottom: 10px;
   }
 
-  #research-starter-close-icon {
+  .research-starter-close-icon {
     border: none;
     position: relative;
     top: -15px;
@@ -35,7 +35,7 @@
     float: right;
     font-size: 2em;
     font-weight: 400;
-    color: #8c1515;
+    color: $sul-research-starter-color;
   }
 }
 
@@ -44,7 +44,7 @@
     margin-left: 10px;
     font-size: 16px;
     font-weight: 400;
-    color: #8c1515;
+    color: $sul-research-starter-color;
     text-transform: uppercase;
     letter-spacing: 1px;
   }
@@ -55,8 +55,8 @@
 
 .research-starter-figure {
   padding: 12px;
-  background-color: #fff;
-  border: 1px solid #e3dfd5;
+  background-color: $white;
+  border: 1px solid $beige-30-percent;
   margin-bottom: 20px;
 
   img {
--- a/app/assets/stylesheets/modules/research-starter.scss
+++ b/app/assets/stylesheets/modules/research-starter.scss
@@ -1,0 +1,72 @@
+.research-starter:before {
+  position: relative;
+  top: 10px;
+  left: 10px;
+  font-size: 16px;
+  font-weight: 400;
+  color: #8c1515;
+  text-transform: uppercase;
+  letter-spacing: 1px;
+  content: "Research Starter";
+}
+
+.research-starter {
+  border: 1px solid #8c1515;
+  box-shadow: 0 4px 8px -2px rgba(0,0,0,0.2);
+
+  h3, p {
+    margin: 10px 20px;
+  }
+
+  img {
+    max-height: 100px;
+    width: auto;
+    float: right;
+    margin-left: 10px;
+    margin-right: 10px;
+    margin-bottom: 10px;
+  }
+
+  #research-starter-close-icon {
+    border: none;
+    position: relative;
+    top: -15px;
+    right: 10px;
+    float: right;
+    font-size: 2em;
+    font-weight: 400;
+    color: #8c1515;
+  }
+}
+
+.research-starter-show {
+  .research-starter-label {
+    margin-left: 10px;
+    font-size: 16px;
+    font-weight: 400;
+    color: #8c1515;
+    text-transform: uppercase;
+    letter-spacing: 1px;
+  }
+  .research-starter-source-title {
+    margin-left: 30px;
+  }
+}
+
+.research-starter-figure {
+  padding: 12px;
+  background-color: #fff;
+  border: 1px solid #e3dfd5;
+  margin-bottom: 20px;
+
+  img {
+    max-height: 200px;
+    width: auto;
+  }
+
+  span {
+    position: absolute;
+    margin-right: 10px;
+    padding: 12px;
+  }
+}
--- a/app/helpers/article_helper.rb
+++ b/app/helpers/article_helper.rb
@@ -6,16 +6,6 @@ module ArticleHelper
     controller_name == 'article'
   end
 
-  def article_research_starter?(document)
-    # TODO: we probably need a better place to put this...
-    document['eds_database_name'] == 'Research Starters' if document
-  end
-
-  def article_restricted?(document)
-    # TODO: we probably need a better way to determine this
-    document['eds_title'] =~ /^This title is unavailable for guests, please login to see more information./
-  end
-
   def link_subjects(options = {})
     return unless options[:value]
     separators = options.dig(:config, :separator_options) || {}
@@ -59,7 +49,7 @@ module ArticleHelper
 
   def sanitize_fulltext(options = {})
     return unless options[:value].present?
-    return safe_join(options[:value]) if article_research_starter?(@document)
+    return safe_join(options[:value]) if @document && @document.research_starter?
     separators = options.dig(:config, :separator_options) || {}
     textblock = options[:value].map(&:to_s).to_sentence(separators)
     textblock = Nokogiri::HTML.fragment(CGI.unescapeHTML(textblock))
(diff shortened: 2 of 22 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -10,9 +10,11 @@ def document_partial_name(document, base_name = nil)
 
     display_type ||= document.display_type
 
+    display_type ||= document['eds_publication_type']
```
drh-stanford: We're overriding this method, so I'm not entirely sure what the best way to ensure that we get the display_type for the research starters. It's a different field than the display_type configuration in ArticleController.

app/assets/stylesheets/modules/research-starter.scss
```diff
@@ -0,0 +1,72 @@
...
+  img {
+    max-height: 100px;
+    width: auto;
+    float: right;
+    margin-left: 10px;
+    margin-right: 10px;
+    margin-bottom: 10px;
+  }
+
+  #research-starter-close-icon {
```
jcoyne: Styling on id is generally frowned upon. Can you add a class for this instead?

app/helpers/article_helper.rb
```diff
@@ -6,6 +6,11 @@ def article_search?
     controller_name == 'article'
   end
 
+  def article_research_starter?(document)
+    # TODO: we probably need a better place to put this...
```
jcoyne: Better place than `ArticleHelper`?

app/helpers/article_helper.rb
```diff
@@ -71,6 +77,37 @@ def render_text_from_html(values)
     end.flatten
   end
 
+  #
+  def transform_research_starter_text(options = {})
```
jcoyne: Could this be in it's own service class?

app/views/article/_show_header_reference.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render_document_heading(document, :tag => :h1) %>
```
jcoyne: why are you using ruby 1.8 hash syntax here and 1.9 syntax elsewhere?

spec/helpers/article_helper_spec.rb
```diff
@@ -124,4 +124,37 @@
       expect(result).to eq '<p>This Journal</p>, 10(1)'
     end
   end
+
+  context '#transform_research_starter_text' do
+    it 'removes anid and title' do
```
jcoyne: I'd write the subject of this context as:
```ruby
subject { helper.transform_research_starter_text(value: value) }
```
each subordinate context can set value:

```ruby
context "with EDS tags and regular HTML" do
  let(:value) { '<anid>abc</anid><title>123</title>' }
  it { is_expected.to eq '' }
end
```

app/assets/stylesheets/modules/research-starter.scss
```diff
@@ -0,0 +1,72 @@
+.research-starter:before {
+  position: relative;
+  top: 10px;
+  left: 10px;
+  font-size: 16px;
+  font-weight: 400;
+  color: #8c1515;
+  text-transform: uppercase;
+  letter-spacing: 1px;
+  content: "Research Starter";
```
jkeck: Any particular reason this is not in the markup itself?

app/helpers/article_helper.rb
```diff
@@ -6,6 +6,11 @@ def article_search?
     controller_name == 'article'
   end
 
+  def article_research_starter?(document)
+    # TODO: we probably need a better place to put this...
+    document['eds_database_name'] == 'Research Starters' if document
```
jkeck: I believe @camillevilla has started putting methods like this in an `EdsDocument` module mixed into the `SolrDocument` class.  That might be a good place for this sort of thing.

app/helpers/article_helper.rb
jkeck: I'm wondering if this might be useful to be in a presenter.  It will probably be a bit easier to reason about/test in a class as opposed to a helper method.

app/helpers/blacklight_helper.rb
jkeck: The display_type configured in the article controller is `eds_publication_type`, so I'm not quite sure I understand what this is trying to do.

app/views/article/_research_starters.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <div class='research-starter clearfix'>
+      <a class="research-starter-close" role="button" data-toggle="collapse" href="#research-starter-body" aria-expanded="true" aria-controls="research-starter-body">
+        <span id='research-starter-close-icon' class='fa fa-times-circle'></span>
+        <span class="sr-only">hide research starter</span>
+      </a>
+      <div id='research-starter-body' class='collapse.in'>
+        <h3>
+          <%= link_to starter['eds_title'], article_path(id: "#{starter['eds_database_id']}__#{starter['eds_accession_number']}") -%>
+        </h3>
+        <img src="<%= starter['eds_cover_thumb_url'] -%>"/>
```
jkeck: Are we guaranteed that the research starter will have a cover thumb?  Just want to make sure we don't unintentionally render a broken image here.

app/helpers/blacklight_helper.rb
drh-stanford: It's set to `format` in the show config -- see https://github.com/sul-dlss/SearchWorks/blob/article-search/app/controllers/article_controller.rb#L84

app/assets/stylesheets/modules/research-starter.scss
drh-stanford: This is how @jvine specified it. I assumed it was needed like this because it's in a `:before` block.

jvine: @jkeck, there was sr-only text specified separately in the ticket. That, combined with the css content, was the only way I could get the heading structure and the open/close behaviour to behave & look how I wanted.

app/helpers/blacklight_helper.rb
jkeck: Could we make that `eds_publication_type`?  The documents from EDS don't have a `format` field and we're not making use of that config anywhere for articles at the moment AFAIK.

app/assets/stylesheets/modules/research-starter.scss
jkeck: 👌

app/helpers/article_helper.rb
drh-stanford: moving into the ShowPresenter as per @jkeck comment

app/helpers/article_helper.rb
drh-stanford: moving into EdsDocument (SolrDocument) as per @jkeck comment

drh-stanford: moving into EdsDocument (SolrDocument) as per @jkeck comment

app/helpers/article_helper.rb
drh-stanford: moving into ShowPresenter as per @jkeck comment

app/helpers/blacklight_helper.rb
drh-stanford: @jkeck I tried doing this but the blacklight_helper.rb still needs modification, like so (adding 2nd line):

```
    display_type ||= document.display_type
    display_type ||= document[view_config.display_type_field]
    display_type ||= 'default'
```

and in ArticleController:

```
    config.show.display_type_field = 'eds_publication_type'
```

See https://github.com/projectblacklight/blacklight/blob/f0e814b1b738bd25d3734039a298eb725663386f/app/helpers/blacklight/render_partials_helper_behavior.rb#L112

Maybe the issue is a drift between BL 5.x and 6.x?

jkeck: 👍 Updating to make our override closer to BL's sounds good (if I'm understanding correctly)

PR #1528, opened 2017-07-28
Add SFX data into an Access Panel (loaded asynchronously)

### LINKED ISSUE #1486: Add SFX menu panel to left sidebar in record view
Background: We are prioritizing one source for the direct links on the results view. Expert recommendation when using direct linking is to also provide alternate access to the menu of available sources. 

- for items that have "link type 3" (see #1476), add a panel below the main "Available online" panel listing all the available full-text sources, and their coverage

screen shot 2017-07-14 at 4 23 35 pm 1

- information comes from SFX XML

```
<target>
   <target_public_name>Elsevier ScienceDirect Journals</target_public_name>
   <target_url>https://this_is_the_source_link</target_url>
   <coverage>
      <coverage_text>
         <threshold_text>
            <coverage_statement>From 1976, volume 1: issue 1</coverage_statement>
            <coverage_statement>...can repeat...</coverage_statement>
         </threshold_text>
      </coverage_text>
   </coverage>
</target>
```

- you get to the SFX XML by appending `&sfx.response_type=multi_obj_xml` to the SFX URL
http://sul-sfx.stanford.edu/sfxlcl41?url%5Fver=Z39.88-2004&ctx%5Fver=Z39.88-2004&ctx%5Fenc=info:ofi/enc:UTF-8&rfr%5Fid=info:sid/sfxit.com:opac%5F856&url%5Fctx%5Ffmt=info:ofi/fmt:kev:mtx:ctx&sfx.ignore%5Fdate%5Fthreshold=1&rft.object%5Fid=954925506289&svc%5Fval%5Ffmt=info:ofi/fmt:kev:mtx:sch%5Fsvc&sfx.response_type=multi_obj_xml

jkeck:
Closes #1486 

## edselp__S0376635716303229
sfx-panel-loading


app/models/concerns/eds_links.rb
```diff
@@ -0,0 +1,73 @@
...
+    def link_within_priority?(links, priority)
+      links.each do |link|
+        map = to_link_mapping(link)
+        return true if map.present? && map[:priority] <= priority
+      end
+      false
+    end
+
+    def sfx_link?(link)
+      (to_link_mapping(link) || {})[:priority] == 3
```
jkeck: I suppose we could either check the priority or our updated label.  Tests will fail if we change priority (which will at least let us know that we needed to update this method as well).

I'm on the fence.

app/models/sfx_data.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  private
+
+  attr_reader :base_sfx_url
+
+  def sfx_xml
+    return unless sfx_response.success?
+    @sfx_xml ||= begin
+      Nokogiri::XML.parse(sfx_response.body)
+    rescue
```
drh-stanford: Shouldn't this be `rescue => e` since you're using `e` below? Also, if you do go into this rescue block, is the return value `nil`?

app/models/concerns/eds_links.rb
drh-stanford: I think this is fine to use the priority. Maybe `priority` is really the wrong term -- maybe `category` is better?

app/views/article/_index_default.html.erb
```diff
@@ -18,8 +18,10 @@
     <li><%= doc_presenter.field_value('eds_abstract') %></li>
   </ul>
 <% end %>
-<% if document['eds_fulltext_links'].present? && !article_restricted?(document) %>
+<% if document.access_panels.online? && !article_restricted?(document) %>
   <ul class='document-metadata dl-horizontal dl-invert'>
-    <%= safe_join(doc_presenter.fulltext_links) %>
+    <% document.access_panels.online.links.each do |link| %>
+      <li><%= link_to(link.text, link.href) %></li>
```
drh-stanford: this change removes the class `article-fulltext-link` which is ok, but we can delete it from the CSS then.

app/models/sfx_data.rb
jkeck: Yes, there are two cases where `sfx_xml` can return `nil`, but the consumer of this method (the `targets` method) checks that and returns an sensical value (`[]`) in that case.

The other options would be to return an empty XML document, but that didn't quite feel right either.

jkeck:
@drh-stanford I believe I have addressed all your comments.  Please let me know if you notice anything else.

PR #1533, opened 2017-07-31
add label/icon to full text links

### LINKED ISSUE #1530: Add label or icon in front of links in search results
In search results only:

- For full text links (that is, all except the last type of link that says "Find it in print...") add this to the front of the link text, inside the `<a>`, and with a space following:
- `<span class="online-label">Full text</span> `

In search results and record view:

- For the last type (text label "Find it in print..."), add `class="sfx"` to the `<a>`.

screen shot 2017-07-28 at 10 08 23 am

screen shot 2017-07-28 at 10 05 40 am

drh-stanford:
This PR fixes #1530 by adding a label/icon to fulltext links on the search and show pages. It modifies the SearchWorks `Link` class to support "ILL" links. Also, permits PDF eBook links to show the same as regular PDF links.

### SFX link in default search results
screen shot 2017-07-31 at 12 32 22 pm

### Full text link in default search results
screen shot 2017-07-31 at 12 32 37 pm

### SFX link in brief search results
screen shot 2017-07-31 at 12 34 55 pm

### Full text link in brief search results (PDF/eBook link)
screen shot 2017-07-31 at 12 35 02 pm

### SFX link on show page
screen shot 2017-07-31 at 12 37 07 pm

### Full text link on show page (no change)
screen shot 2017-07-31 at 12 37 23 pm

### DIFF
--- a/app/views/article/_index_default.html.erb
+++ b/app/views/article/_index_default.html.erb
@@ -34,7 +34,14 @@
 <% if document.access_panels.online? && !document.eds_restricted? %>
   <ul class='document-metadata dl-horizontal dl-invert'>
     <% document.access_panels.online.links.each do |link| %>
-      <li><%= link_to(link.text, link.href) %></li>
+      <li>
+        <% if link.ill? -%>
+          <%= link_to(link.text, link.href, class: 'sfx') %>
+        <% else %>
+          <span class="online-label">Full text</span>
+          <%= link_to(link.text, link.href) %>
+        <% end %>
+      </li>
     <% end %>
   </ul>
 <% end %>
--- a/app/models/concerns/eds_links.rb
+++ b/app/models/concerns/eds_links.rb
@@ -15,11 +15,15 @@ module EdsLinks
     end
 
     def present?
-      type == 'customlink-fulltext' && label.present?
+      %w[customlink-fulltext pdf ebook-pdf ebook-epub].include?(type) && label.present?
     end
 
     def sfx?
       category == 3
+    end
+
+    def ill?
+      category == 5
     end
 
     def category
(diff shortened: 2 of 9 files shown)


app/views/article/_index_default.html.erb
```diff
@@ -22,7 +22,14 @@
 <% if document.access_panels.online? && !document.eds_restricted? %>
   <ul class='document-metadata dl-horizontal dl-invert'>
     <% document.access_panels.online.links.each do |link| %>
-      <li><%= link_to(link.text, link.href) %></li>
+      <li>
+        <% if link.text =~ /^Find it in print/ -%>
```
jkeck: I wonder if we should move this up to the `Link` class somehow given that it's re-used in several places.

I may have confused things by overloading the `sfx` accessor for the class as a broad term for the particular SFX links that we use to pull in data into the async. loaded access panel I created.

If we did go this route, we may need to add an accessor unless one of the existing ones seems like a good fit, and either use that for this purpose.

Thoughts?

drh-stanford: Yeah, I had thought about that too, but I didn't want to modify the SearchWorks Link class. It's an "ILL" category link, so we could add that as a new thing to the `SearchWorks::Links::Link` class...

jkeck: I think that would be 👍

drh-stanford: Also, in my WIP PR, I refactored the EdsLinks module and it'd be easier to add it with that code there, so maybe we should review it first -- #1537

jkeck: Okay, that one looks good, want to swap the `[WIP]` (or just remove it from the other one I suppose).

drh-stanford: Ok, they're swapped now.

PR #1539, opened 2017-08-01
Include a hidden facet field parameter for article search when there …

### LINKED ISSUE #1522: "Stanford has it" should be "on" by default
Searches should start with "Stanford has it" as a selected facet, for searches starting from the home page of articles, or from the bento. Once in article search, it behaves like a regular facet - sticky until the user turns it off.
screen shot 2017-07-26 at 2 58 08 pm

jkeck:
…are no query params (e.g. on the home page).
Closes #1522

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -1,6 +1,8 @@
 <%= form_tag searchworks_search_action_path, method: :get, class: 'navbar-form', role: 'search' do %>
   <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:q, :search_field, :qt, :page, :utf8)) %>
-
+  <% if article_search? && !has_search_parameters? %>
+    <%= hidden_field_tag('f[eds_search_limiters_facet][]', 'Stanford has it') %>
+  <% end %>
   <div class="input-group">
   <% unless search_fields.empty? %>
     <span class="input-group-addon for-search-field">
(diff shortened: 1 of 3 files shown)


app/views/catalog/_search_form.html.erb
```diff
@@ -1,6 +1,8 @@
 <%= form_tag searchworks_search_action_path, method: :get, class: 'navbar-form', role: 'search' do %>
   <%= render_hash_as_hidden_fields(search_state.params_for_search.except(:q, :search_field, :qt, :page, :utf8)) %>
-
+  <% if article_search? && !has_search_parameters? %>
+    <%= hidden_field_tag('f[eds_search_limiters_facet][]', 'Stanford has it') %>
```
jkeck: Definitely up for suggestions on better places to store these values.  ¯\\_(ツ)_/¯

drh-stanford:
@jkeck this change apparently breaks the paging in the search results 🤷‍♂️ . Can you verify that you're seeing the same thing as me?

Here's what I'm doing:

1. Search for keyword in article search
2. Click next page --> see same results
3. Search for another keyword in article search
4. Turn off Stanford has it facet
5. Click next page --> see correct results

drh-stanford:
32f74dc57ac32a43719d5de8b630b753ddcb2275 has the same issue, but only if you turn ON the Stanford has it facet. Looks like that facet might be broken in the EDS API?

drh-stanford:
See https://github.com/ebsco/edsapi-ruby/issues/61

PR #1544, opened 2017-08-02
add italic style for EDS markup searchLink

### LINKED ISSUE #1538: Italicize the title part of the composed title in results and record view
Before
screen shot 2017-08-01 at 2 04 43 pm

After
screen shot 2017-08-01 at 2 05 11 pm

Note that this doesn't seem to be consistent in EDS metadata on either results or record view. Some titles are italicized, some are linked, some are neither. Not sure if we can depend on punctuation, either - some titles end with period, some with comma or other punctuation. And here's an example where a different part of the composed title is italicized and shouldn't be: https://sw-webapp-sandbox-f.stanford.edu/article/cmedm__28357797.

drh-stanford:
This PR is connected to #1538. It makes any `<searchLink>` that we mark as HTML safe show in italics.

### DIFF
--- a/app/assets/stylesheets/modules/article.scss
+++ b/app/assets/stylesheets/modules/article.scss
@@ -9,3 +9,8 @@
     padding-top: 10px;
   }
 }
+
+// EDS markup
+searchLink {
+  font-style: italic;
+}


jkeck:
@drh-stanford 👍 

See @jvine's https://github.com/sul-dlss/SearchWorks/issues/1538#issuecomment-319810594

Do you want to split that out as a separate issue?  Just want to make sure that it doesn't get lost if this PR closes that issue.

drh-stanford:
I changed the top comment so that this PR won't close #1538.

PR #1549, opened 2017-08-03
add support for Prev/Next buttons on show page

### LINKED ISSUE #1471: Previous / Next buttons not working with EDS gem

drh-stanford:
This PR closes #1471. 

This PR implements an EDS API-compatible way to execute a search to find the previous and next documents, and overrides the Blacklight `get_previous_and_next_documents_for_search` method. It does this by figuring out what an appropriate page size (`per_page`) is needed to locate both the previous and next hits and issues that search to the EDS API (but not surfacing those details out to the UI), and handles all the boundary cases shown below (e.g., first hit, last hit, hit on start of page, etc.). See the spec for these cases and let me know if I missed any.  

The cost of this algorithm is the same as the Solr implementation (i.e., a single search to find the IDs for the prev/next hits, and a retrieve to grab the record details). The first search should hit the EDS cache in most cases.

The PR also upgrades to the latest version of the EDS gem which expands its use of file-based caching (Rails' `FileStore`) for the Info, Search (new), and Retrieve (new) APIs, so I've added a pruning task to delete expired files in that cache. I've also updated Capistrano to add this as a hourly cronjob task.

### Typical record with both previous and next buttons
screen shot 2017-08-03 at 12 28 36 pm

### Last record with only previous button
screen shot 2017-08-03 at 12 29 09 pm

### First record with only next button
screen shot 2017-08-03 at 12 30 55 pm

### A search that yields a single record only shows the Back to search button
screen shot 2017-08-03 at 12 31 28 pm

### A record that is first on the 2nd (or later) page
screen shot 2017-08-03 at 12 36 50 pm

### A record that is last on the 2nd (or later) page
screen shot 2017-08-03 at 12 37 41 pm


config/deploy.rb
```diff
@@ -19,7 +19,7 @@
 # set :pty, true
 
 # Default value for :linked_files is []
-set :linked_files, %w{config/database.yml config/blacklight.yml config/honeybadger.yml config/secrets.yml config/schedule.rb public/robots.txt}
+set :linked_files, %w{config/database.yml config/blacklight.yml config/honeybadger.yml config/secrets.yml public/robots.txt}
```
drh-stanford: Note that puppet is managing the cronjobs in production, not whenever. So this change uses the git version of `config/schedule.rb` (which has the EDS cache cleanup)

Issue #1553, opened 2017-08-05
Add text to 'Send to email" article full record


camillevilla:
Split off from #1448 

## Criteria
Same as catalog records:  email is formatted HTML and includes all metadata as structured on the record view 

- title is h1, linked to SW record 
- section headings are h2
- "online" block is at the end with link(s) to full text
- multiple records are separated by `<hr>`

## Notes
- how do we take advantage of show presenter + the rendering pipeline in mailer views?

PR #1525, opened 2017-07-27
Add article email content for brief records

camillevilla:
Fixes part of #1448 
Related: #1552, #1553 

## Todo
- [x] add an email action for the article controller
- [x] create a partial for article brief record
- [x] rebase and utilize new EdsLinks module from Jessie's SFX panel work
- [x] get presenters / helper methods working correctly in mailers
- [x] create a partial for article full record
- [x] write tests
- [x] record links should point to `/article` not `/view`

### Brief record
- some cleanup to apply to current SW catalog as well as articles
  - [x]  remove plaintext 
  - [x]  remove "Message:" from the top of the email
  - [x]  create HTML links to avoid exposed URLs
- output is formatted HTML and similar to the results view:
  - [x]  structured as ordered list; sub-elements can be divs
  - [x]  title, bold, linked to SW record 
  - [x]  authors, not linked
  - [x]  composed title (journal, date, issue, etc.)
  - [x]  link to full text, with whatever wording we end up using in the search results

### Article 
screen shot 2017-08-04 at 12 31 54 pm

### Catalog record with holdings info
screen shot 2017-08-04 at 2 29 41 pm

Issue #1448, opened 2017-07-06
Add content to the "Send to email" emails


jvine:
### Full record 

Same as catalog records:  email is formatted HTML and includes all metadata as structured on the record view 

- title is h1, linked to SW record 
- section headings are h2
- "online" block is at the end with link(s) to full text
- multiple records are separated by `<hr>`

### Brief record

- some cleanup to apply to current SW catalog as well as articles
  - remove plaintext 
  - remove "Message:" from the top of the email
  - create HTML links to avoid exposed URLs
- output is formatted HTML and similar to the results view:
  - structured as ordered list; sub-elements can be divs
  - title, bold, linked to SW record 
  - authors, not linked
  - composed title (journal, date, issue, etc.)
  - link to full text, with whatever wording we end up using in the search results

Before
![screen shot 2017-07-06 at 5 31 29 pm](https://user-images.githubusercontent.com/6945691/27938168-18630130-6271-11e7-9be4-3cbc8fd6aab7.png)

After
![screen shot 2017-07-06 at 5 41 54 pm](https://user-images.githubusercontent.com/6945691/27938361-73357f56-6272-11e7-928f-3906d44be40b.png)

PR #1554, opened 2017-08-05
[WIP] Add article full record email

camillevilla:
#1553 

Continues work from #1448  / PR #1525


camillevilla:
This issue is blocked and not a high priority issue for next work cycle. When creating emails, we currently do not have access to the articles controller and the Blacklight rendering pipeline.

PR #1559, opened 2017-08-07
Disable search dropdown link for current context

### LINKED ISSUE #1556: Clicking the target you are already in, clears the search result and goes to home page
Do a search in articles.
Click the article/catalog dropdown.
Click articles.
Search is cleared and you go to the home page. 

Same deal from catalog.

The target you are already in should be disabled, so clicking it does nothing.

camillevilla:
Closes #1556 

This PR updates the search dropdown. The active search context is highlighted in yellow and the `<a>` tag is removed. Clicking on the active option (1) closes the dropdown menu (2) keeps the user on the same page.

## Demo

disabled-menu

## Example HTML:
```
<ul class="dropdown-menu" role="menu">
  <li>
    <a href="/?q=kittens&amp;search_field=search"><span class="h3">catalog</span> <span class="help-block">books &amp; media in the Stanford Libraries' collections</span></a>
  </li>
  <li class="active">
    <span class="h3">articles</span> <span class="help-block">journal articles, e-books, and other licensed e-resources</span>
  </li>
</ul>
```

### DIFF
--- a/app/views/catalog/_search_targets_widget.html.erb
+++ b/app/views/catalog/_search_targets_widget.html.erb
@@ -17,6 +17,6 @@
      </li>
      <li class="<%= 'active' if article_search? %>">
        <%= link_to_article_search %>
-    </li>
+     </li>
   </ul>
 </div>
--- a/app/assets/stylesheets/modules/search-bar.scss
+++ b/app/assets/stylesheets/modules/search-bar.scss
@@ -35,9 +35,10 @@
         left: auto;
         top: 70%;
 
-        li.active > a {
+        li.active {
           background-color: $public-note-yellow;
           color: $black; // override bootstrap
+          padding: 3px 20px;
         }
       }
     }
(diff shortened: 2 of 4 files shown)


camillevilla:
hi @jvine! Could you review the accessibility of this HTML?

jvine:
👍  🚢

app/views/catalog/_search_targets_widget.html.erb
```diff
@@ -12,11 +12,20 @@
...
+        <%= I18n.t('searchworks.search_dropdown.articles.description_html').html_safe %>
+      </li>
+     <% else %>
+       <li class="active">
+         <%= I18n.t('searchworks.search_dropdown.catalog.description_html').html_safe %>
+       </li>
+       <li>
+         <%= link_to_article_search %>
+       </li>
+     <% end %>
```
jkeck: Hmm, I worry that having the references to the i18n keys both here and in the helper.

I wonder if this might be a good use for something like `link_to_if` or `link_to_unless` in the helper?

camillevilla: @jkeck neat, thanks! I ended up using `link_to_unless_current`.

Issue #1427, opened 2017-06-29
Broken PDF Full Text links in article search


drh-stanford:
Some "PDF Full Text" links are broken -- their href is just "`detail`". We're in guest mode so it might be related to restriction, but if so, we shouldn't show the link, right?

Example: http://localhost:3000/article/aph__118425842

drh-stanford:
From Eric: The direct link to the PDF will be in the RETRIEVE method in that same spot.  We only provide it in the DETAIL record because it is a time bombed, direct access link.  You should only show it 1) when the user has been authenticated and 2) the moment they request it (so the link doesn't sit and get stale on the browser).

bmckinney:
Patrons will need to login to view: https://github.com/ebsco/eds_blacklight_search_app/blob/656774cc2f3ba6b4a4a0273bcd1438609c000ac7/app/helpers/application_helper.rb#L14

![screen shot 2017-07-05 at 4 10 55 pm](https://user-images.githubusercontent.com/509381/27883231-d6d1d30a-619d-11e7-9a97-78d7f4611d56.png)

jvine:
Resolved. PDF link gives a message to login.

PR #1562, opened 2017-08-08
add support for redirect-based PDF links

### LINKED ISSUE #1557: In guest mode, any PDF link gives an error
Discovered in utest - while in guest mode, any PDF link gave a "We're sorry, but something went wrong" or “The page you were looking for doesn't exist” error. 

Should the link direct to a login? 
Can the link say "Log in to view/download PDF"?

### LINKED ISSUE #1580: [SearchWorks/preview] EBSCO::EDS::BadRequest: {"DetailedErrorDescription"=>"DbId detail not available for profile edsapi.", "ErrorDescription"=>"DbId Not In Profile", "ErrorNumber"=>"135"}
## Backtrace

    line 31 of [PROJECT_ROOT]/app/models/eds/repository.rb: eds_find
    line 11 of [PROJECT_ROOT]/app/models/eds/repository.rb: block in find
    line 10 of [PROJECT_ROOT]/app/models/eds/repository.rb: find

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/34291357)

drh-stanford:
This PR fixes #1557 and fixes #1580. It affects the PDF links on the *search* results only (the show pages were already working). To implement these links, we need to add a new controller method that does a EDS Retrieve to get the link (it's not available from the EDS Search data). The link shown on the search results page looks like `/article/12345/pdf/fulltext` or `/article/12345/ebook-pdf/fulltext`, etc. Those links will do a Retrieve and then redirect the user to the actual PDF link.

Also, I disabled turbolinks as it seems to be causing problems negotiating the content type for the PDFs (i.e., you'd get raw PDF back in the window).

On failure, it'll display a flash error message and log the exception to Honeybadger.

### Demo from brief view
pdf-link mov

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -174,7 +174,7 @@ class ArticlesController < ApplicationController
     redirect_to url if url.present?
   rescue => e
     Honeybadger.notify(e) if defined? Honeybadger
-    flash[:error] = 'Sorry, the PDF download was not successful.<br/>Alternative route: click the title of the article you want, then try the PDF link on the detail page.'
+    flash[:error] = flash_message_for_link_error
     redirect_back fallback_location: articles_path
   end
 
@@ -263,5 +263,12 @@ class ArticlesController < ApplicationController
       return link['url'] if link['type'] == type.to_s
     end
     raise ArgumentError, "Missing #{type} fulltext link in document #{document.id}"
+  end
+
+  def flash_message_for_link_error
+    base_key = 'searchworks.articles.flashes.fulltext_link'
+    return I18n.t("#{base_key}.guest_html", login_url: new_user_session_path) if session['eds_guest']
+
+    I18n.t("#{base_key}.non_guest_html")
   end
 end
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -172,6 +172,10 @@ class ArticlesController < ApplicationController
     _response, document = search_service.fetch(params[:id])
     url = extract_fulltext_link(document, params[:type])
     redirect_to url if url.present?
+  rescue => e
+    Honeybadger.notify(e) if defined? Honeybadger
+    flash[:error] = 'Sorry, the PDF download was not successful.<br/>Alternative route: click the title of the article you want, then try the PDF link on the detail page.'
+    redirect_back fallback_location: articles_path
   end
 
   # Used by default Blacklight `index` and `show` actions
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -168,6 +168,12 @@ class ArticlesController < ApplicationController
     config.add_sort_field 'oldest', sort: 'oldest', label: 'date (oldest)'
   end
 
+  def fulltext_link
+    _response, document = search_service.fetch(params[:id])
+    url = extract_fulltext_link(document, params[:type])
+    redirect_to url if url.present?
+  end
+
   # Used by default Blacklight `index` and `show` actions
   delegate :search_results, :fetch, to: :search_service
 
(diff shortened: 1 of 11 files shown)


app/controllers/article_controller.rb
```diff
@@ -208,4 +214,13 @@ def eds_authenticated_user?
...
     params[:q].present? || params[:f].present?
   end
+
+  def extract_fulltext_link(document, type)
+    links = document.fetch('eds_fulltext_links', [])
+    links.each do |link|
+      next if link['url'].blank? || link['url'] == 'detail'
+      return link['url'] if link['type'] == type.to_s
+    end
+    raise ArgumentError, "Missing #{type} fulltext link in document #{document.id}"
```
jkeck: @jvine what should we do in the failure scenario here?  As-written, I think we would give the user the generic rails 500 page (unless I'm missing something @drh-stanford).

Another option would be to redirect back to the url the user came from w/ a flash error message explaining what happened (and we can notify Honeybadger about the issue so we can track instances of it).

Thoughts?

drh-stanford: Yes, this failure is a 500 error. This failure happens when the URL is blank or "detail" (as returned by the EDS API) or when the specified type is wrong (e.g., `/article/123/some-random-type/fulltext`). Maybe it's really a 404 case?

jvine: I'd prefer a flash message on the search result.

drh-stanford: Ok, below is the change for using a flash message. Any preference on the error message itself, right now it's:

> Full text link is unavailable at this time.

fulltext-error mov

jvine:
Does this only affect PDF links, or potentially any full text link? If PDF only, I'd like to be more precise:

> Sorry, the PDF download was not successful. 

And if we're fairly confident that the error will happen from the result page but not from the record view, then a 2-line message:

> Sorry, the PDF download was not successful. 
> Alternative route: click the title of the article you want, then try the PDF link on the detail page.

drh-stanford:
@jvine This specific error would only happen on the search results page (not the record page) and for PDF and eBook-PDF full text links.

jvine:
Okay @drh-stanford, let's use the 2-line message then, please.

drh-stanford:
@jvine ok, i've switched it to the 2-line message.

jkeck:
@drh-stanford I'm now getting a Record Not Found error from EDS when clicking on these links.  I believe it's now considering `/ebook-pdf/fulltext` as part of the identifier.

drh-stanford:
@jkeck 1db9c58 should resolve the problems with the routing that you found. thanks

app/controllers/article_controller.rb
```diff
@@ -167,6 +167,16 @@ def eds_init
     config.add_sort_field 'oldest', sort: 'oldest', label: 'date (oldest)'
   end
 
+  def fulltext_link
+    _response, document = search_service.fetch(params[:id])
+    url = extract_fulltext_link(document, params[:type])
+    redirect_to url if url.present?
+  rescue => e
+    Honeybadger.notify(e) if defined? Honeybadger
+    flash[:error] = "Sorry, the PDF download was not successful.\nAlternative route: click the title of the article you want, then try the PDF link on the detail page."
```
jkeck: The `\n` newline character doesn't result in a line break in the HTML.

flash

drh-stanford: Grrr. Thanks for catching that. 

screen shot 2017-08-11 at 3 53 11 pm

jkeck:
I have some concerns about how this might work for non-authenticated users.  I'll pick this back up next week when I can chat more w/ @jvine (since you'll be away @drh-stanford).

drh-stanford:
ok, let me know if I can help. Also, on the call Friday I learned that we need to do this link (`/article/123/pdf/fulltext`) on the detail page too, not just the search results as this PR does...

jvine:
## Not in guest mode
>Sorry, the PDF download was not successful.
>We don't know the source of the error. Try again, and if it continues to fail, please [report it as a connection problem](url).
  
(link to http://library.stanford.edu/ask/email/connection-problems)
(aside:  can we do a better job with that form internally? auto-capture some of the info?)

## In guest mode
#### Before Shibboleth
>Sorry, the PDF download was not successful because you are currently in guest mode. 
>Login for PDF download is not yet supported.
 
#### After Shibboleth
>Sorry, the PDF download was not successful because you are currently in guest mode. 
>[Log in to try the download again](url).

jvine:
Also fixes #1427, I think.

PR #1569, opened 2017-08-09
store guest mode in Blacklight session

drh-stanford:
This PR is connected to #1531. 

It stores the guest mode in the Blacklight session at the time the EDS session was created. 

The session cookie must have both the EDS session token and the guest mode stored, otherwise, it'll create a new EDS session and calculate the guest mode (based on IP address).

### Before OFF->ON campus

1. User visits site for first time from off-campus
1. An EDS session is created with guest mode ON
1. User visits site from on-campus later and executes a search
1. The EDS session is reused, but we send guest mode OFF to the EDS API  (which is incorrect as the EDS session is tied to the original guest mode in the EDS service and is ignored and will return guest mode ON results, thus causing confusion).

### Before ON->OFF campus

1. User visits site for first time from on-campus
1. An EDS session is created with guest mode OFF
1. User visits site from off-campus later and executes a search
1. The EDS session is reused, but we send guest mode ON to the EDS API (which is incorrect as the EDS session is tied to the original guest mode in the EDS service and is ignored and will return guest mode OFF results, thus causing confusion).

### After OFF->ON campus

1. User visits site for first time from off-campus
1. An EDS session is created with guest mode ON
1. User visits site from on-campus later and executes a search
1. The EDS session is reused and we send guest mode ON to the EDS API. 

### After ON->OFF campus

1. User visits site for first time from on-campus
1. An EDS session is created with guest mode OFF
1. User visits site from off-campus later and executes a search
1. The EDS session is reused and we send guest mode OFF to the EDS API. 

Note that the switch from off to on campus has no effect (as expected for this PR). We need to create a ticket for that case and implemented some kind of IP change detection. That said, this PR is really intended to reduce confusion by coming into parity with the EDS API's expectations.

### DIFF
--- a/app/controllers/article_controller.rb
+++ b/app/controllers/article_controller.rb
@@ -177,7 +177,7 @@ class ArticleController < ApplicationController
 
   def search_service
     eds_params = {
-      guest:          !eds_authenticated_user?,
+      guest:          session['eds_guest'],
       session_token:  session['eds_session_token']
     }
     Eds::SearchService.new(blacklight_config, eds_params)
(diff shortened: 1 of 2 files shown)


jkeck:
@drh-stanford can you explain what the behavior before this change is, and what the behavior is after?

drh-stanford:
ok, i've updated the top comment to have the before/after behaviors.

jkeck:
Thanks for the clarification on the behavior @drh-stanford.  Would you mind also creating a ticket for that other change?  Just want to make sure we don't forget about it.

PR #1618, opened 2017-08-18
Only render library-help chat widget when a user is on campus or...

### LINKED ISSUE #1608: Remove library chat feature from SW Feedback and no results page
In the feedback form (articles and catalog - same form?)
screen shot 2017-08-17 at 12 26 13 pm

On the catalog no-results page:
screen shot 2017-08-17 at 12 26 33 pm

Just remove these blocks. By request of IC.

jkeck:
...Stanford affiliated.

Closes #1608

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,11 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def on_campus_or_su_affiliated_user?
+    IPRange.includes?(request.remote_ip) || current_user.present?
+  end
+  helper_method :on_campus_or_su_affiliated_user?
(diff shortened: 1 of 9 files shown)


jkeck:
👍

FWIW, I think there may be two good ways to test this.

Configure `127.0.0.1` as a singleton IP address for the Stanford network in the settings file.

Alternatively, you can also start the rails server w/ a remote user (e.g. `REMOTE_USER=cvilla rails s`) which should give you current user context (that'll work until we get affiliations integrated).

camillevilla:
@jkeck FYI: I rebased `article-search` into this and deployed to stage. Works like a charm!

PR #1622, opened 2017-08-18
Change text & visual display of full text links in article results

### LINKED ISSUE #1595: Change text & visual display of full text links
Goal: reduce redundant text and visually call out PDF links.

**Content provider links**
full text content provider
- remove "full text" from the link text.

**PDF links**
full text pdf
- add PDF icon between the label and link
- height of icon is 20px
- icon has `alt="PDF"`
- remove "full text" from the link text

**HTML links**
full text html

**If there is both an HTML link and a PDF**, ideally show both
full text pdf html combined
Second record in this search result is an example that has both
https://sw-webapp-sandbox-f.stanford.edu/articles?utf8=%E2%9C%93&f%5Beds_search_limiters_facet%5D%5B%5D=Stanford+has+it&search_field=title&q=Can+cats+teach+us+how+to+live%3F

**PDF icon in 2 sizes** - 20px, and massive:
pdf-icon-20
pdf-icon

Also have an SVG version but can't upload it here. Let me know if you want it.

camillevilla:
Closes #1595 

This PR:
- removes redundant 'full text' from link text
- indicates when an article has both HTML full text and PDF full text available

### Before
sfx-before

### After
sfx-after

### Before
pdf before

### After
pdf-after

### Before
html-and-pdf-before

### After
html-and-pdf-after


app/views/articles/_index_default.html.erb
```diff
@@ -40,8 +40,10 @@
     <% if document.access_panels.online? %>
       <% document.access_panels.online.links.each do |link| %>
         <li>
-        <% if link.href == 'detail' %>
+        <% if link.href == 'detail' # PDFs %>
```
jkeck: Do we know if these the only PDF links?  I'll 👍 so we can :shipit: to preview, but we may want to look into if non timebombed detail links will also be PDFs and should get this icon.

PR #1659, opened 2017-08-30
Add warning for article source type facet behavior

### LINKED ISSUE #1638: Add contextual info at top of Source type block after a selection is made
(Avoiding the whole checkboxes issue.)

When a Source type facet selection has been made, add an alert block at the top of the facet, with this text:

> Additional selections in this facet will <b>expand</b> (rather than narrow) your search result.

### No Source type selection made
screen shot 2017-08-24 at 1 33 26 pm

### At least one selection made (panel has class .facet-limit-active)
screen shot 2017-08-24 at 1 29 52 pm

camillevilla:
Closes #1638 

This PR adds a warning to the `Source type ` facet in article search. After one source type is selected, the user is informed that additional selections will expand (rather than narrow) search results. 

### 0 selected (2,660,418 articles+ results)
0_selected

### 1 selected (1,099,359 articles+ results)
1_selected

### 2 selected (1,701,079 articles+ results)
2_selected

### DIFF
--- a/app/views/catalog/_facet_limit.html.erb
+++ b/app/views/catalog/_facet_limit.html.erb
@@ -1,4 +1,11 @@
 <% # Overridden from Blacklight to use   %>
+<% if article_search? && facet_field.key == 'eds_publication_type_facet'%>
+  <% if facet_field_in_params?('eds_publication_type_facet') %>
+    <div class="alert alert-warning" role="alert">
+      <%= t('searchworks.articles.flashes.source_type_warning_html') %>
+    </div>
+  <% end %>
+<% end %>
 <ul class="facet-values list-unstyled">
   <% paginator = facet_paginator(facet_field, display_facet) %>
   <%= render_facet_limit_list paginator, solr_field %>
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -145,9 +145,8 @@ class ArticlesController < ApplicationController
 
     # Facet field configuration
     config.add_facet_field 'eds_search_limiters_facet', label: 'Options'
-    # Commenting this out temporarily until we have a better date facet option
     config.add_facet_field 'pub_year_tisim', label: 'Date', partial: 'articles/range_limit_panel'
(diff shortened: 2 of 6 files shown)


app/views/catalog/_facet_limit.html.erb
```diff
@@ -1,4 +1,9 @@
 <% # Overridden from Blacklight to use   %>
+<% if article_search? && facet_field.key == 'eds_publication_type_facet'%>
+  <div class="alert alert-warning" role="alert">
+    <%= t('searchworks.articles.flashes.source_type_warning').html_safe %>
```
mejackreed: What do you think about renaming the key to use an `_html` suffix? I _think_ this may allow us to not have to call `html_safe` on it.

camillevilla: neat, that worked! Thanks for the i18n ✨magic✨ @mejackreed!

app/views/catalog/_facet_limit.html.erb
```diff
@@ -1,4 +1,11 @@
 <% # Overridden from Blacklight to use   %>
+<% if article_search? && facet_field.key == 'eds_publication_type_facet'%>
+  <% if facet_field_in_params?('eds_publication_type_facet') %>
```
camillevilla: @mejackreed After going back to my PR, I realized my 0 results behavior wasn't working properly! I added this line to make sure that the warning only shows up if 1 or more source types are selected.

jkeck:
I'm wondering how much we can get out of Blacklight's facet partial configuration to render a partial that allows us to inject our content into the body while rendering the rest of Blacklight's facet functionality.

I think I've done something similar in the past, happy to help look into that further.

camillevilla:
@jkeck I was able to get the partial configuration working! I'm writing a feature spec / taking a look at the facet stubbing work that you and @mejackreed did now.

camillevilla:
alrighty @mejackreed and @jkeck! Feature spec done. Feel free to squash these into one commit 
🤜 💥 🤛

PR #1670, opened 2017-08-31
Make EDS Limiters a "checkbox" facet option

### LINKED ISSUE #1596: Replace Options facet with new interaction
> search-settings-default

- remove "Options" from the facet list
- add a section above the facets, with h2 heading "Search settings"
- add checkboxes below the heading (no panel box)
```
<input type="checkbox" id="thisCheckbox" aria-describedby="willSubmit">
<label for="thisCheckbox">text comes from admin</label>
...
<span id="willSubmit" class="sr-only">Selecting this setting will update your search results</span>
```
### interaction
To account for the delay between checking a checkbox and being able to do anything else, onClick (?) should disable both checkboxes, and show the same "processing" indicator as on the record view for "all available sources", until the results are refreshed.

> search-settings-updating

### breadcrumb
In the search breadcrumb, display the selection as "Settings > [label text]"
> search-settings-breadcrumb

### collapse
OK for now if the new section remains open above the collapsed facets.
> search settings collapse

### future considerations
1. Search settings will appear on the home page as well - may need a different interaction, since it won't be updating the results; will definitely need different aria-describedby text.

2. Goal is to replace the text string from admin with the FT1 or PR code so saved searches will not be broken by text changes. Blocked by needed update from Ebsco.

jkeck:
Closes #1596 

options-facet

### DIFF
--- a/app/assets/javascripts/facet-options-checkboxes.js
+++ b/app/assets/javascripts/facet-options-checkboxes.js
@@ -1,0 +1,71 @@
+/*
+  JavaScript module to handle the behavior of the "Search settings" checkbox options
+ */
+
+var FacetOptionsCheckboxes = (function() {
+  var parentSelector = '[data-behavior="facet-options-checkboxes"]';
+  var itemSelector = '[data-behavior="facet-options-checkbox"]';
+  var loadingSpinnerSelector = '[data-behavior="loading-spinner"]';
+
+
+  function parentEl() {
+    return $(parentSelector);
+  }
+
+  function checkboxItems() {
+    return parentEl().find(itemSelector);
+  }
+
+  function loadingSpinner() {
+    return parentEl().find(loadingSpinnerSelector);
+  }
+
+  function disableAllLinksExcept(href) {
+    checkboxItems().each(function() {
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/facet-options-checkboxes.js
```diff
@@ -0,0 +1,71 @@
+/*
+  JavaScript module to fetch SFX Content into the SFX panel.
```
drh-stanford: copy-n-paste leftover

app/controllers/articles_controller.rb
```diff
@@ -144,7 +144,7 @@ def eds_init
     end
 
     # Facet field configuration
-    config.add_facet_field 'eds_search_limiters_facet', label: 'Options'
+    config.add_facet_field 'eds_search_limiters_facet', label: 'Settings', if: false
     # Commenting this out temporarily until we have a better date facet option
```
drh-stanford: can you add a brief comment here why you're using `if: false`?

jkeck:
@drh-stanford thanks for the review, I believe I have addressed the issues that you raised.

Let me know if you notice anything else.

PR #1677, opened 2017-09-01
Adds  implementation of an article sms and bitly + shortning!

### LINKED ISSUE #1641: Send to... Text needs content
Current txt is empty.

### Best case
- SMS contains only title and URL
- Keep max character length to 160 (no continuation texts)
- **Convert the SearchWorks URL to bit.ly** (which gives stanford.io outcome)
- Title displays first: truncate at (160 - (length of URL) - 3) (for ellipsis)
- URL displays second

So this record:
> Changes in energy intake and cost of transport by skipjack tuna (Katsuwonus pelamis) during northward migration in the northwestern Pacific Ocean
> https://sw-article-preview.stanford.edu/articles/edsgao__edsgcl_498056462

shortens to this:
> Changes in energy intake and cost of transport by skipjack tuna (Katsuwonus pelamis) during northward migration in the northwester...
> http://stanford.io/2gdyhxt

which is 160 chars.

### If bit.ly is not an option
- use the SW URL as-is
- truncate the title in the same way, but it will end up shorter - depending on the length EDS ID, the title may end up too short to be useful

> Changes in energy intake and cost of transport by skipjack tuna (Katsuwonus pelamis) durin...
> https://searchworks.stanford.edu/articles/edsgao__edsgcl_498056462

mejackreed:
Originally outlined in: #1641 , closes #1641 

Previously an article sms was blank. Now:

```
Forasmuch as there have been many disputes and controversies for a long time, concerning the titles to a great part of the lands in ...
http://bit.ly/2wpCGmG 
```

And for Stanford urls (deployed apps) it will get the stanford custom domain.

One a new 6.x version of Blacklight is released with https://github.com/projectblacklight/blacklight/pull/1771/commits/bb6ba0f2ca42dbf08bc8817b424092b7cb7f3678 we can remove our implementation of `to_semantic_values`

### DIFF
--- a/app/models/eds/repository.rb
+++ b/app/models/eds/repository.rb
@@ -17,7 +17,8 @@ module Eds
     def eds_search(search_builder = {}, eds_params = {})
       eds_session = Eds::Session.new(eds_params.update(caller: 'bl-search'))
       bl_params = search_builder.to_hash
-      if bl_params && bl_params['q'] && bl_params['q']['id']
+
+      if bl_params.try(:dig, 'q').try(:dig, 'id')
         # List of ID's
         results = eds_session.solr_retrieve_list(list: bl_params['q']['id'])
       else
--- a/app/models/eds/repository.rb
+++ b/app/models/eds/repository.rb
@@ -17,7 +17,13 @@ module Eds
     def eds_search(search_builder = {}, eds_params = {})
       eds_session = Eds::Session.new(eds_params.update(caller: 'bl-search'))
       bl_params = search_builder.to_hash
-      results = eds_session.search(bl_params).to_solr
+      if bl_params && bl_params['q'] && bl_params['q']['id']
+        # List of ID's
+        results = eds_session.solr_retrieve_list(list: bl_params['q']['id'])
+      else
+        # Regular search
+        results = eds_session.search(bl_params).to_solr
+      end
       blacklight_config.response_model.new(results,
                                            bl_params,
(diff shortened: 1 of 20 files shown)


app/models/eds/repository.rb
```diff
@@ -17,7 +17,13 @@ def find(id, params = {}, eds_params = {})
     def eds_search(search_builder = {}, eds_params = {})
       eds_session = EBSCO::EDS::Session.new(eds_session_options(eds_params.update(caller: 'bl-search')))
       bl_params = search_builder.to_hash
-      results = eds_session.search(bl_params).to_solr
+      if bl_params && bl_params['q'] && bl_params['q']['id']
```
jkeck: we may be able to use `dig` to clean up some of this `&&` logic?

app/presenters/sms_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+    SMS_LENGTH - short_url.length - 4
+  end
+
+  ##
+  # If the Bit.ly call fails, fallback to the URL
+  # @return [String]
+  def short_url
+    @short_url ||= Bitly.client.shorten(url).short_url
+  rescue => e
+    Rails.logger.info("Could not shorten bit.ly url #{e.message}")
```
jkeck: Wondering if this is worth reporting to Honeybadger?  I guess in the case that we've exceeded our rate-limit we wouldn't want HB getting hammered.

idk, just thinking out loud.

PR #1699, opened 2017-09-05
Add an after_sign_out_path to redirect users to Shib's logout page …

jkeck:
…after destroying the devise session.


mejackreed:
👍

PR #1708, opened 2017-09-07
Allow a configurable array of affiliations to determine if a user is …

### LINKED ISSUE #1592: Consider affiliation when determining EDS guest status
We're currently using just the fact that the user is logged in, we need to also ensure that they have the appropriate affiliation.

See https://github.com/sul-dlss/SearchWorks/issues/1348#issuecomment-311776066 for affiliations.

jkeck:
…affiliated with Stanford.

The actual affiliations are stored in shared_configs.

Closes #1592

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,14 @@ Start the rails app
 
     $ rails s
 
+## "Logging in" as a User in development
+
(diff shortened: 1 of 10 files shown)


mejackreed:
👍  and great job on the docs

PR #1712, opened 2017-09-07
Add a SearchTips class to return a random search tip for the Articles…

### LINKED ISSUE #1674: Article home page -- search tips
See #1599.

We also need a list of search tips so that we can randomly rotate them

jkeck:
… home page.

Closes #1674 

## Random tip
tip1

## Random tip
tip2

## Random tip
tip3


app/assets/stylesheets/modules/home-page.scss
```diff
@@ -21,6 +21,11 @@
     .panel-heading {
       color: $cardinal-red;
     }
+
+    dt {
+      width: 100%;
+      text-transform: none !important; // Fighting !important in base.scss
```
jkeck: 😞

config/settings.yml
```diff
@@ -47,3 +47,26 @@ STANFORD_NETWORK:
   ranges: []
 BITLY:
   ACCESS_TOKEN: 'the_bitly_apikey'
+SEARCH_TIPS:
```
jkeck: Not super pleased w/ this, but there is a goal to make this somewhat easily editable (i.e. not buried in a class/view).

Happy to hear other possible approaches.

app/views/articles/_home_page.html.erb
```diff
@@ -56,8 +56,7 @@
       <div class="panel panel-default">
         <div class="panel-body">
           <h2>Search tip</h2>
-          <h4><%= t("searchworks.articles.search_tips.heading") %></h4>
-          <p><%= t("searchworks.articles.search_tips.body") %></p>
+          <%= render SearchTips.random %>
```
jkeck: So this is pretty static and straight forward right now.

Another thought was to render all the tips into the page and use javascript to randomize.  That would have the benefit of allowing some sort of rotation/carousel of tips or a control to update with a new random tip (although that's not necessarily in the design so may be overkill at this time).

config/settings.yml
mejackreed: What about putting in a separate config file and adding source at runtime (initializer?)

https://github.com/railsconfig/config#adding-sources-at-runtime

```ruby
Settings.add_source!("/config/catfacts.yml")
```

mejackreed:
Looks good.. one comment on a potential way to integrate the tips / yml

config/settings.yml
jkeck: Neat!  Didn't know about `add_source!`.  I'll give that a go.

jkeck:
👍 Thanks for the review @mejackreed.  I'll def. have to use that in the future to help organize my settings files.

Let me know if you see anything else!

PR #1714, opened 2017-09-07
Add connection problem form

### LINKED ISSUE #1640: Add "Report a connection problem" to the sub nav bar
### LInk
Add "Report a connection problem" link to the right side of the dark sub nav bar. 
`font-size: 1.1em` so the text is not as large as "Library services".

report - menu option

### Form
Clicking the option opens a slider below the nav bar, above the remaining page content. 
- Form is in a box that is right aligned in that space so it's associated with the link itself. 
- If the user is logged in, can we autofill their name and email? 
- Form needs to secretly capture the users OS and IP address. 
  - note that load-balancing seems to mess with the IP address. this is a problem for the support staff... 

report - slide-down

### Completion
Slider is closed, flash message appears in it's place.
report - success

mejackreed:
This PR fixes #1640 

Some styling changes from ticket (per discussion with @jvine ) include:
 - no additional styling on javascript form body (no right align/ no background color change)
 - underline added for both feedback forms when open
## Update from GIF
Focus now is on top form element for each feedback form when opened
conn


app/models/feedback_mailer_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+    @ip = ip
+  end
+
+  def name
+    if params[:name].present?
+      params[:name]
+    else
+      'No name given'
+    end
+  end
```
drh-stanford: i think you can do `params.fetch(:name, 'No name given')` here

app/controllers/feedback_forms_controller.rb
```diff
@@ -22,6 +33,10 @@ def create
 
   protected
 
+  def set_form_type
+    @form_type = params.permit(:type)[:type]
+  end
+
   def url_regex
     /.*href=.*|.*url=.*|.*http:\/\/.*|.*https:\/\/.*/i
```
drh-stanford: can you just parse the URL with `URI.parse` to see whether it's valid?

config/settings.yml
```diff
@@ -1,5 +1,6 @@
 HOURS_API: <%= ENV['HOURS_API'] || 'http://example.com' %>
 EMAIL_TO: "fake-email@kittenz.com"
+CONNECTION_EMAIL_TO: "fake-connectionz@kittenz.com"
```
drh-stanford: this is a little confusing, is something like this more clear?

```yml
EMAIL_TO:
  FEEDBACK: 'blah'
  CONNECTION: 'blah'
```

app/controllers/feedback_forms_controller.rb
mejackreed: Great idea. I'm not sure of the historical reasoning behind this, but if its ok, I'm going to leave as is since this PR isn't touching that yet.

mejackreed:
Thanks @drh-stanford ! Feature test added aa8456c . @jkeck helped with getting that working for me. Unfortunately we have intermittent failures, so its marked as pending on travis. https://github.com/sul-dlss/SearchWorks/issues/1720 aims to help make these more consistent.

mejackreed:
Also I addressed your other comments accept the URI parsing.

PR #1718, opened 2017-09-07
add subject linking based on EDS searchLink markup

### LINKED ISSUE #1664: Linked subject searches not working
Example search:

https://sw-article-preview.stanford.edu/articles?utf8=%E2%9C%93&search_field=search&q=Helping+the+Rich+get+Richer%3A+A+Re-assessment+of+the+Income+Distributional+Trend+in+Hong+Kong

Review the subjects in the result view:
- multiple subdivisions with `--` separators
- several `&`s

On the result page;
https://sw-article-preview.stanford.edu/articles/bas__BAS899713

- the `&`s show as `&amp;` and are searched as such
- none of the linked subjects gets a search result. Should at least get the current record?
- author-provided keywords do get a result.
- removing the punctuation doesn't affect the result.

drh-stanford:
This PR fixes #1664 using our solution we discussed on 9/7 after standup. The core issue is that subjects returned by the EDS API can be found in different field codes (e.g., `SU`, `SH`, `DE`, `KW`).

This PR solves the issue by creating subject links that use a fielded search, the option of which is hidden in the main search box. All of the hidden fielded searches are labeled "Keyword" in the breadcrumb. Phrases are quoted where as single words are not.

The solution also provides for linking subjects hierarchically (if available) based on the EDS `<searchLink>` markup metadata that we receive.

This PR upgrades the `ebsco-eds` gem which adds the markup to the subjects metadata.

### Breadcrumb
screen shot 2017-09-07 at 1 52 52 pm

### Main search box and breadcrumb. notice "Keyword" as breadcrumb label is not available in the main search box pulldown

screen shot 2017-09-07 at 1 56 06 pm

### Subject link populates main search box with an All fields search

Note that this solution populates the main search box as an *All fields* search -- this is our compromise until EBSCO can come up with a "union" subject fieldCode API.

screen shot 2017-09-07 at 1 53 03 pm

### Hierarchical subject linking

screen shot 2017-09-07 at 2 01 02 pm

## Demo (using `bas__BAS899713`)

(and yes the data for this record includes just "Author" as an author)

subjects mov


app/models/concerns/eds_subjects.rb
```diff
@@ -0,0 +1,98 @@
...
+                  'descriptor'
+                when 'SU'
+                  'subject'
+                when 'KW'
+                  'keyword'
+                else
+                  'subject'
+                end
+        # TODO: we don't have a view context for `link_to` or `articles_path`
+        "<a href=\"/articles?q=#{CGI.escape(searchTerm)}&amp;search_field=#{field}\">#{CGI.escapeHTML(label)}</a>"
```
drh-stanford: I'm not sure how to get the view context here...

jkeck: Hmm, what about using `Rails.application.routes.url_helpers` to at least get the articles path helper. 

I think we can live w/ the html, but hardcoding the path may be a little more risky.

drh-stanford: Ok, I've added that in 190684a3be375c68ec1e96395372878117fa48c1

app/models/concerns/eds_subjects.rb
```diff
@@ -0,0 +1,99 @@
...
+  # EDS-centric subject with linking capability
+  class Subject
+    attr_reader :terms
+    delegate :present?, to: :terms
+
+    class Term < Struct.new(:fieldCode, :searchTerm, :label)
+      delegate :to_s, to: :label
+
+      def to_html
+        field = case fieldCode
```
jkeck: Do you think we can break it down a bit?  It looks like there are a few things this method is doing that could be broken down into smaller responsibilities.

app/models/concerns/eds_subjects.rb
```diff
@@ -0,0 +1,99 @@
...
+                when 'SU'
+                  'subject'
+                when 'KW'
+                  'keyword'
+                else
+                  'subject'
+                end
+        # TODO: we don't have a view context for `link_to`
+        path = Rails.application.routes.url_helpers.articles_path
+        "<a href=\"#{path}?q=#{CGI.escape(searchTerm)}&amp;search_field=#{field}\">#{CGI.escapeHTML(label)}</a>"
```
jkeck: We could pass `q` and `search_field` arguments to `articles_path` instead of constructing the url w/ `?` and `&amp;` manually.

app/models/concerns/eds_subjects.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    # @return [Array<Subject>]
+    def self.from(string_or_array)
+      return [] if string_or_array.blank?
+      xml = case string_or_array
+            when String
+              convert_simple_string_to_xml(string_or_array)
+            when Array
+              string_or_array.map { |s| convert_simple_string_to_xml(s) }.join('<br/>')
```
jkeck: This might be a little more simply written as

```
Array.wrap(string_or_array).map { |s| convert_simple_string_to_xml(s) }.join('<br/>')
```

That would allow us to drop the `case` entirely I believe.

spec/models/concerns/eds_subjects_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    it 'converts searchLinks to hyperlinks with correct search_field' do
+      subj = document.eds_author_supplied_keywords[0]
+      expect(subj.to_html).to eq '<a href="/articles?q=%22income+inequality%22&amp;search_field=keyword">income inequality</a>'
+      subj = document.eds_author_supplied_keywords[1]
+      expect(subj.to_html).to eq '<a href="/articles?q=taxation&amp;search_field=keyword">taxation</a>'
+    end
+
+  end
+
+  pending 'sometimes EDS uses semi-colons rather than <br> as delimiters'
```
jkeck: Do we need to create a ticket about this?

app/models/concerns/eds_subjects.rb
drh-stanford: oh right 👍

spec/models/concerns/eds_subjects_spec.rb
drh-stanford: yes, created https://github.com/sul-dlss/SearchWorks/issues/1732

drh-stanford:
@jkeck thanks for the feedback. i think i've addressed them all and i've force pushed the changes.

PR #1734, opened 2017-09-08
adds in alternate catalog feature

### LINKED ISSUE #1709: Add alternate (catalog/articles) bento-ish insert in search results
Between the 3rd and 4th search result, add a mini bento-like insert that offers results from the alternate SW target.

From articles
catalog mini bento

From catalog
articles mini bento

Facet values - show 5, keep each `<a>` together so it doesn't split across a line.

At some future point, when Bento exists, will add a link to the All search at the bottom of the block.

mejackreed:
Fixes #1709 

altcat


app/assets/javascripts/alternate_catalog.js
```diff
@@ -0,0 +1,82 @@
...
+    })
+    sourceOrResouce.forEach(function (facet) {
+      var count = 0;
+      // Iterating over everything here would be even slower.. sort it and take
+      // the top5
+      var topFive = facet.items.sort(function(a, b) {
+        return b.hits - a.hits;
+      }).slice(0, 5);
+      topFive.forEach(function (item) {
+        linkedUrl = url + '&f[' + facet.name + '][]=' + item.value;
```
drh-stanford: does `item.value` need to be encoded for these facet values? i know some facets have values with punctuation in them. it looks like we're restricting to only the source and resource types.

mejackreed: Great catch. I've encoded this value now.

PR #1764, opened 2017-09-12
[WIP] Add referrer to the login link presented to users who try to access…

jkeck:
… full-text in guest mode.

Currently just going to `/webauth/login` blindly is causing a stale-session error from Shibboleth.

(going to deploy to stage to validate)


jkeck:
Going to submit a PR to kill this link all together.

PR #1771, opened 2017-09-12
add preformatted HTML for fulltext link to JSON API

### LINKED ISSUE #1753: Bento results linking
The results for bento has links to the content. 

We may need to extend the Blacklight JSON results presenter in order to provide our pre-formatted links so that we don't have to duplicate the logic in the bento app.

If this ends up getting extremely complicated we can drop the links and just allow the user to go to the document/result to get the link.

drh-stanford:
This PR fixes #1753. The simple rendering of the link, which the JSON API surfaces, is migrated to a helper method, but there's more complex renderings in the index pages that I didn't migrate.

#### After: via `format=json` query for /articles

like so http://localhost:3000/articles?q=climate+change&format=json
screen shot 2017-09-12 at 2 41 13 pm

^^^The link is relative so the Bento app will need to inject the host.

screen shot 2017-09-12 at 3 05 45 pm

^^^Other links are fully qualified

### DIFF
--- a/app/views/articles/index.json.jbuilder
+++ b/app/views/articles/index.json.jbuilder
@@ -1,0 +1,11 @@
+docs = @presenter.documents.collect do |document|
+  link = document.eds_links.all.first # top priority one only
+  data = document.to_h # avoids deprecation warning
+  data['fulltext_link_html'] = render_fulltext_link(link, document) if link.present?
+  data
+end
+json.response do
+  json.docs docs
+  json.facets @presenter.search_facets_as_json
+  json.pages @presenter.pagination_info
+end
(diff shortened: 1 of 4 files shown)


app/views/articles/index.json.jbuilder
```diff
@@ -0,0 +1,13 @@
+docs = @presenter.documents.collect do |document|
+  link = document.eds_links.all.first # top priority one only
+  next if link.blank?
```
jkeck: Do we really want to not show a result if it doesn't return a link? (not sure if this is a real case or not).  Could this potentially have the affect of not returning the proper number of results (because some are omitted) or not returning EDS restricted results?

drh-stanford: thanks. i've added a test case for that and force pushed.

PR #1775, opened 2017-09-13
add an and

mejackreed:
@jvine what do you think?


jvine:
...ok. Implies a completeness to the list, but the alternative is grammatically suspect.

PR #1782, opened 2017-09-14
Make sure that mini-bento article only includes direct access facet

### LINKED ISSUE #1768: Mini bento articles+ search should include "Direct access" setting
Do a search in Catalog, click the Articles+ button, you get the search in Articles+ without "Direct access" checked. Please add it as the default to the mini-bento, the same way it is from the Articles home page.

mejackreed:
Closes #1768

### DIFF
--- a/app/views/catalog/_alternate_catalog.html.erb
+++ b/app/views/catalog/_alternate_catalog.html.erb
@@ -1,4 +1,4 @@
-<div class='alternate-catalog' data-alternate-catalog="<%= articles_path(q: params.fetch(:q)) %>">
+<div class='alternate-catalog' data-alternate-catalog="<%= articles_path(q: params.fetch(:q), f: { eds_search_limiters_facet: ['Direct access to full text'] }) %>">
   <button type='button' class='alternate-catalog-close close' aria-label="Close">
     <i class='fa fa-times-circle'></i>
   </button>
@@ -13,8 +13,9 @@
           Journal articles, e-books, &amp; other e-resources
         </div>
       </div>
+      <%=  %>
       <div class='col-md-6'>
-        <%= link_to(articles_path(q: params.fetch(:q)), class: 'btn btn-primary') do %>
+        <%= link_to(articles_path(q: params.fetch(:q), f: { eds_search_limiters_facet: ['Direct access to full text'] }), class: 'btn btn-primary') do %>
           See <span class="alternate-catalog-count"%></span> article+ results
         <% end %>
       </div>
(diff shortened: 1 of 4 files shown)


app/views/catalog/_alternate_catalog.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class='alternate-catalog' data-alternate-catalog="<%= articles_path(q: params.fetch(:q)) %>">
+<div class='alternate-catalog' data-alternate-catalog="<%= articles_path(q: params.fetch(:q), f: { eds_search_limiters_facet: ['Direct access to full text'] }) %>">
```
jkeck: I wonder if we could use the `FacetOptionsPresenter` to find the limiter that is `enabled_by_default?` and use that here instead.  We might be able to pull the field reference from that as well.

mejackreed: Thanks for the recommendation. Could you talk about that API a bit more? Looking at `FacetOptionsPresenter` I'm not really sure how I should use that in this context.

jkeck: Check out where we set the checkbox on the [article home page](https://github.com/sul-dlss/SearchWorks/blob/master/app/views/articles/_home_page.html.erb#L20-L22) for context.

There is a `facet_options_presenter` that you can use to easily access the facet options object.  So something along the lines of

```ruby
facet_otpions_presenter.limiters.find(&:enabled_by_default?)
```

should get you the the `Limiter` object that has the EDS administration config label (e.g. 'Direct access to full text').  

This will have the benefit of not requiring a change on our end of that text is updated or the default changes.

mejackreed: Upon further consideration I decided to remove using the `FacetOptionsPresenter` here. A few reasons:

 - I thought that now needing to instantiate an `Eds:Session` in the `CatalogController` to do a query against the `ArticleController` felt a bit unclean. The ajax query should handle the session stuff, and I don't like the idea of needing to do that just to create a link that ajax will hit.
 - I recognize that the default search behavior is hardcoded. I'm not sure of the right abstraction yet and it didn't feel right from `FacetOptionsPresenter` so I'm leaving hard coded.

jkeck: One of our main goals is to not have to change SearchWorks code if this text changes.

I think we would have to ask Jennifer and Alexis to weigh in because this could potentially have large impacts into approaches that we've already decided on.  I would really like to avoid hard-coding this in our codebase at all costs.

Is there a performance concern you're worried about here?  We can move the EDS session up if needed and the session is cached so it wouldn't be an issue w/ duplicating the session. If it's a naming issue, then let's rename it to make it "feel better".

`FacetOptionsPresenter` is something that does exactly what we want here AFAICT, as well as meets a need of honoring EDS admin settings.  If Jennifer and Alexis are willing to completely break that, then 👍.

PR #1786, opened 2017-09-14
Add additional selections warning to EDS Database facet

### LINKED ISSUE #1769: Add "additional selections" message to Databases facet (like Source type)
Databases is now an OR facet like Source type. Pls add the same yellow message after the first Database facet is selected.

camillevilla:
Closes #1769 

This PR:
- adds a warning to the EDS Database facet that additional selections will expand results
- refactors EDS Source type facet to use the same partial as EDS Database

## Before
facet without warning

## After
facet with warning


app/controllers/articles_controller.rb
```diff
@@ -147,13 +147,13 @@ def eds_init
...
     config.add_facet_field 'pub_year_tisim', label: 'Date', partial: 'articles/range_limit_panel', range: true
-    config.add_facet_field 'eds_publication_type_facet', label: 'Source type', partial: "facet_eds_publication_type"
+    config.add_facet_field 'eds_publication_type_facet', label: 'Source type', partial: "facet_additional_selections_warning"
     config.add_facet_field 'eds_language_facet', label: 'Language' # , limit: 20 TODO: Need to handle facet limiting
     config.add_facet_field 'eds_subject_topic_facet', label: 'Topic'
     config.add_facet_field 'eds_subjects_geographic_facet', label: 'Geography'
     config.add_facet_field 'eds_journal_facet', label: 'Journal title'
     config.add_facet_field 'eds_publisher_facet', label: 'Publisher'
-    config.add_facet_field 'eds_content_provider_facet', label: 'Database'
+    config.add_facet_field 'eds_content_provider_facet', label: 'Database', partial: 'articles/facet_additional_selections_warning'
```
jkeck: Hmm, why are we using `articles/facet_additional...` here and just `facet_additional...`  above?

camillevilla: oh hmm. They both work, but I guess `articles/` is redundant because we're already in the articles controller. I've updated this to use the shorter syntax.

PR #1790, opened 2017-09-15
Refactor render_fulltext_link into a presenter

jkeck:
This PR refactors `render_fulltext_link` into a presenter as well as changes the functionality to handle the links that are presented on the search result instead of what's on the record view.

This change is necessary for the bento app (and referenced in sul-dlss/sul-bento-app#57).

### DIFF
--- a/app/presenters/article_fulltext_link_presenter.rb
+++ b/app/presenters/article_fulltext_link_presenter.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+##
+# A presenter class handle the logic to render markup for fulltext links from EDS
+class ArticleFulltextLinkPresenter
+  delegate :article_url, :article_fulltext_link_url, :link_to, :image_tag, to: :context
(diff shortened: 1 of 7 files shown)


app/presenters/article_fulltext_link_presenter.rb
```diff
@@ -0,0 +1,56 @@
...
+  def access_panel_links
+    document.access_panels.online.links || []
+  end
+
+  def document_has_fulltext?
+    document.html_fulltext? && !online_access_panel?
+  end
+
+  def render_fulltext_link(link)
+    if link.href == 'detail' # PDFs
```
drh-stanford: FYI Bill added a new attribute to `eds_fulltext_links` called `expires` which is a boolean. If true, it means we need to use our `/articles/id/type/fulltext` link (e.g., `article_fulltext_link_url`). We're not using it currently -- we'd need to add it to the SearchWorks link class.

PR #1797, opened 2017-09-20
Use absolute URLs for article icons

### LINKED ISSUE #1796: Need to use absolute asset url for article link icons
It is currently giving a relative path and it needs to be absolute.

camillevilla:
Closes #1796 and fixes sul-dlss/sul-bento-app#68 once deployed to production.

This PR uses absolute URLs for the article brief view PDF icon. 
pdf-icon

## Before
`PDF`

## After
`PDF`


drh-stanford:
Do we need to configure `config.action_controller.asset_host` for this to work in production/stage?

jkeck:
@drh-stanford I _think_ it should get it from the request host.  If it doesn't work when we deploy to stage we can see if `asset_host` can do something for us.

PR #1804, opened 2017-09-21
Enable bento links

### LINKED ISSUE #1695: Link to bento app
- [x] Add "All" option to dropdown
- [x] Catalog/Article home pages
- [x] No results page sidebar

camillevilla:
Closes #1695 

This PR:
- directs 'All' links to https://library.stanford.edu/all" and passes simple search queries
- adds 'All' to the search dropdown
- adds 'All of the above' search to home page mini bento

### Before
mini_bento_before
### After
mini_bento_after

### Before
search_dropdown_before
### After
search_dropdown_after

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -109,10 +109,10 @@ module ApplicationHelper
   end
 
   def link_to_bento_search
-    mapped_params = { q: params[:q] }
+    query_string = params[:q].present? ? "?#{{ q: params[:q] }.to_query}" : nil
     link_to(
       t('searchworks.search_dropdown.bento.description_html'),
-      'https://library.stanford.edu/all'
+      "https://library.stanford.edu/all/#{query_string}"
     )
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -108,11 +108,11 @@ module ApplicationHelper
     )
   end
 
-  # TODO: add bento link, see issue #1695
   def link_to_bento_search
+    mapped_params = { q: params[:q] }
     link_to(
       t('searchworks.search_dropdown.bento.description_html'),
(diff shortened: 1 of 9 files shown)


app/views/catalog/_search_targets_widget.html.erb
```diff
@@ -11,12 +11,15 @@
...
-     </li>
+  <ul class="dropdown-menu" role="menu">
+    <li class="<%= 'active' unless article_search? %>">
+      <%= link_to_catalog_search %>
+    </li>
+    <li class="<%= 'active' if article_search? %>">
+      <%= link_to_article_search %>
+    </li>
+    <li>
+      <%= link_to_bento_search %>
```
jkeck: I believe the "All" link is supposed to be first: 

desired-drop-down

camillevilla: 👍

camillevilla:
Alrighty! Requested changes have been made and screenshot has been updated.

app/helpers/application_helper.rb
```diff
@@ -108,11 +108,11 @@ def link_to_article_search
     )
   end
 
-  # TODO: add bento link, see issue #1695
   def link_to_bento_search
+    query_string = params[:q].present? ? "?utf8=✓&#{{ q: params[:q] }.to_query}" : nil
```
jkeck: Sorry I didn't catch this on the first pass.  I don't think we'll want to explicitly call out `utf8` in our URL.  That's a rails thing that gets passed on, we could either pass `params[:utf8]` (so it gets set to what rails set instead of forced) or just drop it all-together.

camillevilla: Okie doke, I decided to drop it. Bento doesn't seem to mind. ¯\_(ツ)_/¯

PR #1822, opened 2017-10-11
Update ebsco-eds gem.

### LINKED ISSUE #1819: Update latest EDS gem
This should resolve https://github.com/sul-dlss/SearchWorks/issues/1800 and related, as well as current pagination issues.  

The flag to apply case tense for language and geography facets should be turned OFF - that is we want no change to the way those facets are displayed at this time.

jkeck:
Closes #1819

### DIFF
--- a/app/models/eds/session.rb
+++ b/app/models/eds/session.rb
@@ -19,20 +19,21 @@ module Eds
 
     def session_options
       {
-        guest:                eds_params[:guest],
-        session_token:        eds_params[:session_token],
-        caller:               eds_params[:caller],
-        user:                 Settings.EDS_USER,
-        pass:                 Settings.EDS_PASS,
-        profile:              Settings.EDS_PROFILE,
-        use_cache:            Settings.EDS_CACHE,
-        eds_cache_dir:        Settings.EDS_CACHE_DIR,
-        timeout:              Settings.EDS_TIMEOUT,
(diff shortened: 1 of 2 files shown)


app/models/eds/session.rb
```diff
@@ -19,20 +19,21 @@ def object
...
+        profile:                      Settings.EDS_PROFILE,
+        use_cache:                    Settings.EDS_CACHE,
+        eds_cache_dir:                Settings.EDS_CACHE_DIR,
+        timeout:                      Settings.EDS_TIMEOUT,
+        open_timeout:                 Settings.EDS_OPEN_TIMEOUT,
+        api_hosts_list:               Settings.EDS_HOSTS,
+        debug:                        Settings.EDS_DEBUG,
+        log:                          File.join(Settings.EDS_LOGDIR, 'faraday.log'),
+        decode_sanitize_html:         true,
+        recover_from_bad_source_type: true
```
jkeck: This line being added is the only non-formatting change to this file.

jkeck:
AFAICT 0.3.18.pre (as tagged in GH) is the same as 0.3.19.pre.  When 0.3.18.pre was released it didn't have the appropriate changes so the gem developer re-released (w/o any change other than the version bump).

jkeck:
@drh-stanford I should also mention this is current deployed to stage.

drh-stanford:
Gotcha 👍

PR #1823, opened 2017-10-17
clear EDS cache biweekly

### LINKED ISSUE #1814: EDS cache is not being cleaned up correctly
The edsapi-ruby gem sets a TTL for all cached items, and the `FileStore.cleanup` method is supposed to remove all the files that have expired. We have an hourly rake task `searchworks:prune_eds_cache` but it's not reducing the size of the cache directory. It's grown to 4GB currently.

See https://github.com/ebsco/edsapi-ruby/blob/9a9b03506d7b71c9259802647904fe08348a2b78/lib/faraday/eds_caching_middleware.rb#L73

drh-stanford:
This PR closes #1814 

cc: @eefahy

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -7,3 +7,7 @@ set :output, 'log/cron.log'
 every 1.hours, roles: %i[app] do
   rake 'searchworks:prune_eds_cache'
 end
+
+every '0 0 1,15 * *', roles: %i[app] do # biweekly
+  rake 'searchworks:clear_eds_cache'
+end


julianmorley:
This rake task fails on the production boxes, and so probably won't fix the issue. 

> [blacklight@sw-webapp-e current]$ rake 'searchworks:clear_eds_cache'
rake aborted!
Gem::LoadError: You have already activated rake 10.4.2, but your Gemfile requires rake 12.0.0. Prepending `bundle exec` to your command may solve this.

jkeck:
@julianmorley when running under whenever the command will be run under the correct context w/ the correct envs set (which does work).

```bash
[blacklight@sw-webapp-e current]$ RAILS_ENV=production bundle exec rake searchworks:clear_eds_cache
Clearing cache in /opt/app/blacklight/SearchWorks/releases/20171012045721/tmp/faraday_eds_cache
[blacklight@sw-webapp-e current]$ 
```

julianmorley:
Great! Can we get this merged then so I don't have to do this manually every time I'm on call?

jkeck:
I'm going to miss your regular visits!

PR #1841, opened 2018-02-08
Switch from using jQuery.text() to jQuery.html() when replacing…

jkeck:
… backend lookup link content.

### Given the following example:
```javascript
var linkContent = "&gt;&lt;script&gt;alert(&quot;BAD JUJU&quot;)&lt;/script&gt;";
var $linkNode = $('<span>' + linkContent + '</span>');
var newLinkContent = $linkNode.text();
newLinkContent += ' other string';
$linkNode.html(); // Does not include unescaped javascript
$linkNode.html(newLinkContent);
$linkNode.html(); // Includes unescaped javascript
```

Using `var newLinkContent = $linkNode.html();` instead of of `var newLinkContent = $linkNode.text();` (effectively what this PR does) maintains the escaped javascript after rendering to the page.

### DIFF
--- a/app/assets/javascripts/backend_lookup.js
+++ b/app/assets/javascripts/backend_lookup.js
@@ -30,9 +30,9 @@ Blacklight.onLoad(function(){
     }
 
     function updateLink(count, el){
-      var linkText = $(el).text();
+      var linkText = $(el).html();
       linkText += "... <strong>found " + count + " results</strong>";
-      $(el).html(linkText)
+      $(el).html(linkText);
     }
     Plugin.prototype = {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/backend_lookup.js
```diff
@@ -30,9 +30,9 @@ Blacklight.onLoad(function(){
     }
 
     function updateLink(count, el){
-      var linkText = $(el).text();
+      var linkText = $(el).html();
```
jkeck: This is the only relevant change.

PR #1846, opened 2018-03-12
URL-Escape the stacks url

cbeer:


config/deploy/preview_prod.rb
```diff
@@ -1,7 +1,7 @@
 set :bundle_without, %w[sqlite development test].join(' ')
 
 # alias searchworks-preview
-server 'sw-webapp-sandbox-a.stanford.edu', user: 'blacklight', roles: %w[web db app]
```
mejackreed: Did this server change?

cbeer: Not sure why this is in here.

PR #1849, opened 2018-03-17
Handle parsing EZProxy url hosts (a tiny bit) better.

jkeck:


app/models/concerns/marc_links.rb
```diff
@@ -49,6 +49,17 @@ def file_id(link_field)
...
     end
 
+    # Parse a URI object to return the host of the URL in the "url" parameter if it's a proxied resoruce
+    def link_host(link)
+      return link.host unless link.to_s =~ SearchWorks::Links::PROXY_REGEX && link.to_s.include?('url=')
+      proxy = CGI.parse(link.query)
+      return link.host unless proxy.key?('url')
+
+      # strip everything off after the query, we're only parsing the host and CGI.parse unencodes (which we don't want)
+      %r{((https?://)?[^\/|\?]+)} =~ proxy['url'].first
```
cbeer: I don't know if this is actually any better, but maybe something like:

```
url = URI.extract(proxy['url'].first).first
return link.host unless url
URI.parse(url).host
``` 

I guess the only benefit is deferring the regex up to `URI`

PR #1856, opened 2018-03-21
Sanitize summaries in search results.

jkeck:
This was being caused by data in the summary.

## Before
dataset-old


cbeer:
👍 does this change deserve a test? I'm persuadable either way.

jkeck:
Yeah, I'll add one. 👍

jkeck:
Test added @cbeer

PR #1865, opened 2018-04-21
[WIP] Load application.js asynchronously to improve 1st draw.

jkeck:
_**[WIP]** while we see there are any gotchas we need to be aware of via sul-dlss/exhibits#1153_


thatbudakguy (approved):
seems to have worked on exhibits!

Issue #1887, opened 2018-07-16
Update dependencies for searchworks


cmh2166:
In preparation for SW Performance monitoring & analysis.

TODO:
- [ ] Update to rails 5.2.0 (or latest) (@camillevilla #1892)
- [x] Update to Blacklight 6.15.0 (or latest) [@jkeck looking into this]
  * #1894 
- [ ] Update all other dependencies (which theoretically shouldn't need action) (@camillevilla #1892)

jkeck:
Happy to pair up w/ somebody on this.  I have gone through this process once or twice trying to do normal dependency updates and bailed (with simply updating vulnerable dependencies and moving on).

PR #1894, opened 2018-07-17
Update to Blacklight 6.15.0

jkeck:
Connected to #1887

TODO:
- [x] Address failing specs
- [x] Validate if we're getting different solr queries in a meaningful way (0f94b13 seems to indicate search results changed from what non-relevancy related tests changed, not sure we should be too concerned)
  * I checked out master and was seeing the same document re-order, so I don't believe it's related.
- [x] Address application_name deprecations


app/views/layouts/searchworks.html.erb
```diff
@@ -37,6 +37,31 @@
       <link rel="stylesheet" type="text/css" href="https://www.stanford.edu/su-identity/css/ie/ie7.css" />
     <![endif]-->
 
+    <% if Rails.env.test? %>
+      <%= javascript_tag '$.fx.off = true;' %>
+      <style>
+        * {
```
mejackreed: Hadn't seen this one before, but makes sense why we would do this.

PR #1897, opened 2018-07-18
Update to use solr_wrapper and test on Solr 6.6.5

### LINKED ISSUE #1891: Use solrwrapper instead of jettywrapper and test application against Solr 6+

mejackreed:
Fixes #1891 

TODO:
 - [x] figure out why the following are not working: (needed to include the correct extras in solrconfig)
   - [x] ICUFoldingFilterFactory
   - [x] ICUNormalizer2FilterFactory
 - [x] ~~Do we need synonyms.txt ? Doesn't seem to be used~~ Yes we do.
 - [x] Question where does the canonical configuration live? It looks as if we have different schemas/configs on servers than we do in the app
 - [x] Should we update the app to use the configs that are live?
   - [x] is `img_info` field deprecated/gone?

The `schema.xml` was copied directly from our sw-solr-g instance, while only a subset of `solrconfig.xml` was copied over (mainly the facets see block here: https://github.com/sul-dlss/SearchWorks/pull/1897/files#diff-69d9952025d4f6b6ce8a4a6c21a2fd8bR830).


jkeck:
> Should we update the app to use the configs that are live?

Yeah, I think we've been slowly moving towards that.  Not sure how much effort doing a full copy of the production config (and addressing failing tests) vs. handling it piecemeal as we've been doing.

mejackreed:
I have it working locally now with live prod configs (indexing and what not). Some failing tests but mostly seem due to deprecated fields still present in app but not in our prod configs e.g `img_info`

jkeck:
Gotcha, yeah, we use `file_id` now right?

This is what we were doing in order to transition to that new field, and I think we can safely remove references to `img_info` at this point.

mejackreed:
Works for me!

PR #1900, opened 2018-07-20
Migrate solr_local_parameters to solr_parameters.

### LINKED ISSUE #1890: Update blacklight configuration to avoid using solr local parameters syntax for non-lucene queries
Local parameters are certainly used as part of the search fields:

https://github.com/sul-dlss/SearchWorks/blob/master/app/controllers/catalog_controller.rb#L202-L305

(and can probably be trivially replaced with regular URL parameters?)

Also, identify anywhere else local parameter queries are produced.

jkeck:
Closes #1890 

[WIP] while we address issues w/ searching CJK.

An example URL that should return results form the production index (but is not):
- http://solr-host/solr/select?defType=edismax&pf=%24%7Bpf_author%7D&pf2=%24%7Bpf2_author%7D&pf3=%24%7Bpf3_author%7D&q=Stanford&qf=%24%7Bqf_author%7D&qt=search&wt=json
  * _(this is a simplified version of the URL that SearchWorks is attempting to send)_

The same search that is returning expected results using local params:
- http://solr-host/solr/select?defType=edismax&q=%7B%21qf%3D%24qf_author+pf%3D%24pf_author+pf3%3D%24pf3_author+pf2%3D%24pf2_author%7DStanford&qt=search&wt=json


jkeck:
Ugh, and from some local smoke tests against the production index, it appears that all fielded search is broken after this change.

cbeer:
Ugh, indeed. It looks like macro expansion was only added in Solr 5.1.

PR #1936, opened 2018-09-11
Update .solr_wrapper.yml to use 7.4.0

jkeck:
In anticipation of an sul-solr impending solr upgrade.


jkeck:
Looks like the OK Computer build in Solr check may not work w/ how solr 7 does it's pings (or we may need to do something different w/ how we pass the solr URL to our OK Computer config).

PR #1996, opened 2018-09-26
Display additional titles in searchworks, just like they are displaye…

### LINKED ISSUE #1634: Display MODS alternative title
Alternative title from MODS record doesn't display. 
https://searchworks.stanford.edu/view/ky910zy0024 
Additional example - https://searchworks.stanford.edu/view/bq097ny3917

**Please see spec in comment from Arcadia below**

https://jirasul.stanford.edu/jira/browse/SW-1838

cbeer:
…d in PURL.

Fixes #1634 

screen shot 2018-09-26 at 15 13 57

screen shot 2018-09-26 at 15 15 09

### DIFF
--- a/app/views/catalog/record/_mods_bibliographic.html.erb
+++ b/app/views/catalog/record/_mods_bibliographic.html.erb
@@ -1,4 +1,9 @@
 <% document ||= @document %>
+
+<% document.mods.subTitle.each do |title| %>
+  <%= mods_record_field(title) %>
+<% end %>
+
 <% document.mods.audience.each do |audience| %>
   <%= mods_record_field(audience) %>
 <% end %>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -317,7 +317,7 @@ GEM
       iso-639
       nokogiri
       nom-xml (~> 1.0)
-    mods_display (0.5.1)
+    mods_display (0.6.0)
       i18n
       stanford-mods (~> 2.1)
     msgpack (1.2.4)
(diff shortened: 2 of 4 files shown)


app/views/catalog/record/_mods_bibliographic.html.erb
```diff
@@ -1,4 +1,9 @@
 <% document ||= @document %>
+
+<% document.mods.send(:mods_field, document.mods.instance_variable_get(:@xml), :title).fields.reject { |x| x.label =~ /^Title/i }.each do |title| %>
+  <%= mods_record_field(title) %>
+<% end %>
```
jkeck: Is this ultimately different than using `document.mods.sub_title` (which is effectively all the titles except the first)?

If not, I'm wondering if this is something that would be good to push upstream since this code is now duplicated in both PURL and SearchWorks (which we can totally split out into a separate ticket).

jkeck: And now that I look at the code, I think ModsDisplay intended to surface `sub_title` as an accessor but maybe accidentally did not.  If it would server this purpose, I think that would be a good way to approach this instead of using `send` in a view.

jkeck: I added a PR to `ModsDisplay` to add a `subTitle` accessor. sul-dlss/mods_display#58

If that is accepted, and we cut a release, we can remove the `#send` on line 3 above and replace it with:

```erb
<% document.mods.subTitle.each do |title| %>
```

PR #1999, opened 2018-09-27
Use the date_range from the EDS response when the pub year field is…

### LINKED ISSUE #1721: Date slider disappears
https://sw-webapp-sandbox-f.stanford.edu/articles?f%5Beds_content_provider_facet%5D%5B%5D=Academic+Search+Premier&f%5Beds_search_limiters_facet%5D%5B%5D=Direct+access+to+full+text&q=school+reform&search_field=search&sort=relevance

jkeck:
…not available.

Closes #1721 

## Before
before

## After
after

### DIFF
--- a/app/models/eds/repository.rb
+++ b/app/models/eds/repository.rb
@@ -23,7 +23,7 @@ module Eds
         results = eds_session.solr_retrieve_list(list: bl_params['q']['id'])
       else
         # Regular search
-        results = eds_session.search(bl_params).to_solr
+        results = eds_results_with_date_range(eds_session.search(bl_params).to_solr)
       end
       blacklight_config.response_model.new(results,
                                            bl_params,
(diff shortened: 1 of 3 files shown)


app/models/eds/repository.rb
```diff
@@ -41,5 +41,21 @@ def eds_find(id, params, eds_params)
...
                                            blacklight_config: blacklight_config)
     end
+
+    # Some data providers in EDS don't have the data that populates the pub_year_tisim field
+    # (although we can query against the data), so this forces the pub_year_tisim field to be populated
+    # when the response from EDS still contains date range data.
+    # https://github.com/sul-dlss/SearchWorks/issues/1721
+    def eds_results_with_date_range(results)
+      return results if results.dig('facet_counts', 'facet_fields', 'pub_year_tisim').present?
+      return results unless results.key?('date_range')
```
jkeck: We could just return the `date_range` `minyear` and `maxyear` unconditionally when available, but this felt potentially a bit safer.

mejackreed: Yeah I get that.

PR #2014, opened 2018-09-28
No results updates

### LINKED ISSUE #1941: No results should show links to WorldCat, HathiTrust, ILL
https://jirasul.stanford.edu/jira/browse/SW-1909
https://jirasul.stanford.edu/jira/browse/SW-1930

mejackreed:
Fixes #1941 

screen shot 2018-10-01 at 12 52 57 pm
screen shot 2018-10-01 at 12 52 51 pm


mejackreed:
@jvine the styling for no-results is a little bit different than how you outlined in the ticket. I’m thinking we can cleanup or modify as needed when we get more eyes on it? https://user-images.githubusercontent.com/1656824/46299268-3a450680-c55e-11e8-9d05-eec56962c310.png

One change is that the “remove limits” “pill” is a continuous link where the pill part is styled differently as a span. Not sure if this was the original intention but should easy enough for us to change as needed.

jvine:
> One change is that the “remove limits” “pill” is a continuous link where the pill part is styled differently as a span. Not sure if this was the original intention but should easy enough for us to change as needed.

This is as intended. 

What's missing is the "More options" heading and the font-size/header formatting of the options & sidebar. I'd like to see at least the "More options" heading added before this is released.

mejackreed:
Got it thanks. Just added.

mejackreed:
Yeah I didn't know what should or shouldn't be there based off a user login. @jvine any pref? How are you getting the links to be disabled, I couldn't get that.

jkeck:
@mejackreed I think it happens when the IC desk is closed.

jvine:
Okay, I see 2 issues here.

1:  The page looks unbalanced when there is no chat option present. True. Applies to users who are not logged in, or not on campus/VPN. We could do something like this for those users, instead of removing the Chat section altogether:

screen shot 2018-10-02 at 12 25 20 pm

2. The function that checks the current status of library chat controls the whole block of chat link plus help text below it, and applies the "disabled" class to all the links in the block.
  a) the help text needs to be outside this block so it can have active links even when chat is unavailable
  b) the font color for `.disabled` should be changed from `#ccc` to `#737373`

Should some or all of this be a separate ticket?

mejackreed:
I've updated this PR to implement the desired functionality as I understand it.

### Disabled color changed, and not logged in text
screen shot 2018-10-03 at 7 51 06 am
### Only "first" link is now disabled
screen shot 2018-10-03 at 7 48 42 am

PR #2034, opened 2018-10-08
add styling and sanitzing for mhld display styles

### LINKED ISSUE #2021: 866 mhold should wrap
Force this field to wrap even if there are no spaces

screen shot 2018-10-02 at 1 38 58 pm
screen shot 2018-10-02 at 1 39 11 pm

https://searchworks.stanford.edu/view/8838463

mejackreed:
Fixes #2021 in a two-prong approach.

### CKEY 8838463

Uses preferred approach for access panel styling:

screen shot 2018-10-08 at 9 02 46 am

Uses preferred approach for librarian view as I suspect we do not want to sanitize this data:

screen shot 2018-10-08 at 9 03 36 am

### DIFF
--- a/lib/holdings/mhld.rb
+++ b/lib/holdings/mhld.rb
@@ -48,7 +48,7 @@ class Holdings
     end
 
     def sanitize_mhld_data(data)
-      CGI.escape_html(data).gsub('),', '), ').gsub('-', '-<wbr/>').html_safe if data.present?
+      CGI.escape_html(data).gsub('),', '), ').gsub(/,\s?/, ', ').gsub('-', '-<wbr/>').html_safe if data.present?
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/holdings/mhld.rb
```diff
@@ -48,7 +48,7 @@ def standard_or_zombie_library
     end
 
     def sanitize_mhld_data(data)
-      CGI.escape_html(data).gsub('),', '), ').gsub('-', '-<wbr/>').html_safe if data.present?
+      CGI.escape_html(data).gsub('),', '), ').gsub(/,\s?/, ', ').gsub('-', '-<wbr/>').html_safe if data.present?
```
cbeer: I'm late to chime in, but couldn't we add a `<wbr />` after commas too and keep the two displays similar?

mejackreed: Yeah that would also work fine too. Though I think we still have to do the thing for librarian view as we aren't sanitizing anything there that is just the value.

@jvine do you have a pref for space vs `<mbr/>` ?

mejackreed: Going with what we have for now.

PR #2050, opened 2018-10-12
updates changelog for recent unreported changes

mejackreed:
~Potentially controversial, as I'm no longer listing full links just referencing the numbers as I think GitHub will auto-magically link them for me.~
Nope I guess they only do this in release notes 😢


jkeck:
Ya, no auto-linking in rando-markdown files unfortunately.

PR #2060, opened 2018-10-17
 Sort managed purls by the sort key or part label

### LINKED ISSUE #2037: Digital Serials - sort
Desired behavior is to order the display the digital object labels in the embed viewer sidebar according to the ordinality specified in the 856, using natural sort.

Example:
856 $x label:December 2010 $x sort:2010.12
856 $x label:February 2010 $x sort:2010.2
856 $x label:January 2010 $x sort:2010.1
856 $x label:July 2015 $x sort:2015.7
 
Displays as:
January 2010
February 2010
December 2010
July 2015

When $x part appears without a number or is absent, sort by $x label. If some 856s in the record have an $x part value and some don’t, the ones with a value should appear sorted at the top of the list. If $x label is not present, default to current behavior.

Justifications 
-this is a display issue, and should be handled in display – the data is there to provide the info the display needs to interpret it properly
-not dependent on a specific backend
-generally, the order of fields in a MARC record is not semantic and should not have a major impact on display
-while data creation is manual, cataloging staff won’t have to redo the ordering every time any digital object attached to the record is versioned in Argo
(These are the same sort outcomes as described in the digital series documentation, just applied in SearchWorks instead of earlier in the pipeline.)

Sample Data on Bodoni
8923346
8810055
4258572
1071767
10551832
12687004

cbeer:
Fixes #2037

### DIFF
--- a/lib/links.rb
+++ b/lib/links.rb
@@ -35,7 +35,7 @@ module SearchWorks
     end
 
     class Link
-      attr_accessor :html, :text, :href, :file_id, :druid, :type
+      attr_accessor :html, :text, :href, :file_id, :druid, :type, :sort
       def initialize(options={})
         @html = options[:html]
         @text = options[:text]
@@ -49,6 +49,7 @@ module SearchWorks
         @druid = options[:druid]
         @ill = options[:ill]
         @type = options[:type]
+        @sort = options[:sort]
       end
 
       def ==(other)
--- a/lib/links.rb
+++ b/lib/links.rb
@@ -26,8 +26,9 @@ module SearchWorks
       all.select(&:sfx?)
     end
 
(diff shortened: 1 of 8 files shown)


lib/links.rb
```diff
@@ -26,16 +26,17 @@ def sfx
       all.select(&:sfx?)
     end
 
+    # sort managed purls by the sort key from the 856$x (with empty values last), then by link text (again with empty values last)
     def managed_purls
-      all.select(&:managed_purl?)
+      all.select(&:managed_purl?).sort_by { |x| [x.sort.present? ? 0 : 1, x.sort, x.text.present? ? 0 : 1, x.text] }
```
jkeck: 😎 Neat, I hadn't thought of doing it like that.

PR #2066, opened 2018-10-22
move refworks on showpage to use post form instead of callback url

### LINKED ISSUE #1944: Refworks export of diacritics
There are several languages with complex diacritics and vernacular characters that are failing to import to RefWorks. 

(Workaround, if you are using the new ProQuest Refworks, you can install the "Save to RefWorks" tool in your browser and that method works.)

https://jirasul.stanford.edu/jira/browse/SW-1977

mejackreed:
This resolves an issue where RefWorks import failed on non-latin like
languages (e.g Russian, Sanskrit). Fixes #1944

### Fixtures to check (failures in current prod):
 - 12786465 Sanskrit
 - 12422691 Russian

One thing I have found is that some records will get imported into RefWorks, but the citation is kind of crappy. I think this is a different issue that could be addressed elsewhere.

refworks

### DIFF
--- a/app/views/shared/record/_record_toolbar_tools_subset.html.erb
+++ b/app/views/shared/record/_record_toolbar_tools_subset.html.erb
@@ -18,7 +18,11 @@
 <% end %>
 <% if @document.export_formats.keys.include?( :refworks_marc_txt ) %>
   <li class="refworks" role="presentation">
-    <%= link_to t('blacklight.tools.refworks_html'), refworks_export_url({url: polymorphic_url(url_for_document(@document), format: :refworks_marc_txt, only_path: false)})  %>
+    <%= form_tag(refworks_export_url) do %>
+      <% # TODO: Remove send when Blacklight::Marc exposes this as a public method %>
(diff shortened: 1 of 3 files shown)


app/views/shared/record/_record_toolbar_tools_subset.html.erb
```diff
@@ -18,7 +18,11 @@
 <% end %>
 <% if @document.export_formats.keys.include?( :refworks_marc_txt ) %>
   <li class="refworks" role="presentation">
-    <%= link_to t('blacklight.tools.refworks_html'), refworks_export_url({url: polymorphic_url(url_for_document(@document), format: :refworks_marc_txt, only_path: false)})  %>
+    <%= form_tag(refworks_export_url) do %>
+      <% # TODO: Remove send when Blacklight::Marc exposes this as a public method %>
```
camillevilla:
```suggestion
      <% # TODO: Remove send when Blacklight::Marc exposes this as a public method projectblacklight/blacklight-marc#64 %>
```
^-- thought it might be nice to throw the issue number in here in case the upstream work doesn't happen by the end of this work cycle

PR #2073, opened 2018-10-24
add support for importing an Article record into RefWorks

### LINKED ISSUE #2058: Allow Article records to be exported via RefWorks and EndNote
RIS data has been available as of ebsco-eds version 1.0.1.  It appears that both EndNote and RefWorks both accept RIS exports (although that is not currently what we're exporting for MARC).

mejackreed:
Fixes #2058 

TODO:
 - [x] Figure out exporting an EDS doc to endnote

If you have a zotero plugin (and assuming Endnote /shrug) it will pickup the export and import it in automagically for you :magic:

### A Good Fixture edsdoj__edsdoj.fbd174b2ea714450a9ea97fce058c880

ris refworks

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -76,6 +76,8 @@ Rails.application.routes.draw do
       concerns :exportable
     end
 
+    get 'articles/:id/ris' => 'articles#show', as: :articles_ris, constraints: ->(req) { req.format = :ris }
+
     post 'articles/:id/track' => 'articles#track', as: :track_articles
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -76,6 +76,8 @@
       concerns :exportable
     end
 
+    get 'articles/:id/ris' => 'articles#show', as: :articles_ris, constraints: ->(req) { req.format = :ris }
```
mejackreed: Had to do this with a crazy constraint instead of `{defaults: { format : ...` so that the right mimetype gets set and picked up

PR #2094, opened 2018-11-01
Allow both article and catalog records to be selected

### LINKED ISSUE #1885: Selection list for articles

jkeck:
Closes #1885 

### TODO
- [x] Address issue with paginating articles
  * We may need a change to the EBSCO gem to handle selections differently. Currently waiting to hear back.  If this isn't going to be possible, we may need to remove certain things from the Article Selections list UI such as sort.
- [x] Address routing/partial overriding for Article selections (particularly for things in `Send to` drop down)
- [x] Add selection counts to tabs
- [ ] Evaluate performance of rendering Article selections (hitting the API 20 times to build a page seems like it may be problematic)
- [x] Add addl. tests as needed
- [x] Update CHANGELOG

## Selections drop down
drop-down

## Catalog selections
catalog-selections

## Article selections
article-selections

### DIFF
--- a/app/controllers/article_selections_controller.rb
+++ b/app/controllers/article_selections_controller.rb
@@ -1,0 +1,58 @@
+class ArticleSelectionsController < ApplicationController
+  include Blacklight::Catalog
+  include Blacklight::Configurable
+  include Blacklight::SearchContext
+  include Blacklight::SearchHelper
+  include Blacklight::TokenBasedUser
+
+  copy_blacklight_config_from(ArticlesController)
+
+  before_action :verify_user
+
+  blacklight_config.http_method = Blacklight::Engine.config.bookmarks_http_method
+  blacklight_config.add_results_collection_tool(:clear_bookmarks_widget)
+
+  blacklight_config.show.document_actions[:bookmark].if = false if blacklight_config.show.document_actions[:bookmark]
+  blacklight_config.show.document_actions[:sms].if = false if blacklight_config.show.document_actions[:sms]
+
+
+  def index
+    @bookmarks = token_or_current_or_guest_user.bookmarks.where(record_type: 'article')
+    bookmark_ids = @bookmarks.collect { |b| b.document_id.to_s }
+
+    eds_params = {
+      guest:          session['eds_guest'],
+      session_token:  session['eds_session_token']
+    }
+
+    if bookmark_ids.present?
+      @response, @document_list = fetch(bookmark_ids)
+    else
+      @document_list = []
+    end
+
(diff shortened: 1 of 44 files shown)


app/controllers/bookmarks_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class BookmarksController < CatalogController
+  include Blacklight::Bookmarks
+
+  # Overidden from Blacklight to add our tabbed interface
+  def index
+    @bookmarks = token_or_current_or_guest_user.bookmarks.where(record_type: 'catalog')
```
jkeck: 🤔 maybe we can inject a default scope to blacklight's `Bookmark` model and prevent us from having to override this action.

app/views/articles/_email_form.html.erb
```diff
@@ -1,9 +1,10 @@
-<%= form_tag email_article_path(@documents.id), :id => 'email_form', :class => "form-horizontal ajax_form", :method => :post do %>
+<%= form_tag email_articles_path, :id => 'email_form', :class => "form-horizontal ajax_form", :method => :post do %>
   <%= render :partial=>'shared/email_form' %>
 
   <%=hidden_field_tag :type, 'brief' %>
-  <!-- TODO: update after article selections  -->
```
jkeck: 😎

app/controllers/article_selections_controller.rb
```diff
@@ -0,0 +1,79 @@
+class ArticleSelectionsController < ApplicationController
```
jkeck: It would not take much convincing to get me to change this to ArticleBookmarks.  It's called "Selections" in the UI, so I started making things called that, but the Blacklight code refers to it as "Bookmarks".  If I do that, I'll prob. squash the commits, so want to get other reviews in first (since I think the commits make it easier to review).

config/routes.rb
```diff
@@ -20,11 +20,19 @@
     concerns :exportable
   end
 
-  resources :bookmarks do
-    concerns :exportable
+  constraints(id: /((?!\/)|(?!articles).)*/) do # EDS identifier rules (e.g., db__acid) where acid has all sorts of different punctuation
```
jkeck: Struggled w/ this a bit.  It kept interpreting `articles` as an identifier in bookmarks.

app/controllers/article_selections_controller.rb
mejackreed: I'm indifferent on the name. I think this is fine.

PR #2097, opened 2018-11-02
Add Iiif badge things

### LINKED ISSUE #2052: IIIF manifest badge in SearchWorks - Application tasks
[#243](https://github.com/sul-dlss/searchworks_traject_indexer/issues/243) will need to be done before putting this into prod. 

Add IIIF badge to appropriate resources in SW
- [x]  SearchWorks records
- [x] search results
- [x] Resource access point 
- [x] masthead 
**_and_** 
- [x] Make it a custom event for google analytics

mejackreed:
Fixes #2052 

iiif


app/views/catalog/_home_page.html.erb
```diff
@@ -62,6 +62,14 @@
             <p>Find books, media, and e-resources set aside for classes.</p>
           </div>
         </div>
+        <div class="panel panel-default">
```
mejackreed: We could add a feature flag here if we don't want to expose this until we have content. I could see arguments either way.

app/assets/javascripts/analytics.js
```diff
@@ -48,4 +48,17 @@ Blacklight.onLoad(function(){
...
+  $('.iiif-dnd').on('click', function(e) {
+    e.preventDefault();
+    var link = $(e.currentTarget)
+    var manifest = link.data().manifest;
+    window._gaq.push(['_trackEvent', 'IIIF DnD', 'clicked', manifest]);
+    window.location = link.attr('href');
+  });
+
+  $('.iiif-dnd').on('dragstart', function(e) {
+    var manifest = $(e.currentTarget).data().manifest;
```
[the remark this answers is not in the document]
mejackreed: I mean.. it "works" but the custom events don't send to Google. So I can't verify that it is actually fully working. Same thing w/ EarthWorks these things don't actually make a request unless its deployed on a server at right url.

app/views/catalog/_home_page.html.erb
mejackreed: Sounds good. Added in a follow on commit so we can revert l8er.

jvine:
> Perhaps @jvine you want to comment on what we call the facet and selected values?

G-search prefers "Available" to "Present". 
`IIIF resources > Available`

I'm wondering if at some point we might want to expand this facet to include the type of object - collections, images, media, etc. - but that's not providing me with any more insight as to what to call it now.

jkeck:
Just took this for a spin, excellent work! ✨🦄 

Should we spin off `Add badge to search results` from #2052 into a separate issue, or do you want to include it in this PR?

mejackreed:
@jkeck thanks! I thought I had added it to search results and was hoping this line would cover it: https://github.com/sul-dlss/SearchWorks/pull/2097/files#diff-0da8ec77361792849e073638f90af850R15

What do you think?

jkeck:
Ah, I think I see where I've confused myself.  I was taking it for a spin but had neglected to reload solr after updating the configs.  It's all looking 😎now.

PR #2103, opened 2018-11-06
Selection error callback

### LINKED ISSUE #1370: Behaviour of "select" and "select all" is flaky
Issue 1. This is really difficult to reproduce, but intermittently when you combine individual and whole-list actions in quick succession (select one, then select all; or clear list, then select 1) the count doesn't increment/decrement properly. Once it's out of sync, actions after that get more incorrect until it starts popping up a modal error message. 

Issue 2. This is easy to reproduce. On an individual item, the "Select" label and the checkbox are not properly associated. 
Click the label Select -  Selections count will increment by 1, but the checkbox is not checked. 
Click the checkbox - Selections count decrements by 1, but the checkbox is now checked and the label says "Selected". 
Now click either the label or the checkbox, and you'll get an error. 

So there's:
1. a Javascript (?) issue where performing an add/remove action on a list that has more or fewer items than expected throws a modal error, and 
2. an HTML issue:  if the label is properly associated with the checkbox, they should behave as one control.

mejackreed:
Fixes #1370 

select custom error


mejackreed:
of course!

PR #2129, opened 2018-11-08
Add the Online h2 to the online access panel.

### LINKED ISSUE #2110: SDR display - Record view - Rework "Online" access panel headings
Split off from #2024 

In our current WIP implementation, we have several different access panels that are considered "Online" panels.
https://github.com/sul-dlss/SearchWorks/blob/master/app/views/catalog/access_panels/_online.html.erb

We still get the "Online" header on MARC records with a WorldCat / Google Books preview panel. We'll probably need to rework this partial....

Sample record: https://searchworks.stanford.edu/view/8655145
screen shot 2018-11-06 at 2 34 17 pm

jkeck:
Closes #2110

### DIFF
--- a/app/views/catalog/access_panels/_online.html.erb
+++ b/app/views/catalog/access_panels/_online.html.erb
@@ -6,6 +6,7 @@
 %>
 
 <div class="panel panel-default access-panel panel-online <%= visible %>">
+  <h2>Online</h2>
   <div class="access-panel-heading panel-heading">
     <h3>
       <% if document.is_a_database? %>


app/views/catalog/access_panels/_online.html.erb
```diff
@@ -6,6 +6,7 @@
 %>
 
 <div class="panel panel-default access-panel panel-online <%= visible %>">
+  <h2>Online</h2>
```
jkeck: @camillevilla, I put this inside the panel that gets displayed/hidden as appropriate w/ the Google links.  Maybe I was miss-understanding the issue, so please let me know if I've missed something.

camillevilla: Huh. Looking back at the original design ticket, I may have taken "Add `<h2>Online</h2>` above `panel-online`" too literally 😅

PR #2133, opened 2018-11-08
Raise a NotFound error when the EDS API returns a BadRequest error

### LINKED ISSUE #2120: Missing records in Articles give 500 error rather than 404 ??
https://searchworks.stanford.edu/articles/edsgao__edsgcl.151592087
https://searchworks.stanford.edu/articles/eric__EJ739770

These articles appear to have been removed from the Ebsco database. They give this error:
screen shot 2018-11-07 at 10 06 49 am

When I would have expected this:
screen shot 2018-11-07 at 10 07 00 am

jkeck:
Closes #2120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 ### Changed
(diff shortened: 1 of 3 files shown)


jkeck:
I'll submit an issue upstream to hopefully get them to return a more specific error for this case that we can capture for.

PR #2135, opened 2018-11-08
tracks external link clicks using Google Analytics

### LINKED ISSUE #2114: Add a tracking event for outbound links
We want to capture:
- where the link goes to (e.g. Yewno, eResource vendor, SFX menu, Worldcat, Google Books, Library services menu, ...)
- the kind of page the link was clicked from (home page, results, record view, Selected databases, no results)

mejackreed:
Fixes #2114 

The URL will be reported which i think is sufficient to accomplish: 

> the kind of page the link was clicked from (home page, results, record view, Selected databases, no results)

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -60,5 +60,21 @@ Blacklight.onLoad(function(){
   $('.iiif-dnd').on('dragstart', function(e) {
     var manifest = $(e.currentTarget).data().manifest;
     window._gaq.push(['_trackEvent', 'IIIF DnD', 'dragged', manifest]);
-  })
+  });
+
+  // Track external link clicks
+  $('a[href]:not([href^="/"],[href^="#"])').on('click', function(e) {
+    e.preventDefault();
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/analytics.js
```diff
@@ -60,5 +60,21 @@ Blacklight.onLoad(function(){
   $('.iiif-dnd').on('dragstart', function(e) {
     var manifest = $(e.currentTarget).data().manifest;
     window._gaq.push(['_trackEvent', 'IIIF DnD', 'dragged', manifest]);
-  })
+  });
+
+  // Track external link clicks
+  $('a[href]:not([href^="/"],[href^="#"])').on('click', function(e) {
```
mejackreed: I _think_ this should be the correct selector. I'm not sure if there are any other considerations we should make here?

jkeck: This will end up tracking any time we use a fully qualified URL?  I don't think we use that a lot, but just want to make sure I know the considerations we may have to take moving forward.

PR #2147, opened 2018-11-09
Fixes #2136 Hoover Archive items are only in process if no finding is…

### LINKED ISSUE #2136: Hoover Archives Requests should not default to "In process"
Hoover Archives records _with_ Finding aid should _not_ default to Red X - In process. Only those without the Red X. 

Currently on stage
screen shot 2018-11-08 at 2 42 01 pm

Should still be Orange truck - In-library use
screen shot 2018-11-08 at 3 00 27 pm

mejackreed:
… present

## Finding aid present 6658571

screen shot 2018-11-09 at 8 01 11 am

## Finding aid missing 12822793
screen shot 2018-11-09 at 8 01 15 am

### DIFF
--- a/lib/holdings.rb
+++ b/lib/holdings.rb
@@ -50,7 +50,7 @@ class Holdings
   def callnumbers
     return [] unless @item_display.present?
     @callnumbers ||= @item_display.map do |item_display|
-      Holdings::Callnumber.new(item_display)
+      Holdings::Callnumber.new(item_display, document: @document)
     end.sort_by(&:full_shelfkey)
   end
   def mhld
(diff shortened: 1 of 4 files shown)


lib/holdings.rb
```diff
@@ -50,7 +50,7 @@ def libraries
   def callnumbers
     return [] unless @item_display.present?
     @callnumbers ||= @item_display.map do |item_display|
-      Holdings::Callnumber.new(item_display)
+      Holdings::Callnumber.new(item_display, document: @document)
```
mejackreed: Hoping this isn't too sketchy

jkeck: It's all just pointers right? 😅

PR #2162, opened 2018-11-13
Remove document accessor from json response.

jkeck:
document was added as an accessor in 02f4d26d34c227e065a23c77b278fe48b901f02e.
However, when we render availability as json this puts a SolrDocument in every item in the holdings.
For records with a lot of items this blows that json response's size up to a huge amount.

### DIFF
--- a/lib/holdings/callnumber.rb
+++ b/lib/holdings/callnumber.rb
@@ -136,7 +136,7 @@ class Holdings
     end
 
     def as_json(*)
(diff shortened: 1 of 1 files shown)


cbeer:
Can you add a reason for this change to the description or commit comment?

jkeck:
@cbeer better?

cbeer:
👍 Thanks

PR #2176, opened 2018-11-15
Fixes #2132 managed purl alignment issue

### LINKED ISSUE #2132: Split off - Fix single managed PURL alignment
There is one issue I'm noticing for normal managed PURL records (e.g. does not have multiple), which may have already existed and is just more obvious w/ the new design, is some odd right-column alignment.

screen shot 2018-11-07 at 10 22 31 pm

If anybody wants to take a look at that with, that would be great.  We could also split that off into a separate issue.

Split off #2131

mejackreed:
## Before
screen shot 2018-11-15 at 4 10 24 pm

## After
screen shot 2018-11-15 at 4 11 34 pm

### DIFF
--- a/app/assets/stylesheets/modules/record-metadata-section.scss
+++ b/app/assets/stylesheets/modules/record-metadata-section.scss
@@ -9,7 +9,8 @@
 }
 
 .record-metadata,
-.upper-record-metadata {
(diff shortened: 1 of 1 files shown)


camillevilla:
📏🖼✅

PR #2178, opened 2018-11-15
SDR display - Record view - Update Context Panels

### LINKED ISSUE #2108: SDR display - Record view - Update Context Panels
Split out from #2024

## 'Context' panels

- [x] Group together these two sections: `panel-appears-in` and `panel-in-collection`. When either or both appears, display `<h2>Context</h2>` at the top.

### `panel-in-collection` section
- [x] Change the `<h3>` in the panel header to `Item belongs to a collection`
- [x] Remove the body/footer separation, so there's no line between the title/summary and `<dl>`. (Maybe move the  `<dl>` into the body section, and add 15px margin to the top of it.) 
- [x] Change `<dt>Digital content</dt>` to `<dt>Digital collection</dt>`
- [x] Change `<dd>N items</dd>` to `<dd>N digital items</dd>`
- [x] Change `<dt>[Resource type]</dt>` to `<dt>Physical collection</dt>`
- [x] Remove the Collection PURL `<dt>` and `<dd>` pair
screen shot 2018-10-25 at 6 38 26 pm

### `panel-appears-in` section
- [x] Change the `<h3>` in the panel header to `Item is included in another record`

The two panels together panels look like this
screen shot 2018-10-25 at 5 01 38 pm

camillevilla:
Closes #2108

## Before
screen shot 2018-11-16 at 2 47 58 pm
screen shot 2018-11-16 at 2 48 04 pm

## After
screen shot 2018-11-16 at 2 46 27 pm
screen shot 2018-11-16 at 2 46 23 pm

### DIFF
--- a/app/views/catalog/record/_metadata_panels.html.erb
+++ b/app/views/catalog/record/_metadata_panels.html.erb
@@ -1,11 +1,16 @@
+<%
+  appears_in = render "catalog/access_panels/appears_in"
+  in_collection = render "catalog/access_panels/in_collection"
+%>
+
 <div class="metadata-panels">
   <%= render "catalog/access_panels/online" %>
   <%= render "catalog/access_panels/course_reserve" %>
   <%= render "catalog/access_panels/location" %>
-  <% if (@document.set_member? && @document.parent_sets.present?) || (@document.is_a_collection_member? && @document.parent_collections.present?) %>
+  <% if appears_in.present? || in_collection.present? %>
     <h2>Context</h2>
   <% end %>
-  <%= render "catalog/access_panels/appears_in" %>
-  <%= render "catalog/access_panels/in_collection" %>
+  <%= appears_in %>
+  <%= in_collection %>
   <%= render "catalog/access_panels/related" %>
 </div>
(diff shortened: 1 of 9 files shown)


app/views/catalog/record/_metadata_panels.html.erb
```diff
@@ -2,6 +2,9 @@
   <%= render "catalog/access_panels/online" %>
   <%= render "catalog/access_panels/course_reserve" %>
   <%= render "catalog/access_panels/location" %>
+  <% if (@document.set_member? && @document.parent_sets.present?) || (@document.is_a_collection_member? && @document.parent_collections.present?) %>
```
jkeck: Is this to try and only render context if `appears_in` or `in_collection` has content?  If so, perhaps we can render both of those into a variable and check that one of them has content.  This way, if the logic w/i one of those partials changes, we don't have to remember to update the logic in this partial to match.

camillevilla:
👍 Thanks for showing me this neat pattern, @jkeck

https://github.com/sul-dlss/SearchWorks/blob/c4b4e0a85da6660036d13d58b9b784daaafec896/app/views/catalog/record/_marc_metadata_sections.html.erb#L1-L6

PR #2193, opened 2018-11-19
Add a 1px transparent border around selection nav buttons.

### LINKED ISSUE #2188: Selections selector on hover seems a little wonky
When looking at the selections stuff, it seems like maybe the hover style is a bit off?

screen shot 2018-11-19 at 6 22 10 am

jkeck:
Fixes #2188


app/assets/stylesheets/modules/bookmarks.scss
```diff
@@ -32,4 +32,8 @@ label {
     border-bottom: 0;
     display: inline;
   }
+
+  a:hover {
```
mejackreed: Should this also be on focus and active too?

PR #2181, opened 2018-11-16
[WIP] switch from Dejavu EL -> Condensed Book fixes #2167

mejackreed:
_**[WIP]** while we figure out some inconsistencies w/ some diacritics in particular browsers._  

Fixes #2167 
Closes #2204 

screen shot 2018-11-16 at 8 47 05 am
screen shot 2018-11-16 at 8 46 44 am

Issue #2211, opened 2018-11-26
IC chat widget exposed to non-Stanford users


jvine:
Report from IC that non-Stanford users are accessing chat reference. On the no-results page, when I am not on VPN and not logged in, I see Chat and Login links when I should _only_ see the Login link. 

<img width="400" alt="screen shot 2018-11-26 at 3 50 36 pm" src="https://user-images.githubusercontent.com/6945691/49049368-92316f80-f193-11e8-8886-dca03975395b.png">

PR #2213, opened 2018-11-27
Use Google Noto Sans font for titles

### LINKED ISSUE #2167: Replace DejaVu Sans Extralight with DejaVu Sans Condensed Book
This font still seems to be the best choice for display of diacritics and ligatures for non-English languages, but the Extralight version was a poor choice for readability in the results view.

https://www.fontsquirrel.com/fonts/dejavu-sans

mejackreed:
Fixes #2167 

Uses Noto Sans font instead of DejaVuSans.

One thing to note here, I found out what the previous issue was with the other DejaVu change #2181 . It seems that both DejaVuSansCondensed Book and Noto Sans do not have the "COMBINING DOUBLE INVERTED BREVE" character `U+0361` or `&#865;` This character found in the title of record 440849 should be placed over the character to its left and right. In this case, the `i` and the `e`.

See https://www.fileformat.info/info/unicode/char/0361/index.htm

```javascript
"ui͡e".charCodeAt(0)
117
"ui͡e".charCodeAt(1)
105
"ui͡e".charCodeAt(2)
865
"ui͡e".charCodeAt(3)
101
```

For Noto Sans, this glyph is not included and so it will fallback to other fonts.

```diff
-  font-family: 'Noto Sans', 'Arial Unicode MS', Helvetica, sans-serif;
+  font-family: 'Noto Sans', Helvetica, sans-serif;
```

It is also important to note that I removed the `Arial Unicode MS` fallback. I think that was the source of the misplaced character as the spacing displayed the diacritical character in the wrong place.

### Fallback to Arial Unicode MS
screen shot 2018-11-27 at 6 24 46 am

### Fallback to Helvetica, sans-serif only
screen shot 2018-11-27 at 6 24 39 am

One issue presents is that Firefox does display this particular character differently

### Noto Sans in Firefox
screen shot 2018-11-27 at 6 43 12 am

Something similar is currently happening with DejaVu so I'm not too worried about it, but wanted to call it out.


mejackreed:
My understanding is that no, we do not need to include that in the font-family property. I'm going by the Google Font installation instructions:

screen shot 2018-11-27 at 11 52 56 am

camillevilla:
FYI: we'll deploy this to stage for testing after #2211 is resolved.

jkeck:
2noxcg

camillevilla:
We're cutting bait on a font change. 🤞 we have better options for webfonts + Russian diacritics next work cycle.

PR #2214, opened 2018-11-27
Update to rails 5.2.1.1

jkeck:


jkeck:
Will let regular dep updates take care of this instead.

PR #2220, opened 2018-11-29
Update Rubocop version and TODOs

camillevilla:
Part of https://github.com/orgs/sul-dlss/projects/2#card-15037933

The todo file is pretty long but (!)doesn't contain deprecated methods. A lot of it can get picked off by `--auto-correct`. 

I've picked off about 3000 offenses for now. That leaves us with... 10898 to go! And some 4000 of those are string literals.


lib/constants.rb
```diff
@@ -667,7 +667,7 @@ module Constants
     'W7-STFSHD' => 'Request for use at Map Center (W7 shadowed)',
     'W7-FRAME' => 'Request for use at Map Center (W7 framed)',
     'W7-POCKET' => 'Request for use at Map Center (W7 pocket)',
-    'WELLNESS' =>' Wellness wall',
+    'WELLNESS' => ' Wellness wall',
```
cbeer: Unrelated to the PR, but is the leading space in the string value intentional?

PR #2221, opened 2018-11-30
Remove support for the marcbib_xml field

cbeer:


cbeer:
We should hold on this until #2222 is reverted.

PR #2222, opened 2018-11-30
Revert "Use structured data fields instead of parsing raw marc" for m…

cbeer:
…arc links

This reverts part of commit 41cec04a6cec94d800563641cfce70d424251583.


cbeer:
This is a temporary fix until https://github.com/sul-dlss/searchworks_traject_indexer/pull/287 is merged and the affected records are reindexed.

PR #2229, opened 2018-12-05
logger 1.2.8 was yanked.

cbeer:
This might make searchworks deployable; logger 1.2.7 might also turn out to be total garbage.

https://github.com/ebsco/edsapi-ruby/pull/89 is probably a better fix.


cbeer:
There's a significant delta between logger 1.2.7 and 1.2.8, so this probably isn't a good idea.

PR #2289, opened 2019-05-09
Swap chromedriver-helper for webdrivers

cbeer:


spec/spec_helper.rb
```diff
@@ -13,14 +13,7 @@
 require 'fixtures/marc_records/marc_metadata_fixtures'
 require 'fixtures/mods_records/mods_fixtures'
 
-Capybara.javascript_driver = :headless_chrome
-
-Capybara.register_driver :headless_chrome do |app|
-  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
-    chromeOptions: { args: %w[headless disable-gpu no-sandbox window-size=1000,700] }
```
mejackreed: I'm wondering if us losing the size is causing some of the feature specs to fail. Is there a way to configure this?

jcoyne: I had the same problem in argo. I did this: https://github.com/sul-dlss/argo/pull/1402/files#diff-93830fa29d616f7c87903d08b5b1b29aR18

Issue #2150, opened 2018-11-09
Improve TOC metadata styling?


cbeer:
Current implementation:
<img width="533" alt="screen shot 2018-11-09 at 10 24 50" src="https://user-images.githubusercontent.com/111218/48280955-00430c00-e40a-11e8-9e2c-c7dd8662d53b.png">

Naive, new implementation:
<img width="752" alt="screen shot 2018-11-09 at 10 27 33" src="https://user-images.githubusercontent.com/111218/48280983-19e45380-e40a-11e8-9504-1a483bd4214d.png">

Replicating the current behavior is a little tricky, so I'm wondering if this is an opportunity to do something better with those digits (maybe an ISBN and some random date?). I may be wrong, but it may be something Nielsen-specific

cbeer:
Answer: suppress them.

PR #2308, opened 2019-05-28
Use the summary_data partial on the record view…

### LINKED ISSUE #2307: Handle Summary Struct Appropriately
See: https://searchworks.stanford.edu/view/3786532

jkeck:
…(already used in results).

Fixes #2307 

## Before
3786532-before

## After
3786532-after

(Note: @jvine approved the label change as it now matches what is in the results)

### DIFF
--- a/app/views/catalog/record/_marc_contents_summary.html.erb
+++ b/app/views/catalog/record/_marc_contents_summary.html.erb
@@ -46,10 +46,12 @@
   </dd>
 <% end %>
 
-
-<% summary = document.fetch(:summary_struct, []).first %>
-<% unless summary.nil? %>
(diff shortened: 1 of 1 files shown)


jvine:
What are those numbers at the end and are they meant to display? I thought we were meant to suppress them.

jkeck:
Hmm, not really sure, but they are showing in the results too FWIW.

jvine:
This may not be considered part of this work, but gryphon-search says the numbers should not be displaying.

cbeer:
We did this for TOC, but summary: https://github.com/sul-dlss/SearchWorks/issues/2150. I'm hearing we should also suppress $x from summaries?

jvine:
Yes.

PR #2351, opened 2019-09-27
Proof of concept on the fly sitemap generation built during Blacklight-LD meeting

mejackreed:
Paired on by:
@magibney
@mejackreed
@agazzarini
@netsensei

Todos / ?:
 - Test this at scale
 - Extract to a gem?
 - Are there any concerns about the `url_for` with a hostname in a load balanced scenario? (I don't seem to remember)
 - Validate the show action id param to make sure that the length is at least the `access`
 - Better variable names

### DIFF
--- a/app/controllers/sitemap_controller.rb
+++ b/app/controllers/sitemap_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class SitemapController < ApplicationController
+  def index
+    @access_list = access_list
+  end
+
+  def show
+    @solr_response = Blacklight.default_index.connection.select({
+      params: {
+        q: "{!prefix f=hashed_id_si v=#{access_params}}",
+        fl: 'id,last_updated'
+      }
+    })
+  end
+
+  private
+
+  def access_params
+    params.require(:id)
+  end
+
+  def max_documents
+    Rails.cache.fetch('index_max_docs', expires_in: 1.day) do
+      Blacklight.default_index.connection.select({ params: { q: '*:*', rows: 0 } })['response']['numFound']
+    end
+  end
+
+  def access_list
+    average_chunk = [10000, max_documents].min # Sufficiently less than 50,000 max per sitemap
+    access = (Math.log(max_documents / average_chunk) / Math.log(16)).ceil
+    (0...(16**access))
+      .to_a
+      .map { |v| v.to_s(16).rjust(access, '0') }
+  end
+end
--- a/app/views/sitemap/index.xml.builder
+++ b/app/views/sitemap/index.xml.builder
@@ -1,0 +1,12 @@
+xml.instruct! :xml, version: '1.0', encoding: 'UTF-8'
+xml.sitemapindex(
+  'xmlns:xsi' => 'http://www.w3.org/2001/XMLSchema-instance',
+  'xsi:schemaLocation' => 'http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd',
+  'xmlns' => 'http://www.sitemaps.org/schemas/sitemap/0.9'
+) do
+  @access_list.each do |id|
+    xml.sitemap do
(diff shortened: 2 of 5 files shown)


cdmo:
I know this if for searchworks, not blacklight, but, I gave this a try locally because I was curious and it is not working. The index view renders like below

```xml
<sitemapindex xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/siteindex.xsd">
<sitemap>
<loc>http://localhost:3000/sitemap/0</loc>
</sitemap>
</sitemapindex>
```

But the `show` does not. Just an empty `<urlset>`. I think that the update dedupe action isn't actually adding the `signatureField`. Is that the expectation? Like, when you run an update it should also go through and generate a new field (`hashed_id_ss`) for every record?

magibney:
Yeah if you're not getting a `hashed_id_ss` field generated for every record, then none of the rest of this will work. Did you configure all the relevant aspects of `solrconfig.xml` and `schema.xml`?

agazzarini:
I see from the following schema snippet

```xml
...
<dynamicField name="*_si" type="string" stored="true" indexed="true" omitNorms="true" />
  ...
```

that the signature field "hashed_id_si" is indexed and stored, so that means if you run a query in Solr you should see each returned document with that new field. 
Could you please you confirm that? I'm checking the solrconfig.xml in the patch and it seems ok

cdmo:
Yeah, I do have that dynamic field in my schema.xml and I believe my solrconfig.xml has the reflected changes too. What will trigger the dedupe chain specifically? [Here's my branch](https://github.com/psu-libraries/psulib_blacklight/tree/on-demand-sitemap) (it is messy a bit, sorry, but you can compare to master for changes).

I tried running fresh indexes, tried running normal update commits and tried `/solr/psul_blacklight/update/json?update.chain=dedupe&commit=true` too. I never did see the `hashed_id_ss` show up in any record, which sounds like I should have. Maybe I have something conflicting in my solrconfig.

magibney:
Shouldn't you be looking for `hashed_id_ssi`, not `hashed_id_ss`? Perhaps that was just a typo, but the dynamic field and the signatureField configured in your schema.xml and solrconfig.xml are `_ssi` ...

cdmo:
@magibney sorry - I deleted my message because I realized the branch was not quite right, will report back soon

cdmo:
Ok, so yes, I am using `hashed_id_ssi` because `*_si` is not stored in my `schema.xml`, and, happily, I can generate the `hash_id_ssi` - it shows up in records like `hashed_id_ssi":"81524bef2092a2df",`

[Here's my branch](https://github.com/psu-libraries/psulib_blacklight/tree/on-demand-sitemap)

Still the same issue though, nothing in the urlset on the show view.

I wasn't sure if I needed a signatureField in my schema.xml. It wasn't on this PR, but, documentation seemed to allude to needing to do this.

I guess I don't see how this works:
```rb
    @solr_response = Blacklight.default_index.connection.select({
      params: {
        q: "{!prefix f=hashed_id_ssi v=#{access_params}}", # changed f to my signature field
        fl: 'id,last_updated'
      }
    })
```
Or how it's supposed to work.

cdmo:
It looks like the hashed_id generated as the signatureKey should be associated with a sitemap `id` (not to be confused with a record `id` - I might suggest calling the sitemap `id` `sitemap_id` or something). Otherwise, how does Solr pull up a list of documents just by passing in the sitemap id in the show method for the sitemap controller? Again, thanks for any help.

cdmo:
With generous help from @magibney and @tampakis too - I was able to get this working. The most important missing bit was that the `defType` had to be `lucene`. I now have a working POC for our Blacklight instance at Penn State for a few hundred records, next I'll try a few million locally, and after that we'll test it out on a test server environment. Thank you all for your help.

mejackreed:
Reimplemented in a gem here: https://github.com/sul-dlss/blacklight_dynamic_sitemap/pull/2#discussion_r365479149

Issue #2365, opened 2019-12-02
Bad vernacular matches cause an endless lookup loop


jkeck:
Still figuring out the exact cause, but we're seeing some data that looks like this:

610 2 0 $6 610-00 $a Sub A Content $b Sub B Content

That $6 should be pointing to an 880, not a 610.  This is bad data, but we need to ensure our MarcField class can handle this w/o endlessly spinning.

jkeck:
Closed by #2366 & #2367

PR #2367, opened 2019-12-02
Teach MarcFieldWrapper to not identify fields that have a …

jkeck:
…non-compliant vernacular matching field as a vernacular field.

Part of #2365

### DIFF
--- a/spec/fixtures/marc_records/marc_metadata_fixtures.rb
+++ b/spec/fixtures/marc_records/marc_metadata_fixtures.rb
@@ -297,6 +297,10 @@ module MarcMetadataFixtures
           <subfield code="6">880-03</subfield>
           <subfield code="a">This is not Vernacular</subfield>
         </datafield>
+        <datafield tag="610" ind1="1" ind2="0">
+          <subfield code="6">610-00</subfield>
+          <subfield code="a">Bad Venacular Matcher</subfield>
+        </datafield>
         <datafield tag="700" ind1=" " ind2=" ">
           <subfield code="6">880-04</subfield>
           <subfield code="a">This is not Vernacular</subfield>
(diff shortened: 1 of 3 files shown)


jkeck:
@shelleydoljack is it true that all normal MARC fields that are pointing to a vernacular field's $6 should point to some `8xx` field (and vice versa w/ the $6 in the 880 itself)?

spec/fixtures/marc_records/marc_metadata_fixtures.rb
```diff
@@ -297,6 +297,10 @@ def bad_vernacular_fixture
           <subfield code="6">880-03</subfield>
           <subfield code="a">This is not Vernacular</subfield>
         </datafield>
+        <datafield tag="610" ind1="1" ind2="0">
+          <subfield code="6">610-00</subfield>
+          <subfield code="a">Bad Venacular Matcher</subfield>
+        </datafield>
```
shelleydoljack: Yes, this is true as being not valid (a bad vernacular matcher).

spec/fixtures/marc_records/marc_metadata_fixtures.rb
```diff
@@ -325,6 +329,10 @@ def bad_vernacular_fixture
           <subfield code="a">710-05</subfield>
           <subfield code="6">This is Vernacular</subfield>
         </datafield>
+        <datafield tag="880" ind1=" " ind2=" ">
+          <subfield code="6">880-00</subfield>
+          <subfield code="6">This is A vernacular matching field that does not point to a valid field</subfield>
+        </datafield>
```
shelleydoljack: Yes, this is also a bad vernacular matcher.

PR #2382, opened 2020-01-30
Adds a custom event for when exhibit links are clicked fixes #2381

### LINKED ISSUE #2381: Add a custom event for when people link to exhibits

mejackreed:

### DIFF
--- a/app/assets/javascripts/exhibitPanel.js
+++ b/app/assets/javascripts/exhibitPanel.js
@@ -126,6 +126,13 @@
       if (exhibit.subtitle && exhibit.subtitle !== '') {
         heading.append(exhibit.subtitle)
       }
+      // Exhibit link clicked, add a custom event
+      wrapper.find('a').on('click', function(e) {
+        ga('send', 'event', 'Exhibit link', 'clicked', heading.text(), {
+          'transport': 'beacon'
+        });
+      });
+
       return wrapper;
     }
   }


app/assets/javascripts/analytics.js
```diff
@@ -124,6 +124,13 @@ Blacklight.onLoad(function(){
     });
   });
 
+  // Exhibit link clicked
+  $('.panel-exhibit a').on('click', function(e) {
+    ga('send', 'event', 'Exhibit link', 'clicked', $(e.currentTarget).text().trim(), {
```
jkeck: Would there be any use/desire in distinguishing between a users click to navigate to an exhibit vs. toggling to see more exhibits?

mejackreed: Do you have any markup / screenshots of that toggle functionality? I couldn't figure that out exactly.

jkeck: SHPC (Ckey 4769498) is in multiple exhibits in production (and is currently configured to display in stage).

Also happy to ship this, as more work is likely to have to happen to that toggle before we roll this out and addl. GA work can be done there.  Up to you (don't mean to block this PR)

exhibit-toggle

mejackreed: Thanks that got me rolling. I ended up just modifying the approach to be a bit more stable with the async nature of this. No we are only tracking exhibit title clicks which is probably what we want. 🤷‍♂

PR #2385, opened 2020-01-31
Update exhibit toggle link to a button.

### LINKED ISSUE #2384: Style "see all # Exhibits" link to match equivalent SW control
Call number expand button looks like this:
Screen Shot 2020-01-30 at 6 08 05 PM

Exhibits expand link looks like this:
Screen Shot 2020-01-30 at 6 12 57 PM
with a different visual style and wording.

Preferred style:
Screen Shot 2020-01-30 at 6 09 38 PM

- button is a more appropriate element than link for this action
- change text from "see all" to "show all" for consistency
- add 15px margin above the button to separate it from the last item in the list

(And it feels odd that "show less" is missing, though it might not be critical.)

jkeck:
Closes #2384 

## Before
link-before

## After
button-after

### DIFF
--- a/app/assets/javascripts/exhibitPanel.js
+++ b/app/assets/javascripts/exhibitPanel.js
@@ -18,7 +18,7 @@
       if (panel.find('.media').length === 0) { // Only run the fetch if the panel has no exhibits yet
         this.fetchExhibits();
       } else {
-        this.addToggleLinkBehavior(); // Add toggleLinkBehavior if inited with exhibits already (e.g. back button click)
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/exhibitPanel.js
```diff
@@ -81,12 +81,12 @@
...
           }
         });
-        toggleLink.on('click.see-all-exhibits-link', function(e) {
+        toggleButton.on('click.see-all-exhibits-button', function(e) {
           e.preventDefault();
 
           // We don't have good text for a toggle off link.
           // When we do we can remove this and update the text appropriately
-          toggleLink.remove();
+          toggleButton.remove();
```
camillevilla: 👍

PR #2389, opened 2020-02-10
Disable the Hours API request in test.

jkeck:


config/settings.yml
```diff
@@ -1,4 +1,6 @@
-HOURS_API: <%= ENV['HOURS_API'] || 'http://example.com' %>
+HOURS_API:
```
jkeck: This will necessitate a change to shared_configs w/ this.  I'll be happy to add those changes if/when we roll this out.

PR #2412, opened 2020-03-17
Don't use recaptcha 5.4.0

jkeck:
ambethia/recaptcha#356 aims to address this issue, and will hopefully be released as 5.5.0

This is causing reCaptcha restricted emails to not be sent (even w/ a valid recaptcha challenge)


anarchivist:
Ref VUF-7619

grosser:
5.4.1 should work

jkeck:
Thanks @grosser

PR #2416, opened 2020-03-19
Add a task to prune old guest users who do not have bookmarks.

jkeck:


lib/tasks/searchworks.rake
```diff
@@ -74,6 +74,17 @@ namespace :searchworks do
     end
   end
 
+  desc "Prune old guest users from the database"
+  task :prune_old_guest_user_data, [:months_old] => [:environment] do |t, args|
+    old_bookmarkless_guest_users_ids = User.includes(:bookmarks)
```
jkeck: Like the prune searches task, this likely won't be performant when things have gotten out of hand, but should be useful as a daily cleanup task to keep things under control.

thatbudakguy (approved):

PR #2437, opened 2020-04-17
Remove eReserves from 'At the Library' panels

### LINKED ISSUE #2436: Remove eReserve "At the Library" block
Currently we have a eReserve block showing up when an item has a Home Location = INSTRUCTOR, Current Location = E-RESV

image

Example records:
https://searchworks.stanford.edu/view/12378453
https://searchworks.stanford.edu/view/12958161
(and basically any record found in the Course Reserve catalog right now.)

999 example:
image

The information in the "At the Library" block for the eReserves item record is really confusing to the user, and there isn't any helpful information being displayed.  The item is always available (for now - there is no digital lending where it might not be available), and the user needs to just use the information in the "Available online" block. 

Can we just remove the "At the Library" section for all items in E-RESV?

cbeer:
Fixes #2436

@saseestone : I assume/hope we also want to remove similar panels with this data (like under Check Availability on search results, in the call number bar for the gallery view, etc)?

### DIFF
--- a/lib/holdings/library.rb
+++ b/lib/holdings/library.rb
@@ -28,11 +28,15 @@ class Holdings
     end
 
     def holding_library?
-      !zombie?
+      !zombie? && !eresv?
     end
 
     def zombie?
       @code == 'ZOMBIE'
+    end
+
+    def eresv?
+      @code == 'E-RESV'
     end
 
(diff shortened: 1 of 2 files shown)


saseestone:
Yes, remove everywhere.

PR #2438, opened 2020-04-17
Exclude E-RESV at the holding level

### LINKED ISSUE #2436: Remove eReserve "At the Library" block

cbeer:
Actually fixes #2436

### DIFF
--- a/lib/holdings/callnumber.rb
+++ b/lib/holdings/callnumber.rb
@@ -16,7 +16,7 @@ class Holdings
 
     def present?
       @holding_info.present? &&
-        !(item_display[1] == 'SUL' && internet_resource?) &&
+        !(item_display[1] == 'SUL' && (internet_resource? || eresv?)) &&
         !bound_with?
     end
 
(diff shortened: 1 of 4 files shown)


lib/holdings/callnumber.rb
```diff
@@ -151,6 +151,10 @@ def internet_resource?
       home_location == 'INTERNET'
     end
 
+    def eresv?
```
jkeck, line 154: One random thought would be to add this case somehow to `internet_resource?` so that E-RESV items were also considered internet resources.  This would have the benefit of not needing a change in `present?` as well as label the resulting item in the call number browse as `eResource` instead of `(no call number)`.  It could be a slight improvement (but should def. not be considered a blocker)

cbeer: Yeah, I'm still not clear whether they're the same case or just suspiciously similar.

PR #2449, opened 2020-04-28
Allow users to report URLs in their feedback and record emails.

jkeck:
Now that we're only allowing Captcha authenticated anonymous users to submit feedback/record emails so it seems like we can ease the rule about blocking content with URLs as spam.

I'm interested in the take of others (in particular those who manage the feedback @trapido / @jlkalchik) on allowing users to send links via feedback now.  It would be tough to say what legitimate spam is being blocked by that filter (my assumption would be zero as it would get stopped by the Captcha).  I'm guessing we're only seeing actual users being frustrated by this filter now, and we can/should remove it.

I'm also not sure if we can/should provide more guidance to the user that we report the URL of the page they submit the feedback/etc from (because it seems like more often than not that's the URL they're trying to report to us anyway).

### DIFF
--- a/app/controllers/concerns/email_validation.rb
+++ b/app/controllers/concerns/email_validation.rb
@@ -5,8 +5,6 @@ module EmailValidation
     case
     when !%w(full brief).include?(params[:type])
       flash[:error] = I18n.t('blacklight.email.errors.type')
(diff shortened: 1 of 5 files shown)


saseestone:
Irina and Jeanette have confirmed that adding urls would be great.

Issue #2455, opened 2020-05-07
ETAS vs. Publicly Available Hathi Content


jkeck:
This issue is to document / discuss / track setting progress on pulling Hathi content that is accessible outside of the ETAS program into links that can / will remain accessible in records once we are no longer eligible for ETAS.

One current issue is, while we don't apply the Stanford S icon for HathiTrust links that will be publicly available, we are forcing users through a Stanford login.  We could easily remove the authentication for these links, however; then questions arise about why the link would be in a "Temporary Access" block (along w/ the notices that are available there).

A proposal to incrementally improve this:
- [x] Add Stanford S icons to all temporary access links (we can do this immediately to remove any confusion that users may have about this)
  * #2456
- [x] Refactor a HathiLinks (or something like that) class out of TemporaryAccess that can return publicly available links (for use in the Online access panel) and also return the TempAccess links that are only going to be available through ETAS.
  * The reason why this _may_ be useful to be in the SearchWorks codebase is the ability to choose if we should present a link to a work or item URL based on both the number of holdings and the other full-text/SFX links available.  This could theoretically be done in the index, but we'll likely be duplicating application code.
  * #2459
- [x] Update the indexing code to ensure that public domain documents will show up as Online in the access facet (even after we turn ETAS off via feature flag and/or more permanently).
  * sul-dlss/searchworks_traject_indexer#444

PR #2459, opened 2020-05-13
Refactor HathiTrust Publicly Accessible Links to Online Panel

jkeck:
Part of #2455 

The goal for this is to have no change to existing links that are truly Stanford only, and that the public domain links can be pulled into the regular Online section w/o forcing the user through Stanford auth.

## 8935702 Before
8935702-before

## 8935702 After
8935702-after

### DIFF
--- a/app/models/hathi_trust_links.rb
+++ b/app/models/hathi_trust_links.rb
@@ -42,10 +42,15 @@ class HathiTrustLinks
   end
 
   def work_url
-    "https://catalog.hathitrust.org/Record/#{hathitrust_work_id}?signon=swle:#{stanford_shib_id}"
+    work_base_url = "https://catalog.hathitrust.org/Record/#{hathitrust_work_id}"
+    return work_base_url if publicly_available?
+
+    "#{work_base_url}?signon=swle:#{stanford_shib_id}"
   end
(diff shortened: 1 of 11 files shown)


jvine:
This looks accurate to me!

app/models/hathi_trust_links.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def present?
+    (hathitrust_work_id.present? || hathitrust_item_id.present?) && !fulltext_available?
+  end
+
+  # According to https://www.hathitrust.org/hathifiles_description access will be "allow"
+  # even in the case that the item is copyrighted in the US, or only Public Domain in the US
+  # (and therefore not available to users outside the US).
+  def publicly_available?
```
jkeck, line 14: We could flip this to fit more of the Stanford domain to be `stanford_only?` (as we basically reverse this when we check it elsewhere).  It was just a smaller change to keep the previous name.

cbeer: 👍 ; I think `publicly_available?` probably makes sense in this context (and once ETAS is over, the stuff that's SU only won't be available)

Issue #2466, opened 2020-05-26
Update request link logic for paging service roll-out


jkeck:
Proposal:
- [x] Remove existing request links for all materials
- [x] Add new configurable request link logic that meets the following criteria
  - Can be configured to be enabled (at the location level) for a custom set of home locations.
  - Should be able to exclude links from valid locations that only contain non-circulating items
  - Should be able to exclude links from valid locations based on other aspects of the document (holdings, metadata, etc).
  - Possibly add something in the availability API to allow the requests app to prevent users from selecting non-circulating material for request.
  - ...
- [x] Enable this for Green + Locations TBD
- [x] Deploy out for review

Questions
- SSRC Requests?
- Hold/Recalls?
- Changes to requests?

PR #2467, opened 2020-05-29
Update Request Link logic for re-opening of paging services

jkeck:
Part of #2466

### DIFF
--- a/app/helpers/request_link_helper.rb
+++ b/app/helpers/request_link_helper.rb
@@ -57,7 +57,7 @@ module RequestLinkHelper
   end
 
   def hoover_request_url(document, callnumber)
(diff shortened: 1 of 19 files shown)


app/helpers/request_link_helper.rb
```diff
@@ -57,7 +57,7 @@ def link_data_attributes(callnumber)
   end
 
   def hoover_request_url(document, callnumber)
-    HooverOpenUrlRequest.new(callnumber.library, document, self).to_url
```
jkeck, line 60: This helper code is all dead now, but keeping it around for a bit for reference while we transition.  We'll eventually want to 🔪 the entire helper.

PR #2508, opened 2020-07-29
Handle bad (PHP-style array) params in Stanford Work Facet params …

jkeck:
…parsing.

Fixes https://app.honeybadger.io/projects/50022/faults/43892552


cbeer:
I don't know if https://github.com/projectblacklight/blacklight/pull/2313 can also be backported at all?

PR #2509, opened 2020-07-30
Augment JSON API documents to include link data (similar to Articles+)

jkeck:
This is to facilitate the Online link described in sul-dlss/sul-bento-app#293

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -526,6 +526,21 @@ class CatalogController < ApplicationController
     flash[:success] = I18n.t('blacklight.email.success')
   end
 
+  def augment_solr_document_json_response(documents)
+    documents.map do |document|
+      next document.as_json unless document&.access_panels&.online?
+
+      links = document&.access_panels&.online&.links&.map do |link|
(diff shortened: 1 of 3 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -526,6 +526,21 @@ def send_emails_to_all_recipients
...
   end
 
+  def augment_solr_document_json_response(documents)
+    documents.map do |document|
+      next document.as_json unless document&.access_panels&.online?
+
+      links = document&.access_panels&.online&.links&.map do |link|
+        html = link.stanford_only? ? "<span class=\"stanford-only\">#{link.html}</span>" : link.html
+
+        "<span class=\"online-label\">Online</span> #{html}"
```
jkeck, line 536: FWIW, I don't think we should be providing this label in the API.  We were already doing this in the Articles+ response and the bento app has a shared partial for both catalog and articles+ results.

This kept us consistent, but if this seems agreeable, I would like to ticket (and happily work on) adding PRs to SearchWorks and Bento to move these labels to the search service for each endpoint respectively (and we would then need to sync up on deploying those changes out).

camillevilla: @jkeck This sounds like a good change! Feel free to ticket 🎟

Issue #2511, opened 2020-07-31
Add formatted composed title field for Article API results


jkeck:
We add formatting to otherwise unformatted composed title data for display to users (sometimes it comes formatted through EDS, sometimes not).

We want to use this data, as consistently formatted as it is in SearchWorks, in other places (without necessarily having to duplicate the formatting logic.

Part of sul-dlss/sul-bento-app#294

PR #2514, opened 2020-08-03
Pass eds_composed_title through the italicize_composed_title helper …

jkeck:
…before sending through the API.

Closes sul-dlss/SearchWorks#2511
Part of  sul-dlss/sul-bento-app#294

### DIFF
--- a/app/views/articles/index.json.jbuilder
+++ b/app/views/articles/index.json.jbuilder
@@ -1,7 +1,9 @@
 docs = @presenter.documents.collect do |document|
   link = ArticleFulltextLinkPresenter.new(document: document, context: self).links.try(:first) # top priority one only
+  composed_title = document['eds_composed_title']
   data = document.to_h # avoids deprecation warning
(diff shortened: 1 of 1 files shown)


app/views/articles/index.json.jbuilder
```diff
@@ -1,7 +1,9 @@
 docs = @presenter.documents.collect do |document|
   link = ArticleFulltextLinkPresenter.new(document: document, context: self).links.try(:first) # top priority one only
+  composed_title = document['eds_composed_title']
   data = document.to_h # avoids deprecation warning
   data['fulltext_link_html'] = link if link.present?
+  data['eds_composed_title'] = italicize_composed_title({ value: Array.wrap(composed_title) }) if composed_title.present?
```
jkeck, line 6: I was going to look into moving this over into the new JsonResults presenter but realized that the code change to get this updated was going to be a lot smaller than refactoring into that (or a similar) class.

Issue #2561, opened 2020-09-30
Determine if SmartText search is enabled and indicate this to the user


jkeck:
### Context
We've enabled SmartText search which is a mode that is only executed under when the search query yields zero results.  This happens automatically and is not apparent to the user.  This can, however, yield millions of results for certain queries (as it appears that one of the changes that is done under the hood is to "OR" the search terms). 

### Tasks
- [ ] Determine if we can tell from the gem/response if SmartText search has been executed (we may need to request that EBSCO add this feature if not already available)
- [ ] Add a note to the search results display when this mode is enabled
- [ ] DON'T call it SmartText (from the users perspective)

cbeer:
Context: Came up through feedback; people pasted citations that got 0 results and EBSCO does a thing. Enabling the EBSCO thing yielded a lot of search results, but it would be better as an opt-in thing. #2574 is a PR to enable that, but needs text + design (and naming?). EBSCO now has a thing to know if you’re in SmartText now, which might be useful.

So.. more design needed to explore these options.

saseestone:
Closing this for now as public services staff determined that they would prefer that we give better instructions to users on the zero results page, and NOT enable EBSCO's thing (i.e.SmartText).

PR #2574, opened 2020-11-03
Add a link/backend lookup on zero results page of articles to SmartText mode

jkeck:
Connected to #2561

## TODO
- [ ] Come up with better texty-text (as to meet the "Do not call it SmartText" mandate)
- [ ] Move logic into SearchQueryModifier for consistency

## Zero Results Page
Screen Shot 2020-11-03 at 1 56 01 PM

## Search Results in SmartText Mode
Screen Shot 2020-11-03 at 1 56 23 PM


saseestone:
Discussions with public services staff revealed that they want to put time into updating/refining the instructions given to users on the zero results page.  EBSCO's SmartText tool (and the potential to get millions of results when a very specific citation is input) isn't a desired solution at this time.

PR #2597, opened 2021-01-16
Attempt to address an a11y issue with the Send to menu

jkeck:


app/views/shared/record/_record_toolbar_tools_subset.html.erb
```diff
@@ -1,44 +1,44 @@
 <% if @document.respond_to?( :to_sms_text ) %>
   <li class="sms" role="presentation">
     <% if @document.eds? %>
-      <%= link_to t('blacklight.tools.sms_html'), sms_articles_path(:id => @document), {:id => 'emailLink', :data => {:ajax_modal => "trigger"}} %>
+      <%= link_to t('blacklight.tools.sms_html'), sms_articles_path(:id => @document), {:id => 'emailLink', role: 'menuitem', :data => {:ajax_modal => "trigger"}} %>
```
jkeck, line 4: It's not clear if this should just go onto the `<li>` (I've seen examples of both).  I'll bring it up with the a11y folks..

mejackreed (approved):

Issue #2600, opened 2021-01-22
Update "Send to" keyboard a11y


jkeck:
There are a few issues surrounding the "Send to" button / menu (and likely other drop down menus) accessibility 

There are details in the links provided below, but some of the main issues that were called out are:
1. When the button is expanded the focus should be on the first element in the menu.
2. The menu should be expandable with a space bar as well as enter (this seems to be the case for me already)
3. Arrow keys should cycle back to the beginning/end of the list when navigating past the first/last element.
    * Note: I believe the spec also includes some language around when pressing alphanum keys and jumping to matching options in the menu, but idk if we need to go that far for a menu with 5 options unless we get it for free
4. Update tab indexes
    * Note: The a11y consultant mentioned implementing a [roving tab index](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex) to capture focus, but I notice that W3C Example element (linked below) just adds `tabindex="-1"` to each `menitem`, which would be a lot simpler.

Example element: https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html
Other useful links:
http://whatsock.com/training/matrices/#menu
https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton

PR #2608, opened 2021-02-11
Update keyboard a11y of several dropdown menus

jkeck:
Coses #2600 

_Keeping this in draft until we can get it reviewed by the a11y folks._

This updates:
* Send to menus on article records, catalog records, and the selections UI
* The Help/SUL Services menu
* The All/Catalog/Articles search selection menu
* The Catalog/Articles/Clear Selections menu


jkeck:
Note: This ends up adding a link to the current search endpoint in the menu where it was not before (so that the contents are perceivable by a screen reader as they're navigating the menu) but it's listed as the current option via `aria-current` and is just an anchor link (that we trap for and close the menu when selected).  Hopefully this has the same intent while improving the screenreader accessibility.

Screen Shot 2021-02-19 at 6 16 10 PM


Another issue that was brought up is that the active state is only conveyed to sighted users via the shade of yellow.  A suggested improvement would be to add some non-color based indicator of the active option (possibly bolding the font weight), but wanted to pass that by the design folks @ggeisler / @ilya-ux.

mejackreed (approved):
I've reviewed this and tested it in stage. 👍 to the attention to detail in moving this important work forward.

PR #2611, opened 2021-02-18
Re-add the restriction for emails to not include links  …

jkeck:
…(but only for non-logged in users)

If we do go this route, we will likely want to update some relevant help text/flash messages.

##  Feedback Form/Help Text
feedback form

## Feedback Flash Message
feedback flash message

## Record Email Form/Help Text
record email form

## Record Email Flash Message
record email feedback

---

# Possible Alternative 
The main reason why we are re-implementing this is reCaptcha has been [solved](https://prowebscraper.com/blog/top-10-captcha-solving-services-compared/) and we're seeing an increasing amount of spam coming through. I looked at updating to V3 (although that also appears to have been solved), but it would present some additional design changes.  This captcha is no-friction (e.g. no action required by the user) and instead monitors behavior to "score" how likely the user is human vs. a bot.  This could allow us to block the emails outright, present a traditional captcha, etc. however; might not actually do us any good if the bots are solving the captcha in other ways.

If we did look into captcha V3, the design changes we'll need to account for is there is a requirement to have the reCaptcha brand, privacy policy link, and ToS link as part of the user flow ([source](https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed)).  They add a little icon on all pages, but we could avoid that by adding something like what is described in the above link on the feedback / email forms themselves.

## On all pages
default captcha badge example

**vs**

## Only in the forms/user flow
text captcha badge example

### DIFF
--- a/app/controllers/concerns/email_validation.rb
+++ b/app/controllers/concerns/email_validation.rb
@@ -5,6 +5,8 @@ module EmailValidation
     case
     when !%w(full brief).include?(params[:type])
       flash[:error] = I18n.t('blacklight.email.errors.type')
+    when anonymous_user_sent_link?
+      flash[:error] = I18n.t('blacklight.email.errors.message.spam')
     when params[:to].blank?
       flash[:error] = I18n.t('blacklight.email.errors.to.blank')
     when too_many_email_addresses?
@@ -14,6 +16,12 @@ module EmailValidation
     end
 
     flash[:error].blank?
+  end
+
+  def anonymous_user_sent_link?
+    return false if current_user.present?
+
+    params[:message] =~ %r{href|url=|https?://}i
   end
 
   def valid_email_addresses?
--- a/app/controllers/feedback_forms_controller.rb
+++ b/app/controllers/feedback_forms_controller.rb
@@ -40,12 +40,26 @@ class FeedbackFormsController < ApplicationController
   def validate
     errors = []
 
-    errors << 'You must pass the reCAPTCHA challenge' if current_user.blank? && !verify_recaptcha
+    if current_user.blank?
+      errors << 'You must pass the reCAPTCHA challenge' unless verify_recaptcha
+
+      if params[:message] =~ url_regex
+        errors << "Your message appears to be spam, and has not been sent. Please try sending your message again without any links in the comments."
+      end
+
+      if params[:user_agent] =~ url_regex || params[:viewport] =~ url_regex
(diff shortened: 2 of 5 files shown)


ggeisler:
@jkeck @ilya-ux Below are my suggestions. Hopefully I've remembered our discussion well enough that these make sense but let me know if you have questions or suggestions for alternative ways to handle this.

I assume we are going with the captcha V3, using the alternative approach that enables us to hide the reCaptcha badge. The updates below are pretty similar for the two cases (feedback form and item email form) but I'll list them separately.


## Item email form

Screen Shot 2021-02-26 at 4 25 02 PM

---

- [ ] A: Add the `.help-block` text under the Message field: `Do not include links. The email will automatically include a link to the record.`
- [ ] B: Replace the current reCaptcha box with Google's [required text for the alternative](https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed): `This site is protected by reCAPTCHA and the Google Privacy Policy and Terms of Service apply.` with the "Privacy Policy" and "Terms of Service" phrases links as they suggest.
- [ ] B: Left-align it in the modal.
- [ ] B: Style this as a `.help-block` if possible to get the gray text, or otherwise style it so it matches the `.help-block` text styling. (De-emphasizes it a bit.)

## Item email form flash message

Screen Shot 2021-02-26 at 4 25 51 PM


---

- [ ] This is pretty similar to the current flash message text but I edited it a bit:
```text
Your message appears to be spam and has not been sent. Remove any links in the message and try sending it again.
```

## Feedback form 
Screen Shot 2021-02-26 at 4 26 17 PM

---

- [ ] A: Add the `.help-block` text under the Message field: `Do not include links. Your feedback will automatically include a link to this page.`
- [ ] B: Replace the current reCaptcha box with Google's [required text for the alternative](https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed)
- [ ] B: Put the reCaptcha text within a `.col-sm-offset-3` so it gets the same horizontal alignment as the "Reporting from" text at the top of the form.
- [ ] B: Style this as a `.help-block` if possible to get the gray text, or otherwise style it so it matches the `.help-block` text styling. 


## Feedback form flash message

Screen Shot 2021-02-26 at 4 00 16 PM

---

This is the same flash message as for the item email modal:

```text
Your message appears to be spam and has not been sent. Remove any links in the message and try sending it again.
```

PR #2612, opened 2021-02-18
Use #mods? instead of #mods to avoid parsing the mods record if we don't have to

cbeer:


app/views/catalog/_summary_data.html.erb
```diff
@@ -52,7 +52,7 @@
     <%= document[:summary_display].join('<br/>').html_safe %>
   </div>
 <%# TODO: Remove this elsif block after new mods index is in place %>
-<% elsif document.respond_to?(:mods) && document.mods.present? %>
+<% elsif document.respond_to?(:mods?) && document.mods? && document.mods.present? %>
```
mjgiarlo, line 55: Elsewhere in this PR `document.mods?` replaces `document.mods.present?` but here they're both used. Intentional or cruft?

cbeer: 👍 ; i think this check is unnecessary anyway... if there's mods, we'll be parsing it anyway 🤷

mjgiarlo (approved):

camillevilla (approved):

PR #2614, opened 2021-02-19
Install blacklight dynamic sitemap

jkeck:
🤷  we don't have to enable it now, but this could allow us to start looking into sitemaps


jkeck:
Took this out of draft.  Had put it in briefly to see if we could/should do something dynamically to the robots.txt, but it's probably not worth it, so we'll need to update the robots.txt when we're ready for crawlers to start consuming the sitemap.

PR #2619, opened 2021-02-23
Replace current RecordHelpers with upstream ported code

mejackreed:
Screen Shot 2021-02-22 at 5 00 24 PM

I've only pointed this at the local index, nothing with production data. But looking good to me.

### DIFF
--- a/app/views/catalog/_mods_search_results_document_fields.html.erb
+++ b/app/views/catalog/_mods_search_results_document_fields.html.erb
@@ -6,13 +6,13 @@
     <% if document.mods_display_name.present? %>
       <li>
         <% name = document.mods_display_name.first %>
-        <%= link_to_mods_name(name['link']) %> <%= name['post_text'] %>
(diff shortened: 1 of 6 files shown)


app/views/catalog/_mods_search_results_document_fields.html.erb
```diff
@@ -6,13 +6,13 @@
     <% if document.mods_display_name.present? %>
       <li>
         <% name = document.mods_display_name.first %>
-        <%= link_to_mods_name(name['link']) %> <%= name['post_text'] %>
+        <%= link_to(name['link'], search_catalog_path(q: "\"#{name['link']}\"", search_field: 'search_author')) %> <%= name['post_text'] %>
```
mejackreed, line 9: This could be ported back out to a helper, or we could not delete the one we had previously? But.. 🤷

jkeck (approved):
LGTM 👍

jkeck:
I've approved, but also want to give the oppo for anybody else to take a look (and also possibly take this for a spin in stage if anybody wants to poke around.

mejackreed:
@jkeck thanks for deploying that up. I did notice an issue and pushed another commit to resolve the extra `<dt>`

PR #2641, opened 2021-03-31
Make 050 call numbers browseable

cbeer:


app/views/catalog/record/_record_metadata_default.html.erb
```diff
@@ -13,6 +13,6 @@
   </div>
 </div>
 
-<% if document.holdings.present? && document.holdings.browsable_callnumbers.present? %>
+<% if document.holdings.browsable_callnumbers.present? %>
```
mjgiarlo, line 16: @cbeer is `document.holdings` always `present?` then? Not sure if this nil/blank check is still useful or not.

mjgiarlo (approved):

app/views/catalog/record/_record_metadata_default.html.erb
cbeer: Yeah, and unfortunately has some backstory, context, and legacy. `#holdings` is always defined:

https://github.com/sul-dlss/SearchWorks/blob/1e03ee62765fb3cd06d7235ccc2c4f878a15db05/app/models/concerns/solr_holdings.rb

But `Holdings#present?` diverges  from the expected implementation:

https://github.com/sul-dlss/SearchWorks/blob/master/lib/holdings.rb#L35-L39

But it's implementation is more-or-less right for all the other places we display holdings so....

mjgiarlo: Wheeeee! OK.

mjgiarlo (approved):

PR #2647, opened 2021-04-13
Extract MarcExtractor to extract marc

cbeer:


app/models/marc_extractor.rb
```diff
@@ -0,0 +1,91 @@
...
+  attr_reader :marc, :tags, :subfields
+
+  def initialize(marc, tags)
+    @marc = marc
+    @tags = []
+    @subfields = {}
+    tags = tags.split(':') if tags.is_a? String
+
+    Array.wrap(tags).each do |tag|
+      t, subfields_spec = tag.to_s.scan(/\d{3}|\w+/)
```
camillevilla, line 11: 👀 this regex magic and mushing fields and subfields together (e.g. `%w(382abdnpsv)`) feels a little weird to me but... I guess we're only using it in 3 places.

camillevilla (approved):
Thanks for doing this refactor, @cbeer. My comment on the regex stuff is non-blocking. Just something to consider for next time. I do think it's an improvement to have the information about the field and subfields closer together in the model, but I think the regex magic might be a little harder to grok for someone new to this codebase.

app/models/marc_extractor.rb
cbeer: yeah; it's kinda funny, but matches up with what we're doing over in traject:

https://github.com/traject/traject/blob/master/lib/traject/marc_extractor.rb#L11-L13

If we ever want to add indicator support, maybe we'd just pull the traject parser in entirely 🤷‍♂️

PR #2659, opened 2021-05-10
Tweak MARC display to match production

cbeer:


app/models/concerns/marc_bound_with_note.rb
```diff
@@ -3,12 +3,18 @@
 ###
 module MarcBoundWithNote
   def bound_with?
-    bound_with_note.present?
+    bound_with_note_for_access_panel.present?
```
mjgiarlo, line 6: The effect here is to change how we're determining whether something is a bound_with or not. OK.

app/models/concerns/marc_bound_with_note.rb
```diff
@@ -3,12 +3,18 @@
 ###
```
mjgiarlo, line 3: Comment typo on prior line if you're inclined to toss that in here.

app/models/marc_extractor.rb
```diff
@@ -18,7 +18,8 @@ def extract
     return to_enum(:extract) unless block_given?
 
     merged_and_matched_fields.each do |field|
-      yield field, relevant_subfields(field).to_a
+      subfields = relevant_subfields(field).to_a
+      yield field, subfields if subfields.present?
```
mjgiarlo, line 22: So, do not yield unless the relevant subfields are there. OK.

app/models/marc_fields/awards.rb
```diff
@@ -0,0 +1,14 @@
+##
+# A class to handle MARC 586 field logic
+# https://www.loc.gov/marc/bibliographic/bd586.html
+class Awards < MarcField
+  private
+
+  def display_value(field, subfields)
+    safe_join [super, Constants::SOURCES[field['1']]].compact, subfield_delimeter
```
mjgiarlo, line 8: Looks like this method has a typo in it and it appears in four places. Not a biggie, and this PR didn't add the method, but thought I'd point out `s/delimeter/delimiter/` :man_shrugging:

app/models/marc_field.rb
```diff
@@ -70,6 +70,8 @@ def subfield_delimeter
   end
 
   def display_value(field, subfields)
+    return unless subfields.present?
```
mjgiarlo, line 73: More guarding against nil/empty subfields.

app/models/marc_fields/bound_with_note.rb
```diff
@@ -2,22 +2,17 @@
...
-      { id: id, value: display_value(field, subfields.reject { |subfield| subfield.code == 'c' }) }
-    end
-  end
+  private
 
-  def to_partial_path
-    'marc_fields/bound_with_note'
-  end
+  def subfield_value(field, subfield)
+    return super unless subfield.code == 'c'
```
mjgiarlo, line 10: Specialize behavior only when the subfield code is `c`. Consistent with class comments above, so that's good.

app/models/marc_fields/bound_with_note.rb
```diff
@@ -2,22 +2,17 @@
...
 
-  private
+    ckey = subfield.value[/^(\d+)/]
+    ckey_link = link_to(ckey, Rails.application.routes.url_helpers.solr_document_path(ckey))
```
mjgiarlo, line 13: Ah, yes, thanks for giving this a name.

app/models/marc_fields/bound_with_note.rb
```diff
@@ -2,22 +2,17 @@
...
 
-  def extracted_fields
-    super.select { |field, _subfields| field['c'].present? }
+    subfield.value.gsub(ckey, ckey_link).html_safe
```
mjgiarlo, line 15: Makes sure what is returned is a clickable link?

app/models/marc_fields/bound_with_note_for_access_panel.rb
```diff
@@ -0,0 +1,26 @@
+###
+#  MarcBoundWith class to return 590 fields that have a $c
```
mjgiarlo, line 2: Same class comment as for `BoundWithNote`. Might be nice to have a diff comment here for easier grokking of how the classes differ.

app/models/marc_fields/bound_with_note_for_access_panel.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+  def extracted_fields
+    super.select { |field, _subfields| field['c'].present? }
+  end
+
+  def display_value(field, subfields)
+    safe_join subfields.reject { |subfield| subfield.code == 'c' }.map { |subfield| subfield_value(field, subfield) }, subfield_delimeter
```
mjgiarlo, line 24: OK, this one is twisting my wig a bit. First we're rejecting all the subfields with a code of `c`, and then we're using `subfield_value` to return values... but that means `subfield_value`'s special behavior, as defined in this class's parent (`BoundWithNote`), as isn't getting triggered so this is ultimately going up the chain to `BoundWithNote`'s parent. Is that what's intended? I feel like this nuance might make a good addition to the class-level documentation for this new class.

app/models/marc_fields/bound_with_note_for_access_panel.rb
```diff
@@ -0,0 +1,26 @@
+###
+#  MarcBoundWith class to return 590 fields that have a $c
+###
+class BoundWithNoteForAccessPanel < BoundWithNote
+  def values
+    extracted_fields.map do |field, subfields|
+      id = subfields.find { |subfield| subfield.code == 'c' }.value[/^(\d+)/]
```
mjgiarlo, line 7: And `id` here is a catkey like above? If so, would you consider renaming the variable to `ckey` (also, like above)?

app/models/marc_fields/issn.rb
```diff
@@ -1,15 +1,19 @@
...
+# A class to handle MARC 022a + 022z field logic
+# https://www.loc.gov/marc/bibliographic/bd022.html
 class Issn < MarcField
   def values
-    super.map(&:strip).select { |v| v =~ issn_pattern }
+    return [] if marc.blank?
+
+    @values ||= extracted_fields.flat_map do |field, subfields|
+      subfields.map(&:value).map(&:strip).select { |v| v =~ issn_pattern }
+    end.compact
```
mjgiarlo, line 10: Can you say more about how the ISSN changes relate to the bound_with changes? Or maybe they don't.

app/views/catalog/record/_marc_contents_summary.html.erb
```diff
@@ -42,8 +42,8 @@
   </dd>
 <% end %>
 
-<% summary = document.fetch(:summary_struct, []).first %>
-<% if summary.present? %>
+<% summary = document.fetch(:summary_struct, []).first || {} %>
+<% if summary[:fields].present? || summary[:unmatched_vernacular].present? %>
```
mjgiarlo, line 46: Can you say more about this change?

app/views/catalog/record/_marc_contents_summary.html.erb
```diff
@@ -42,8 +42,8 @@
   </dd>
 <% end %>
 
-<% summary = document.fetch(:summary_struct, []).first %>
-<% if summary.present? %>
+<% summary = document.fetch(:summary_struct, []).first || {} %>
```
mjgiarlo, line 45: I assume `summary_struct` is an enumerable that returns hash-like objects?

app/views/marc_fields/_awards.html.erb
```diff
@@ -0,0 +1,6 @@
+<% if marc_field.values.present? %>
+  <dt><%= marc_field.label %></dt>
+  <% marc_field.values.each do |value| %>
+    <dd><%= auto_link(value) %></dd>
+  <% end  %>
+<% end %>
```
mjgiarlo, line 6: :ok_hand:

config/locales/searchworks.en.yml
```diff
@@ -60,11 +60,11 @@ en:
         label: all
         description_html: <span class="h3">all</span> <span class="help-block">catalog, articles, website, &amp; more in one search</span>
     marc_fields:
-      020:
+      "020":
```
mjgiarlo, line 63: Quotes added I assume because of the leading zeros?

spec/fixtures/marc_records/marc_metadata_fixtures.rb
```diff
@@ -1310,4 +1310,17 @@ def managed_purl_fixture
       </record>
     xml
   end
+
```
mjgiarlo, line 1313: Code coverage of the changes above looks :trophy:

app/models/marc_fields/bound_with_note_for_access_panel.rb
cbeer: Asking the hard questions... I don't know why this is the display behavior in the access panel or why it's different than the normal place. This code replaced

https://github.com/sul-dlss/SearchWorks/pull/2647/files#diff-c6b37349f78dc1bef9f5073e68179f4fcd4e51388aa19d7051ce4637fbc7037fL5

🤷‍♂️

app/models/concerns/marc_bound_with_note.rb
mjgiarlo: @cbeer Makes sense.

app/models/marc_fields/bound_with_note_for_access_panel.rb
mjgiarlo: Yes, good point: changing the hash here would break stuff that depends on it. NBD.

app/views/catalog/record/_marc_contents_summary.html.erb
cbeer: Yeah -- long story short, we push some of the MARC parsing into solr so we don't need to parse MARC records from the `catalog#index` page.

app/models/marc_fields/bound_with_note_for_access_panel.rb
mjgiarlo: @cbeer I suspect it's OK. Looking at the code it replaced, it looks like the intent may have been to extract the catkey (from the 'c' subfield) but then ignore that subfield when displaying the value since we're apparently special-case displaying the catkey as the ID.

config/locales/searchworks.en.yml
mjgiarlo: Thanks, :computer: !

app/models/marc_fields/bound_with_note_for_access_panel.rb
cbeer: Sounds right; but I think the why and wherefore have been lost to history.

mjgiarlo (approved):

PR #2661, opened 2021-05-12
Ditch a bunch of deprecation warnings

mjgiarlo:
Sadly, not all of them. Many BL7-related deprecation warnings remain.

### DIFF
--- a/app/models/hathi_trust_links.rb
+++ b/app/models/hathi_trust_links.rb
@@ -51,7 +51,7 @@ class HathiTrustLinks
   def item_url
     return item_target(hathitrust_item_id) if publicly_available?
 
-    "https://babel.hathitrust.org/Shibboleth.sso/Login?entityID=#{stanford_shib_id}&target=#{URI.encode(item_target(hathitrust_item_id))}"
+    "https://babel.hathitrust.org/Shibboleth.sso/Login?entityID=#{stanford_shib_id}&target=#{item_target(hathitrust_item_id)}"
   end
 
   def stanford_shib_id
@@ -67,6 +67,6 @@ class HathiTrustLinks
   end
 
   def item_target(id)
-    "https://babel.hathitrust.org/cgi/pt?id=#{URI.encode(id)}&view=1up&seq=9"
+    "https://babel.hathitrust.org/cgi/pt?id=#{ERB::Util.url_encode(id)}"
   end
 end
--- a/spec/models/hathi_trust_links_spec.rb
+++ b/spec/models/hathi_trust_links_spec.rb
@@ -3,9 +3,10 @@
 require 'spec_helper'
 
 describe HathiTrustLinks do
+  subject(:ht_links) { described_class.new(document) }
+
   let(:document_data) { { ht_bib_key_ssim: ['abc123'], ht_htid_ssim: ['1234567'] } }
   let(:document) { SolrDocument.new(document_data) }
-  subject(:ht_links) { described_class.new(document) }
 
   describe '#present?' do
     context 'when there is hathi data' do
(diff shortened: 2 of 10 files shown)


app/models/hathi_trust_links.rb
```diff
@@ -51,7 +51,7 @@ def work_url
   def item_url
     return item_target(hathitrust_item_id) if publicly_available?
 
-    "https://babel.hathitrust.org/Shibboleth.sso/Login?entityID=#{stanford_shib_id}&target=#{URI.encode(item_target(hathitrust_item_id))}"
+    "https://babel.hathitrust.org/Shibboleth.sso/Login?entityID=#{stanford_shib_id}&target=#{item_target(hathitrust_item_id)}"
```
mjgiarlo, line 54: No need to encode here because encoded below. I double-checked to make sure this output was the same as before.

mjgiarlo: I'll give those a try tomorrow. I picked one or two real HT IDs, not knowing how varied they are.

cbeer: Yeah, they're all over the place... probably created by the contributing institution. I think it's a little less wild-west than I thought, but probably worth being extra-cautious about regardless. 

```
| aub.i10592830             |
| loc.ark:/13960/t17m0qv0b  |
| bc.ark:/13960/t5s790w6p   |
| mdl.reflections.ach00001  |
| umd.31430058079171        |
| msu.31293024597860        |
```

mjgiarlo: @cbeer thanks for that list! I'll add some test coverage and toss up a delta soon.

mjgiarlo: @cbeer These IDs are now encoded using `ERB::Util.url_encode`. I tested some real SW URLs pointing at HathiTrust with the sort of escaping that comes out of this method, and they are routed properly when bouncing through Shib. I changed the tests to make sure we cover this.

app/models/hathi_trust_links.rb
```diff
@@ -67,6 +67,6 @@ def fulltext_available?
   end
 
   def item_target(id)
-    "https://babel.hathitrust.org/cgi/pt?id=#{URI.encode(id)}&view=1up&seq=9"
+    "https://babel.hathitrust.org/cgi/pt?id=#{ERB::Util.url_encode(id)}"
```
mjgiarlo, line 70: We decided to dump these extra query args because 1) they are not needed, and 2) in practice, they are not properly encoded so they do not survive the bounce through Shib.

spec/models/hathi_trust_links_spec.rb
```diff
@@ -67,10 +68,10 @@
       end
 
       context 'when not publicly_available' do
-        before { document_data[:ht_access_sim] = ['access'] }
+        before { document_data[:ht_access_sim] = ['deny'] }
 
-        it 'does not include the signon param (and sends the user directly to the work)' do
-          expect(ht_links.url).not_to include('?signon=swle:urn:mace')
+        it 'includes the signon param' do
+          expect(ht_links.url).to include('hathitrust.org/Record/abc123?signon=swle:urn:mace:incommon:stanford.edu')
```
mjgiarlo, line 74: Yeah, this test was not set up right so it was misleading.

spec/models/hathi_trust_links_spec.rb
```diff
@@ -96,21 +97,23 @@
...
-          expect(ht_links.url).to include('target=https://babel.hathitrust.org/cgi/pt?id=1234567&view=1up')
+          expect(ht_links.url).to include('target=https://babel.hathitrust.org/cgi/pt?id=1234567')
         end
       end
 
       context 'when not publicly_available' do
-        before { document_data[:ht_access_sim] = ['access'] }
+        before do
+          document_data[:ht_access_sim] = ['deny']
+          document_data[:ht_htid_ssim] = ['loc.ark:/13960/t17m0qv0b']
```
mjgiarlo, line 112: Test URLs with special characters and set it up properly for testing not public URLs (was not done correctly before).

spec/models/hathi_trust_links_spec.rb
```diff
@@ -96,21 +97,23 @@
...
+        end
 
-        it 'is the item URL directly (not passed as a the targer param)' do
-          expect(ht_links.url).not_to include('target=')
-          expect(ht_links.url).to start_with('https://babel.hathitrust.org/cgi/pt?id=1234567&view=1up')
+        it 'bounces the user to shibboleth with the encoded item URL in the target param' do
```
mjgiarlo, line 115: Corrected the description too.

PR #2671, opened 2021-05-26
Guard against the 740 field not existing

### LINKED ISSUE #2670: [SearchWorks/prod] ActionView::Template::Error: undefined method `indicator2' for nil:NilClass
## Backtrace

    line 8 of [PROJECT_ROOT]/app/models/marc_fields/added_entry.rb: label
    line 2 of [PROJECT_ROOT]/app/views/marc_fields/_marc_field.html.erb: _app_views_marc_fields__marc_field_html_erb___3074005374992994173_119080
    line 3 of [PROJECT_ROOT]/app/helpers/marc_helper.rb: render_if_present

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/79450140)

cbeer:
This might happen if the data is coming from an unmatched vernacular field.

Fixes #2670

### DIFF
--- a/app/models/marc_fields/added_entry.rb
+++ b/app/models/marc_fields/added_entry.rb
@@ -5,7 +5,7 @@ class AddedEntry < MarcField
   def label
     f = marc[tags.first]
 
-    if f.indicator2 == '2'
+    if f&.indicator2 == '2'
       'Included Work'
     else
       super


cbeer:
I have some other concerns about this display (740 is a repeatable field; is it possible a record has a mix of 740s with and without an indicator2 value? or unmatched vernacular that should display as included works?). I'm not sure it's worth solving this before at least displaying the record at all, though.

camillevilla (approved):

Issue #2679, opened 2021-06-03
screenreader accessibility on callnumber browse 


camillevilla:
## quick fixes:
- [x] Add meaningful labels to the arrow buttons ("Previous call numbers" and "Subsequent callnumbers"??)
- [x] Hide the bookcover text (or possibly the entire link?) from screenreaders to make this page less noisy (see the duplicated title links in the screenshot below)

Sample pages:
- [Catalog record with callnumber browse](https://searchworks.stanford.edu/view/10096898)
- [full page callnumber browse](https://searchworks.stanford.edu/browse?start=10096898&view=gallery)

![Voiceover links menu](https://user-images.githubusercontent.com/5402927/120566766-d3d6ad00-c3c4-11eb-86f7-fb38bf3d8bf8.png)

![Screen Shot 2021-06-02 at 5 06 27 PM](https://user-images.githubusercontent.com/5402927/120566880-1f895680-c3c5-11eb-8343-fdb67b6bd4a9.png)

---
## Medium term:
- Can we change the tab focus jumps to the record opened when a user toggles the "Preview" button? 
- Can we use a more meaningful label than "Select"?
---
## Long term: 
Some more holistic review from SODA office hours and redesign of this page would be needed to make it more accessible.
![Screen Shot 2021-06-02 at 5 01 29 PM](https://user-images.githubusercontent.com/5402927/120567177-f5846400-c3c5-11eb-912f-a9be00a79ca1.png)

camillevilla:
I'm good with closing this since we resolved the quick fixes. The other stuff has design implications, so I'll defer until we have more bandwidth for that and can triage the SiteImprove reports.

PR #2680, opened 2021-06-03
Improve callnumber browse accessibility

cbeer:
Before:
Screen Shot 2021-06-03 at 07 15 39

After:
Screen Shot 2021-06-03 at 07 15 48

Part of #2679


camillevilla (approved):
Definite improvement!

PR #2711, opened 2021-07-29
Display abstract + toc labels using mods_display

### LINKED ISSUE #2710: Table of contents label not displaying on PURL or SW record
The MODS for this item has a value for element tableOfContents but there is no label displaying, so the value text looks like it's part of the Abstract/Summary.

The default label should be "Table of contents" unless someone adds an attribute to the element for a custom displayLabel.

Example item: https://purl.stanford.edu/mn521kw5258

As a temporary workaround, I typed in "Table of contents:" as part of the data, so it is clear to the user what all that text means. But it should display separately from the Abstract/Summary and have its own header label.

cbeer:
fixes #2710

Screen Shot 2021-07-29 at 07 55 50

### DIFF
--- a/app/views/catalog/record/_mods_abstract_contents.html.erb
+++ b/app/views/catalog/record/_mods_abstract_contents.html.erb
@@ -1,16 +1,18 @@
 <% document ||= @document %>
 <% document.mods.abstract.each do |abstracts| %>
   <% if abstracts.respond_to?(:values) %>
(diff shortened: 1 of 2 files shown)


camillevilla (approved):

cebraj:
Thanks all for working on this! The above screenshot looks very cool, if that is the way it will look in production, I support this :)

PR #2746, opened 2021-09-15
Refactoring Holdings 

cbeer:


mjgiarlo (approved):
Like moving stuff out of constants.rb and into config. It'd be good for @camillevilla to look too.

camillevilla (approved):

PR #2748, opened 2021-09-17
Allow reserve desks to have their own location translations

### LINKED ISSUE #2373: Law requests new location translation for LAW-RESV
Law has requested that we translate the current Location for LAW-RESV to "On reserve: Teaming room"

It appears that we're not using the translations that are listed in the constants list for Locs.  I can't tell how we're deciding to translate -RESV locations, but all seem to display "On reserve: Ask at circulation desk"

Can we have a different location translation for LAW-RESV?

cbeer:
Fixes #2373


lib/constants.rb
```diff
@@ -931,7 +929,6 @@ module Constants
     'HOOV-RESV' => 'HOOVER',
     'HOP-RESV' => 'HOPKINS',
     'LAW-RESV' => 'LAW',
-    'MATH-RESV' => 'MATH-CS',
```
camillevilla, line 933: 👏

camillevilla (approved):

PR #2765, opened 2021-11-17
articles+ auth fixes/cleanup

### LINKED ISSUE #2445: User reports intermittent issue logging in
Just as an FYI on this issue, we've had several people try to reproduce this with most folks being unable to (@trapido has been able to, but that's about it).  Filing the ticket for more visibility and in case there is any pattern that this fits into that could help us solve it but it might not necessarily be actionable at this point (FWIW we haven't received any reports from any other users)

From SW-3893

> I use SearchWorks articles+ routinely. Most times I click on “Login” at upper right, which would take me to a screen for entering my sunet ID and password, which I would do and that would take me back to SearchWorks articles+ with “Logout” now appearing at upper right. I would click on “Logout” to leave the session and close my browser. I’m running Firefox under Windows 10.

> Twice this past week, however, I’ve clicked on “Login” at upper right, which would take me to a screen for entering my sunet ID and password, which I would do and that would take me back to SearchWorks articles+ where “Login” still appears, not “Logout”; and some materials remain unavailable because I remain in guest mode. Why is this login process failing? I’m able to use my sunet ID and password to access other resources, such as webmail. Please advise.

It might also help if added some sort of notification (Honeybadger or other) if we were able to capture when a user returns from the IDP but isn't logged in as we would expect them to be.

thatbudakguy:
fixes #2445 

- Use warden after_authentication callback to store Shib data (#2445)
- Clear EDS session token in login/logout callbacks
- Remove dead EDS params code from ArticleSelectionsController


config/initializers/devise.rb
```diff
@@ -13,6 +13,27 @@
...
+  Warden::Manager.after_authentication do |user, auth, opts|
+    auth.session(:user)['suAffiliation'] = auth.env['suAffiliation']
+    user.affiliations = auth.env['suAffiliation']
+    # Reset EDS session token so that a new session is established
+    auth.env['rack.session']['eds_session_token'] = nil
+  end
+
+  Warden::Manager.after_fetch do |user, auth, opts|
+    user.affiliations = auth.session(:user)['suAffiliation']
+  end
```
thatbudakguy, line 29: why do we set `user.affiliations` in both these callbacks? warden docs don't explain what `after_fetch` does; wondering when & in what context it is called

spec/controllers/login_controller_spec.rb
```diff
@@ -12,45 +12,5 @@
...
-      end
-
-      it 'is set to the ENV attribute when present' do
-        expect(ENV).to receive(:[]).with('suAffiliation').and_return('stanford:staff')
-
-        get :login
-
-        expect(session['suAffiliation']).to eq 'stanford:staff'
-      end
-    end
```
thatbudakguy, line 54: 👍🏻 these tests seemed like they weren't actually testing behavior and knew too much about internals, so I'm down to just delete them

corylown (approved):
Looks good me and seems to behave as expected as deployed to https://searchworks-uat.stanford.edu/. 👍 refactoring to move `affiliations` to the `User` model.

Issue #2771, opened 2021-12-02
Articles+ full text pdf downloads, authentication and an "S" icon


saseestone:
This might need to be split into multiple tickets but we continue to get reports from users that are logged into SearchWorks, using Articles+ and attempt to download a full text pdf.  

1. They see the message "Sorry, the PDF download was not successful because you are currently in guest mode. Log in to try the download again." even though they are logged in.  
2. We don't indicate to the user that authentication is required for these resources. Eric told Chris in August 2021 that full text pdf downloads will need authentication. Chris wanted to do some more due diligence to make sure that's true (e.g. open access databases)
3. Whatever we can determine needs auth should probably have the "S" added 

Example records: 
https://searchworks.stanford.edu/articles/edsbro__edsbro.A236553399
https://searchworks.stanford.edu/articles/edsbl__vdc.100092542464.0x000001

Jira ticket: [SW-4029 ](https://jirasul.stanford.edu/jira/browse/SW-4029)

camillevilla:
![Screen Shot 2022-01-18 at 4 12 37 PM](https://user-images.githubusercontent.com/5402927/150039221-e649e1e8-e521-469a-ac36-16d1a7215957.png)
The "S" icon is now displayed in PR #2796, @corylown asks: should we also be displaying the "S" in the index view? I did a little research into how we're using the icon so that @saseestone (or @ggeisler in her absence?) could make the call.

## SearchWorks
In SearchWorks, we currently use the "S" icon in a limited number of places. They're all links out to services on other Stanford websites that are only available to Stanford affiliates.
![Screen Shot 2022-01-18 at 4 04 57 PM](https://user-images.githubusercontent.com/5402927/150041392-83ce7ac5-5590-4405-8ee3-5ec01597479b.png)
![Screen Shot 2022-01-18 at 3 56 30 PM](https://user-images.githubusercontent.com/5402927/150041396-b2ba5d28-276e-4b98-9fb4-2a670e5bf3b2.png)

Meanwhile, the majority of things in Articles+ hand off to auth. Here's a useful example search:
https://searchworks.stanford.edu/articles?q=Uptake+of+HPV+vaccine&search_field=search
- View / download PDF 
- View on content provider's site

"Find full text or request" SFX links don't require auth, but they may or may not redirect to sources that require auth. 


## EarthWorks
In Earthworks we have the "S" icon sprinkled all over to distinguish Stanford-contributed data from other institutions' data. 
[...]

PR #2796, opened 2022-01-14
EDS PDF download links display with "S" icon (Stanford Only).

corylown:
Set category 2 links (PDF and PDF E-Book) to Stanford only and add "S" icon in show view when displaying these links.

QUESTION: Not sure whether we should be showing the "S" icon on the index view as well?

Partially addresses #2771


camillevilla (approved):
I think this is good to ship for show views. Reading the original tickets, I'm not sure whether the intention is for the "S" icon to show up in index views. It would add ... a lot of S icons to SearchWorks and we've historically used it in very few places in the application.  I'll continue that discussion in the original ticket #2771.

PR #2855, opened 2022-06-30
Refactor summary_struct display to account for special handling of content advice from MARC 520|4*

### LINKED ISSUE #2839: New display label for MARC 520 with 1st indicator 4
The anti-racist description group has determined that they will place content warnings in the MARC 520 field with a first indicator 4.  Records with a 520 4 should display the label "Content advice". 

Records with a 520 and any other indicators should continue to display under the label "Summary"

Example record: https://searchworks.stanford.edu/view/12725921
In this example, the 520 2 should display under "Summary" and the 520 4 should display under "Content advice".

Also note that if a record has a 520 4 AND a 920, both fields should display.  We're currently suppressing the 520 if a 920 is present, and that will need to change.
Example record: https://searchworks.stanford.edu/view/11474436

corylown:
Closes #2839

### DIFF
--- a/app/views/catalog/record/_marc_contents_summary.html.erb
+++ b/app/views/catalog/record/_marc_contents_summary.html.erb
@@ -44,7 +44,7 @@
 
 <% document.fetch(:summary_struct, []).each do |summary|%>
   <% summary.fetch(:fields, []).each do |field| %>
-    <dt><%= summary[:label] %></dt>
+    <dt><%= summary[:label] || 'Summary' %></dt>
     <dd>
       <%= render_struct_field_data document, field[:field] %>
       <%= "<br/>#{field[:vernacular]}".html_safe unless field[:vernacular].nil? %>


app/views/catalog/record/_marc_contents_summary.html.erb
```diff
@@ -42,17 +42,14 @@
...
-    <% summary[:fields].each do |field| %>
-      <dd>
-        <%= render_struct_field_data document, field[:field] %>
-        <%= "<br/>#{field[:vernacular]}".html_safe unless field[:vernacular].nil? %>
-        <%= "<br/>".html_safe if field != summary[:fields].last %>
-      </dd>
-    <% end %>
+<% document.fetch(:summary_struct, []).each do |summary|%>
+  <% summary.fetch(:fields, []).each do |field| %>
+    <dt><%= summary[:label] %></dt>
```
cbeer, line 47:
```suggestion
    <dt><%= summary[:label] || 'Summary' %></dt>
```

Is there any sense falling back on a default label, or do we know all summaries have labels already?

corylown: They do always have a label set in searchworks_traject_indexer

PR #2898, opened 2022-07-13
Update mods_display to v1.2 so relatedItem will display

### LINKED ISSUE #2860: MODS doesn't display relatedItem in SW & PURL
Why does the related item note display in  https://purl.stanford.edu/gw191zv5762 but not  https://searchworks.stanford.edu/view/gw191zv5762?

corylown:
Fixes #2860

Screen Shot 2022-07-13 at 2 08 31 PM

Branch deployed to uat: https://sw-webapp-sandbox-g.stanford.edu/view/gw191zv5762

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -346,7 +346,7 @@ GEM
       iso-639
       nokogiri (>= 1.6.6)
       nom-xml (~> 1.0)
-    mods_display (1.2.0)
+    mods_display (1.2.1)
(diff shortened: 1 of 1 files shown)


cbeer:
I suggest we hold off on this until we're satisfied with our Blacklight 7/Rails 7 whackamole, and maybe fix https://github.com/sul-dlss/purl/issues/597?

PR #2932, opened 2022-07-20
Rename /webauth/ routes to /sso/

thatbudakguy:
- Remove webauth-specific Devise config
- Rename webauth login route to sso login


thatbudakguy:
blocked on https://github.com/sul-dlss/puppet/pull/8416

Issue #2885, opened 2022-07-11
Do not show "digital content" header for released SDR items that have no published content


andrewjbtw:
This came out of discussion on [PURL-1030](https://jirasul.stanford.edu/jira/projects/PURL/issues/PURL-1030?filter=allopenissues): there are some use cases where SDR depositors want to release only metadata to SearchWorks and do not want thumbnails shown or files listed on the page. Cases include deposits where content is available only by request and the depositor would like the metadata discoverable so that people can find out that the content exists and can be requested.

Currently, releasing an SDR item with no public content results in a display of the "Digital content" header followed by an empty section of the page. This happens for both MARC-origin records (which can display metadata with no content already, via MARC editing) and SDR-origin records, which can only get into SW via release.

Example in stage: https://searchworks-preview-stage.stanford.edu/view/bb661ck4801

corylown:
There's currently nothing in the Solr index available to SearchWorks that tells us whether an SDR item has published content or not. But here are a few options we might consider:

1. Use JavaScript to hide the "Digital content" header if no content is loaded for the embed (look for absence of `iframe` element within `div.purl-embed-viewer` )
2. Add something to the Solr index that indicates whether an SDR item has published content and use the value of this new field to toggle the embed and header. It's not entirely clear to me how we would know this at index time. But maybe we can look for `resource` elements listed within the `contentMetadata` element in `publicObject` -- for example: https://purl.stanford.edu/bt817gh6163.xml
3. Attempt to pre-fetch the embed (`PURLEmbed.new('bb661ck4801').html`) and if the response is a 400 or maybe just a `OEmbed::UnknownResponse` we don't render the header or embed.

cbeer:
I kinda like option 2. If we have to start indexing contentMetadata data, I'd toss a couple other things onto the pile (for internal use finding fixtures objects, if nothing else...):

- a count of the resources
- the `//contentMetadata[@type]` value
- the `//contentMetadata/resource[@type]`values
- the `//contentMetadata/resource/file[@mimetype]`values
- the `//rightsMetadata/access[@type="read"]/machine/*` tag names?

cbeer:
I think the MARC records will need different handling. We don't pull in any information from SDR when indexing those records (for performance reasons, if nothing else).

Is there an opportunity in the SDR<->856 linking to filter those out (afaik, the embed viewer is the only thing the 856 entry gets us, so maybe there's not a use case there?)

cbeer:
[...]

PR #2935, opened 2022-07-21
Suppress embed for MODS resources when :dor_resource_count_isi field exists and is 0

corylown:
This will suppress the embed and header for MODS resources if we know there is no published content (`:dor_resource_count_isi` field is present and not a positive number). Because this is a new field requiring re-indexing, the logic allows the embed to display if the `:dor_resource_count_isi` field is not yet set on the record.

Partially addresses #2885

### DIFF
--- a/app/views/catalog/_show_mods.html.erb
+++ b/app/views/catalog/_show_mods.html.erb
@@ -4,7 +4,7 @@
 
 <div class="col-md-12 upper-record-metadata row">
   <%= render 'catalog/record/mods_upper_metadata_section' %>
-  <% if document.druid && (document['dor_resource_count_isi']&.positive? || !document.key?('dor_resource_count_isi')) %>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_show_mods.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="col-md-12 upper-record-metadata row">
   <%= render 'catalog/record/mods_upper_metadata_section' %>
-  <% if document.druid %>
+  <% if document.druid && (document['dor_resource_count_isi']&.positive? || !document.key?('dor_resource_count_isi')) %>
```
cbeer, line 7: Could we make a new method on SolrDocument for `document['dor_resource_count_isi']&.positive? || !document.key?('dor_resource_count_isi')` ? It'd give a good place to explain what we're doing, if nothing else 😁

PR #2939, opened 2022-07-22
HTML-safe titles from OSTI

thatbudakguy:
- Extract EDS-specific logic for show and index presenters
- Remove dead BlacklightHelper method
- Remove dead PresenterResearchStarter
- Add rendering step for html-safe titles


app/controllers/catalog_controller.rb
```diff
@@ -86,7 +86,9 @@ class CatalogController < ApplicationController
 
     # solr field configuration for search results/index views
     config.index.document_presenter_class = IndexDocumentPresenter
-    config.index.title_field = 'title_display'
+
+    # TODO: add a pipeline step that concatenates values
```
corylown, line 90: Is this `TODO` comment still needed?

thatbudakguy: oops, no — that became `TitleRenderingStep`. removed the comment, thanks for catching

corylown (approved):

PR #2941, opened 2022-07-22
Clean out some old settings + feature flags

cbeer:


app/views/feedback_mailer/_shared_closing.text.erb
```diff
@@ -1,5 +1,6 @@
 Message sent from: <%= @mailer_parser.url %>
 Host: <%= Settings.HOSTNAME %>
+Revision: '<%= Settings.REVISION&.slice(0..7) %>'
```
thatbudakguy, line 3: do we do this on other projects? because we totally should

thatbudakguy (approved):

corylown (approved):

PR #2944, opened 2022-07-25
Re-enable SAL-NEWARK paging

cbeer:
See VUF-8885


cbeer:
As far as we know, SAL-NEWARK is still closed for paging.

PR #2952, opened 2022-08-01
Add Last updated to librarian view

### LINKED ISSUE #983: Add last_updated from the solr document to the Librarian View
screen shot 2014-12-15 at 3 58 41 pm

marlo-longley:
Closes #983

📝 CHANGELOG update?
Add last updated to librarian view

### DIFF
--- a/app/views/catalog/librarian_view.html.erb
+++ b/app/views/catalog/librarian_view.html.erb
@@ -6,6 +6,11 @@
     </button>
   <% end %>
   <h3 class="modal-title"><%= t('blacklight.search.librarian_view.title') %></h3>
+  
+  <% if @document[:last_updated] %>
+    <p>Last updated in SearchWorks on <%= l @document[:last_updated].to_time, format: :long %></p>
+  <% end %>
+  
 </div>
 <% if @document.respond_to?(:to_marc) %>
   <%= render "marc_view" %>
--- a/spec/views/catalog/librarian_view.html.erb_spec.rb
+++ b/spec/views/catalog/librarian_view.html.erb_spec.rb
@@ -5,22 +5,24 @@ describe "catalog/librarian_view" do
   include MarcMetadataFixtures
   describe "MARC records" do
     before do
-      assign(:document, SolrDocument.new(id: '12345', marcxml: metadata1))
+      assign(:document, SolrDocument.new(id: '12345', last_updated: '2022-08-01T23:01:18Z', marcxml: metadata1))
       render
     end
 
     it "should render the marc_view" do
+      expect(rendered).to have_content('August  1, 2022 11:01pm')
       expect(rendered).to have_css('#marc_view')
     end
   end
 
   describe "MODS records" do
     before do
-      assign(:document, SolrDocument.new(id: '12345', modsxml: mods_everything))
+      assign(:document, SolrDocument.new(id: '12345', last_updated: '2022-08-01T23:01:18Z', modsxml: mods_everything))
       render
     end
 
     it "should render the mods_view" do
+      expect(rendered).to have_content('August  1, 2022 11:01pm')
       expect(rendered).to have_css('.mods-view')
     end
   end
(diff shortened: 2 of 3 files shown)


marlo-longley:
Curious to get feedback on this warning. I tried to follow the outline in [this BL PR](https://github.com/projectblacklight/blacklight/pull/2718) by adding `last_updated` to the `search_state_fields` params in the catalog controller. But I am still getting the warning. 

```
Processing by CatalogController#librarian_view as */*
  Parameters: {"id"=>"20"}
Unpermitted parameter: :id. Context: { controller: CatalogController, action: librarian_view, request: #<ActionDispatch::Request:0x00007faec2b94178>, params: {"controller"=>"catalog", "action"=>"librarian_view", "id"=>"20"} }
DEPRECATION WARNING: Blacklight 8 will filter out non-search parameter, including: id. (called from permit_search_params at /Users/marloelilongley/.gem/ruby/3.0.4/gems/blacklight-7.28.0/lib/blacklight/parameters.rb:77)
Solr query: get select {:qt=>"document", :ids=>"20"}
Solr fetch (13.6ms)
  Rendering catalog/librarian_view.html.erb
  Rendered catalog/_marc_view.html.erb (Duration: 6.5ms | Allocations: 9968)
  Rendered catalog/librarian_view.html.erb (Duration: 8.3ms | Allocations: 10769)
```

cbeer:
I think that's safe to ignore for now 🤷‍♂️

marlo-longley:
@cbeer Ok, thanks. I didn't know whether in that case I needed to add `:last_updated` to `search_state_fields` at all or not

app/views/catalog/librarian_view.html.erb
```diff
@@ -6,6 +6,11 @@
     </button>
   <% end %>
   <h3 class="modal-title"><%= t('blacklight.search.librarian_view.title') %></h3>
+  
+  <% if @document[:last_updated] %>
+    <p>Last updated in SearchWorks on <%= @document[:last_updated].to_time %></p>
```
cbeer, line 11: I'm guessing we'll want to format this value for display. I don't know that we've had to do that in searchworks yet, but we do in other projects, e.g.:

https://github.com/sul-dlss/mylibrary/blob/main/config/locales/en.yml#L33-L38

So I think it'd be safe to pull those over, and to use:
```suggestion
    <p>Last updated in SearchWorks on <%= l @document[:last_updated].to_time, format: :long %></p>
```

cbeer:
The error message was complaining about `:id`:
> Unpermitted parameter: :id.

Must be a new Rails warning or something, but should be fine... or Blacklight will have to address it eventually

app/controllers/catalog_controller.rb
```diff
@@ -59,6 +59,7 @@ class CatalogController < ApplicationController
 
     config.search_state_fields += [
       :prefix, # used to filter by database title
+      :last_updated
```
cbeer, line 62:
```suggestion
```

I think this is probably unnecessary?

spec/views/catalog/librarian_view.html.erb_spec.rb
```diff
@@ -5,22 +5,24 @@
   include MarcMetadataFixtures
   describe "MARC records" do
     before do
-      assign(:document, SolrDocument.new(id: '12345', marcxml: metadata1))
+      assign(:document, SolrDocument.new(id: '12345', last_updated: '2022-08-01T23:01:18Z', marcxml: metadata1))
       render
     end
 
     it "should render the marc_view" do
+      expect(rendered).to have_content(l('2022-08-01T23:01:18Z'.to_time, format: :long))
```
cbeer, line 13: For clarity about our expectations, I'd suggest checking the actual text:

```suggestion
      expect(rendered).to have_content('August  1, 2022 11:01pm')
```

(no guarantees I got that right 😬 )

Issue #2955, opened 2022-08-03
Output FOLIO JSON in the librarian view


cbeer:
It'd be super handy to see the folio struct fields that we've indexed (https://github.com/sul-dlss/searchworks_traject_indexer/blob/7c8125a86b4b87294cbcc92cf5ef5617ca927aa1/lib/traject/config/folio_config.rb#L222-L232) display in the "librarian view" modal (under the MARC?) so we can see the FOLIO data easier.

PR #2965, opened 2022-08-04
Create .folio_json export format

marlo-longley:
This just enhances work for https://github.com/sul-dlss/SearchWorks/issues/2955 to make it easier to debug using browser JSON formatting / expand / collapse

You can navigate to e.g. `/view/13839521.folio_json` and see something like this:
Screen Shot 2022-08-04 at 14 57 37

The main question I have about this PR is from the [BL docs](https://www.rubydoc.info/gems/blacklight/6.0.1/Blacklight/Document/Export) about exports:

> The convention for methods contained in extensions that transform to an exportable file of some kind is “export_as_*”. For instance, “export_as_marc21” would return a String object containing valid marc21, and “export_as_marcxml” would return a String object containing valid marcxml. The tokens used after “export_as” should normally be the format names as registered with Rails Mime::Type.

^ So according to BL docs `folio_json` is a bad name since it is not a MIME type. But we already use plain `.json`, and need to do custom merging of fields rather than dumping the whole document for this folio_json.

📝 CHANGELOG update?
Create .folio_json export format

### DIFF
--- a/app/models/concerns/folio_json_export.rb
+++ b/app/models/concerns/folio_json_export.rb
@@ -1,0 +1,12 @@
+module FolioJsonExport
+  def self.extended(document)
+    document.will_export_as(:folio_json, 'application/json')
+  end
+
+  def export_as_folio_json
+    {
+      "holdings_json_struct" => JSON.parse(self[:holdings_json_struct].first),
+      "folio_json_struct" => JSON.parse(self[:folio_json_struct].first)
+    }.to_json
+  end
+end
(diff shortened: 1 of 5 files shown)


app/models/concerns/folio_json_export.rb
```diff
@@ -0,0 +1,13 @@
+module FolioJsonExport
+  def self.extended(document)
+    document.will_export_as(:folio_json, 'application/json')
+  end
+
+  def export_as_folio_json
+    {
+      "holdings_json_struct" => JSON.parse(self[:holdings_json_struct].first),
+      "items_json_struct" => JSON.parse(self[:items_json_struct].first),
```
cbeer, line 9:
```suggestion
```

I've removed items_json_struct, now that we're hitting `/inventory-hierarchy/items-and-holdings` to get both the holdings and the items in one request 🤷‍♂️

marlo-longley: @cbeer I removed items_json_struct from this code in latest push

PR #2966, opened 2022-08-04
Handle missing location names in stackmap links

thatbudakguy:
Sends a honeybadger notification if we try to fetch the name
for a location that doesn't have a code->name mapping.


thatbudakguy:
prevents this from happening due to new location codes from FOLIO:

Screen Shot 2022-08-04 at 16 08 09

marlo-longley (approved):

Issue #22, opened 2014-05-02
Top Level Menu Hamburger


jkeck:
![nav-hamburger](https://cloud.githubusercontent.com/assets/96776/2867805/e5623134-d23b-11e3-8806-ef5ffe171617.png)

PR #2967, opened 2022-08-04
Fix icon rotation on toggle in mobile menu

### LINKED ISSUE #1461: Navbar toggle button rotates when submenus are opened / closed
Split off from #1408 

- when you toggle the Library services and Selections items within the collapsed navbar menu, they rotate the Menu icon - only toggling Menu itself should do this

### Files of interest:
- [search_navbar_top_menu.js](https://github.com/sul-dlss/SearchWorks/blob/master/app/assets/javascripts/search_navbar_top_menu.js) (should be renamed)
- [jquery.rotatehelper.js](https://github.com/sul-dlss/SearchWorks/blob/master/app/assets/javascripts/jquery.rotatehelper.js)

### Example
navbar_rotate_bug_newdesign

marlo-longley:
Fixes #1461

Currently the carets in the mobile menu don't rotate at all,
and when you toggle inner menu items, it causes the main Menu hamburger to incorrectly toggle (see live gif in issue)

This PR updates some incorrect jQuery selectors and stops event propagation from the inner menu items.

current (help caret not moving)
Screen Shot 2022-08-04 at 16 40 24

fix
Screen Shot 2022-08-04 at 16 32 45

current (main menu icon is in "closed" state even though it is open)
Screen Shot 2022-08-04 at 16 42 50

fix
Screen Shot 2022-08-04 at 16 44 09

📝 CHANGELOG update?
Fix icon rotation in mobile menu

### DIFF
--- a/app/assets/javascripts/jquery.rotatehelper.js
+++ b/app/assets/javascripts/jquery.rotatehelper.js
@@ -4,8 +4,6 @@
 
       Usage: $(selector).rotateHelper(option);
 
-    Available option values = '90cw' [default]
-
     This plugin :
       - listens for '[show|hide].bs[collapse|dropdown]' event on callee's
         parent or element specified in 'data-target' attribute
@@ -13,22 +11,15 @@
         specified option
   */
 
-  $.fn.rotateHelper = function(rotate) {
+  $.fn.rotateHelper = function() {
     return this.each(function() {
       var $button = $(this),
           $toRotate,
           $parent,
-          cssRotate90cw = 'search-navbar-dropdown-btn-rotate-90',
           cssRotate270cw = 'search-navbar-dropdown-btn-rotate-270',
-          cssRotate;
 
-      cssRotate   = getRotateCssClass();
       $parent     = getParent();
       $toRotate = getRotatableElement();
-
-      // in SearchWorks the main Menu in .navbar-header (mobile) uses data-toggle="collapse" 
-      $parent.on('show.bs.collapse', function() { $toRotate.addClass(cssRotate90cw); });
-      $parent.on('hide.bs.collapse', function() { $toRotate.removeClass(cssRotate90cw); });
 
       // in SearchWorks nested menu items beneath the main menu (mobile) use data-toggle="dropdown" 
       // Because they are nested, stop event propation so they only impact the intended element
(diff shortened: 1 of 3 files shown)


cbeer:
I can't remember what our usual approach for submenu indicators are; @ggeisler, how do you want the disclosure triangles to display?

ggeisler:
@marlo-longley I'd suggest we want the disclosure triangles to work the opposite of what I think you have here. In other words, when a menu is closed, the triangle should point to the right. When it is open, the triangle should face down. 

Although accordions are different things with different disclosure icons, this approach will match the behavior of the facet icons (point right when the facet is closed), which are in close proximity to this particular menu.

marlo-longley:
@ggeisler perfect, thank you-- I tried to remember how those usually work, but got it opposite!

marlo-longley:
@ggeisler sorry, just to confirm, you are OK with the the accordion behavior as-is in this PR?

ggeisler:
(Or, if you just mean the menu behavior, yes, just swap the icon direction and I think we're okay.)

marlo-longley:
Hi @ggeisler,  here is the updated icon rotation:

Screen Shot 2022-08-16 at 11 31 28

Screen Shot 2022-08-16 at 11 31 24


Screen Shot 2022-08-16 at 11 31 33

ggeisler:
@marlo-longley The arrow disclosure icon looks good to me now.

I guess I thought we were just discussing the arrow disclosure icon and hadn't focused on the hamburger menu icon aspect. That does look odd to me when the menu is open. I've not seen anyone rotate the hamburger icon before and not sure it makes sense (it would imply the menu opens horizontally instead of vertically?). That part feels like unintentional so perhaps we should fix that, too.

marlo-longley:
@ggeisler thanks for checking it out. I believe the hamburger rotation is intentional. The original bug reporting was about the hamburger rotation. https://github.com/sul-dlss/SearchWorks/issues/1461 The bug involved the rotation happening at the wrong triggers, but to me it definitely implied that the rotation was intended. 

Happy to change though, the ticket is quite old!

ggeisler:
@marlo-longley Yeah, I understand now. My preference would be to **not rotate** the hamburger menu at all. 

Although it clearly was intentional, going by [this ticket](https://github.com/sul-dlss/SearchWorks/issues/22) that choice was made 8 years ago, when these kind of responsive menus were pretty new. I don't think the idea of rotating the menu to indicate state has really caught on (at least, I've never seen it done anywhere else that I can think of) in the time since then. Also, it seems like seeing the menu items displayed is enough to indicate the menu is open, so the rotation aspect is unnecessary. That, combined with the fact that it is a bit odd to see it displayed sideways, makes me want to just ditch the rotation aspect.

marlo-longley:
@ggeisler ok I hear you! Thanks for looking into it. I'll remove the hamburger rotation altogether.

marlo-longley:
@ggeisler I agree this seems way less weird in 2022! I pushed a new commit that removes the hamburger icon rotation.

thatbudakguy (approved):

PR #2975, opened 2022-08-08
Use before/after shelfkey offsets to drive browse nearby pagination

cbeer:
Update browse nearby to use before/after shelfkey offsets (replacing forward/backward pagination from a known point). This   ensures consistent + complete pagination through high-count shelfkeys (but may also need https://github.com/sul-dlss/searchworks_traject_indexer/pull/679 for best display)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,3 +38,6 @@ RSpec/DescribeClass:
 Metrics/ParameterLists:
   Exclude:
     - 'app/components/**/*'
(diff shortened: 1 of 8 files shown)


thatbudakguy (approved):
tests definitely made reading this a lot easier. i also like the change in the naming of instance variables in the partial to be more clear about what's actually in there.

corylown (approved):

PR #2976, opened 2022-08-09
Refactor  Browse nearby

cbeer:
Extracted from #2975 with just the mechanical bits.


thatbudakguy (approved):
amazed at how much shorter this became without even changing it too radically!

PR #2977, opened 2022-08-09
Remove dead code

cbeer:


lib/search_query_modifier.rb
```diff
@@ -53,21 +31,9 @@ def has_query?
...
 
-  def self.stopwords
-    %w{and of the}
-  end
-
   private
 
   def has_filters_and_query?
     has_filters? && has_query?
   end
```
corylown, line 38:
```suggestion
```

Can this method be deleted also? Looks like it was only used by `#present?`

lib/search_query_modifier.rb
```diff
@@ -31,9 +31,4 @@ def has_query?
     @params[:q].present?
   end
```
corylown, line 33:
```suggestion
```

I think this is what Rubocop wants.

corylown (approved):

PR #2990, opened 2022-08-19
Move HTML link generation into Link class (replaces attribute from Solr)

### LINKED ISSUE #2969: Move link HTML generation from searchworks_traject_indexer to SearchWorks
Currently HTML link generation is handled in `searchworks_traject_indexer`: https://github.com/sul-dlss/searchworks_traject_indexer/blob/master/lib/marc_links.rb#L21

It might be preferable to generate the HTML structure for links within SearchWorks instead.

corylown:
Closes #2969 

sul-bento-app displays links that are generated in SearchWorks and appear in the JSON search results in `fulltext_link_html` so no changes are needed in bento: 

https://searchworks.stanford.edu/?f[access_facet][]=Online&search_field=search&q=Air+and+Marine+Operations+vision+and+strategy&format=json

### DIFF
--- a/lib/search_works/links.rb
+++ b/lib/search_works/links.rb
@@ -39,26 +39,38 @@ module SearchWorks
     end
 
     class Link
-      attr_accessor :html, :text, :href, :file_id, :druid, :type, :sort
+      include ActionView::Helpers::TagHelper
+
+      attr_accessor :href, :file_id, :druid, :type, :sort
 
       def initialize(options = {})
-        @html = options[:html]
-        @text = options[:text]
-        @href = options[:href]
-        @fulltext = options[:fulltext]
-        @stanford_only = options[:stanford_only]
-        @finding_aid = options[:finding_aid]
-        @sfx = options[:sfx]
-        @managed_purl = options[:managed_purl]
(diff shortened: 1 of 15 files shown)


lib/search_works/links.rb
```diff
@@ -39,26 +39,38 @@ def ill
...
+        @type = options[:type]
       end
 
       def ==(other)
-        [@html, @fulltext, @stanford_only, @finding_aid, @sfx] == [other.html, other.fulltext?, other.stanford_only?, other.finding_aid?, other.sfx?]
+        [@href, @fulltext, @stanford_only, @finding_aid, @sfx] == [other.href, other.fulltext?, other.stanford_only?, other.finding_aid?, other.sfx?]
+      end
+
+      def html
+        @html ||= [link_html, casalini_text, additional_text_html].compact.join(' ').strip
```
cbeer, line 69:
```suggestion
        @html ||= safe_join [link_html, casalini_text, additional_text_html].compact.map(&:strip), ' '
```

I think the regular join is going to lose its html safety otherwise?

corylown: you're right...and strip does the same

PR #2994, opened 2022-08-22
Configure the sms tool action

cbeer:


corylown (changes requested):
This fixes the issue, but the after-submit display is broken both in Firefox and Chrome:
Screen Shot 2022-08-23 at 9 13 59 AM

cbeer:
😬

corylown:
There is something oddly pleasing about the glitch though :-)

cbeer:
76884f4 ought to do it. I'm not sure why those `_success` templates wandered off, but Blacklight 6 -> 7 changed some stuff about the modal markup and it's incompatible with the ajax_modal.js we copied over from Blacklight 6 🤷‍♂️

corylown (approved):

PR #3010, opened 2022-09-13
Update request link text for Aeon requests

thatbudakguy:
See https://github.com/sul-dlss/SearchWorks/issues/3014

<s>As per conversation with @saseestone, we can change the request link text to "Request via Aeon" for `HOOVER`, `HV-ARCHIVE`, `RUMSEYMAP`, `SPEC-COLL`, and `ARS` because they're using Aeon for everything that's relevant. But, we can't do this for `EAST-ASIA` because they're only using Aeon for some item types/locations, and SearchWorks doesn't know enough to distinguish at the time the request link is generated (that's requests' job).</s>

This should be deployed at the same time as https://github.com/sul-dlss/sul-requests/pull/1339.


thatbudakguy:
I got kind of confused here; I hope this is the right approach. Some link texts change to "request via Aeon", but finding aid ones stay as "request via Finding Aid". But, our `FindingAidRequestLinkComponent` can go away, because we no longer send people to OAC — we always send them to requests first, and let requests handle bouncing them out to OAC.

app/components/location_request_link_component.rb
```diff
@@ -64,9 +60,15 @@ def external_link_icon
   end
 
   def link_text
+    return I18n.t('searchworks.request_link.finding_aid') if finding_aid?
```
cbeer, line 63: Isn't it possible there are records with findings aids that are not in a library that's doing Aeon?

thatbudakguy: It's definitely possible...would it make sense to move the check for "is it in an Aeon library" here too? I feel like logic is kinda bleeding over from requests into here

thatbudakguy:
Found some stuff that allowed me to condition link text on library + location and not just location, so we're in business.

thatbudakguy:
rebased & cleaned up.

config/settings.yml
```diff
@@ -392,8 +392,22 @@ mediated_locations:
     - LOCKED-STK
   SAL3:
     - PAGE-MP
+
+aeon_locations:
   RUMSEYMAP: "*"
   SPEC-COLL: "*"
+  HOOVER: "*"
+  SAL:
```
cbeer, line 400: > Currently enabled library/location combinations, and requests need to go to Aeon
> library: SAL, location: L-PAGE-EA, Item type = NH-INHOUSE
> library: SAL3, location: PAGE-AS, Item type = NH-INHOUSE
> library: SAL3, location: PAGE-EA, Item type = LCKSTK
> Note SAL3 PAGE-EA items that do NOT have item type = LCKSTK should display request links, but the request should go to Symphony.

Don't we also need to differentiate by item type, or is that done elsewhere?

thatbudakguy: Hmm. Do we need to borrow the whole `LocationRule` idea from requests, then? idk if Searchworks can differentiate by item type right now

thatbudakguy: Is the best way to do it to pull it out of `building_location_facet_ssim`, or is there a different accessor? I see some "item_type" related stuff but it's mostly in configs.

cbeer: I don't think we're using (or returning) `building_location_facet_ssim`. Your best bet is probably pulling the item types off the `Holdings::Item`: https://github.com/sul-dlss/SearchWorks/blob/master/lib/holdings/item.rb#L60

I did confirm with Sarah much earlier that Aeon-eligible material all have Aeon-eligible types, so at least there's that.

thatbudakguy: ok, took a stab at this using a config syntax similar to what we have over in requests.

PR #3012, opened 2022-09-13
Add author, pub date, and finding aid to availability.json

### LINKED ISSUE #3008: Add more metadata to the availability.json response
these responses, e.g. https://searchworks.stanford.edu/view/11835941/availability.json, get consumed by sul-requests and used to populate data when making a request. that data can then be used to send the relevant metadata out to external systems like Aeon (see https://github.com/sul-dlss/sul-requests/issues/1322) so that the request form there is pre-populated.

the assumption behind this ticket is that the more metadata we can send out to Aeon, the better — it will help staff disambiguate items and find the relevant item for the request more quickly. if this isn't true (or it's not helpful to have anything more than title, call number, barcode) then this ticket can be discarded.

the available fields we can populate on the Aeon request form, and thus the things we _could_ send if we have data, are:

- title (we have this data already)
- author/creator
- place of publication
- publisher
- date
- edition
- volume/box
- issue/copy
- call number (we have this data already)
- collection name/location

thatbudakguy:
Closes #3008

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -516,7 +516,13 @@ class CatalogController < ApplicationController
     return {} unless document.is_a?(SolrDocument)
 
     {
+      # this data is used by sul-requests and passed to Aeon
       title: document['title_display'],
+      author: document['author_person_display']&.first ||
+        document['author_corp_display']&.first ||
+        document['author_meeting_display']&.first,
+      pub_date: document['pub_date'],
+      finding_aid: document.index_links&.finding_aid&.first&.href,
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -516,7 +516,13 @@ def render_document_with_availability_as_json(document, live = true)
     return {} unless document.is_a?(SolrDocument)
 
     {
+      # this data is used by sul-requests and passed to Aeon
```
cbeer, line 519:
```suggestion
      # this data is used by sul-requests, and some of it is passed through to Aeon
```

PR #3016, opened 2022-09-20
Try some things to help ensure the URL is clickable

cbeer:
We don't want carriers to digest messages, but if they do, we want to make sure there's trailing whitespace to separate our message from any wrapping they inject

SW-4130


thatbudakguy (approved):
what an odd problem. thx for diving into this.

PR #3024, opened 2022-10-10
Upgrade to Jquery 3

jcoyne:
Jquery 1 is not supported an may suffer from XSS exploits.


jcoyne:
It looks like Bootstrap 3 doesn't support Jquery 3.

PR #3032, opened 2022-11-04
Display the last updated time using the app's time zone

cbeer:


jcoyne:
Would you consider using @github/time-elements `local-time` here? Then it can display in the timezone the current user is in: https://github.com/github/relative-time-element

cbeer:
Is it worth it? I think our staff generally expect Stanford-local time (but definitely not UTC) 🤷‍♂️

jcoyne (approved):

PR #3034, opened 2022-11-09
Sanitize MarcField display value output

cbeer:
Prior to #2647, we didn't do any html sanitization. Turns out, `safe_join` is too aggressive with html entities (see e.g. SW-4131), so we actually have to do some proper sanitization.

### DIFF
--- a/app/models/marc_field.rb
+++ b/app/models/marc_field.rb
@@ -4,6 +4,7 @@
 ###
 class MarcField
   include ActionView::Helpers::OutputSafetyHelper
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

jcoyne (approved):

corylown (commented):
If I'm understanding this right, I think we'll see that this fixes the encoded quotes but that angle brackets will be displayed in their encoded form -- maybe that's fine?

PR #3058, opened 2023-01-19
Test on ruby 3.2

jcoyne:


jcoyne:
This has already been done in #3144

PR #3067, opened 2023-01-30
Allow item public notes to wrap at a punctuation mark

### LINKED ISSUE #3061: Long note in access panel not wrapping correctly. Icon outside drawn box too.
We received a report from staff that a long Public note is displaying outside the At the Library access panel. I noticed the availability icon is also pulled outside the access panel.

Example record: https://searchworks.stanford.edu/view/10855038

image

cbeer:
Fixes #3061 

before:
Screen Shot 2023-01-30 at 15 32 38

after:
Screen Shot 2023-01-30 at 15 32 20

---

This PR also addresses some HTML markup issues in the catalog#index version of the access panel:
- [x] more/less buttons were injected straight into the `<table>` markup; now they are wrapped in a table row and have appropriate ARIA attributes
- [x] group each library location section in its own `<tbody>` so the location header can apply to the whole `rowgroup`
- [x] remove the need for explicit `.indent-callnumber` styling; just make it a property of the table


marlo-longley (approved):

marlo-longley:
This looks good to me @cbeer, the note wrapping as well as the more/less button for long-lists works as expected.

PR #3068, opened 2023-01-31
Document how to get data in local development

marlo-longley:
I couldn't find anywhere that described how to do this...

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Start the rails app
 
     $ rails s
 
+## Getting data in development
+
+For fixture data, you can run `solr_wrapper` locally as above, and populate sample data using `rake searchworks:fixtures`
+
 ## "Logging in" as a User in development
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,6 +31,12 @@ Start the rails app
 
     $ rails s
 
+## Getting data in development
+
+For fixture data, you can run `solr_wrapper` locally as above, and populate sample data using `rake searchworks:fixtures`
+
+For more realistic data, you can tunnel into a Solr instance. For example, in one terminal window, you can ssh to staging Solr: `ssh -L 8080:sul-solr-stage-a:80 blacklight@sw-webapp-a`. Then in another window, you can start your app: `SOLR_URL=http://localhost:8080/solr/searchworks-stage be rails s -p 3000`.
```
jcoyne, line 38: I'm not sure this practice should be promulgated.  We're one typo away from accidentally causing a production outage of our most visible systems.

marlo-longley: Fair enough. It seems like a common practice on our team that people keep having to ask about on Slack. But I'm happy to remove it from the docs.

jcoyne (approved):

Issue #1766, opened 2017-09-12
[SearchWorks/production] Faraday::TimeoutError: too many connection resets


drh-stanford:
[SearchWorks/production] Faraday::TimeoutError: too many connection resets (due to execution expired - Net::OpenTimeout) after 0 requests on 70099183618440, last used 1505247344.7452774 seconds ago

## Backtrace

    line 7 of [PROJECT_ROOT]/app/models/eds/session.rb: search
    line 26 of [PROJECT_ROOT]/app/models/eds/repository.rb: eds_search
    line 5 of [PROJECT_ROOT]/app/models/eds/repository.rb: block in search

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/34626585)

drh-stanford:
See https://github.com/ebsco/edsapi-ruby/issues/78

drh-stanford:
We're still seeing this with the new timeout (from 5 to 10 seconds).

<img width="1090" alt="screen shot 2017-09-20 at 10 55 26 am" src="https://user-images.githubusercontent.com/1861171/30659311-43b860e8-9df2-11e7-8bfe-8c0625a1bcda.png">

bmckinney:
Can you send me the honeybadger backtrace?

drh-stanford:
Sure

```
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:880 :in `initialize`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:880 :in `open`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:880 :in `block in connect`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:101 :in `timeout`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:878 :in `connect`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:1481 :in `begin_transport`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:1433 :in `transport_request`
/usr/local/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/net/http.rb:1407 :in `request`
[...]

PR #3089, opened 2023-03-22
Throttle article searching more aggressively to protect EDS API

thatbudakguy:
Ref #1766

This borrows DLME's rack-attack config to send headers to clients informing them about throttle settings and when they'll get unblocked. It applies a stricter throttle (12 requests per minute) to article searching but disables all throttling for Stanford local clients.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -37,4 +37,7 @@ if Settings.THROTTLE_TRAFFIC
 
     [429, headers, ["Throttled\n"]]
   end
+
+  # Disable throttling for Stanford-local users
+  Rack::Attack.safelist_ip("171.64.0.0/14")
 end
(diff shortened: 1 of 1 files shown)


config/initializers/rack_attack.rb
```diff
@@ -1,23 +1,40 @@
...
-  if Settings.THROTTLE_TRAFFIC
-    throttle('req/ip', limit: 300, period: 5.minutes) do |req|
-      req.ip unless req.path.start_with?('/assets')
-    end
+  Rack::Attack.throttle('req/ip', limit: 300, period: 5.minutes) do |req|
+    req.ip unless req.path.start_with?('/assets')
+  end
+
+  # Throttle article searching more aggressively to protect EDS API (30rpm)
+  Rack::Attack.throttle('articles/ip', limit: 150, period: 5.minutes) do |req|
```
corylown, line 22: Just a thought: if we excluded Stanford IPs `171.64.0.0/14` from this throttling, maybe we could get away with being more aggressive -- something like 60/5min or 12rpm?

thatbudakguy: Updated to use 60/5min

corylown (approved):

PR #3097, opened 2023-04-05
Read config values from the environment

thatbudakguy:
This ensures the SETTINGS__THROTTLE_TRAFFIC envvar will take effect

### DIFF
--- a/config/initializers/rails_config.rb
+++ b/config/initializers/rails_config.rb
@@ -1,3 +1,5 @@
 Config.setup do |config|
   config.const_name = "Settings"
+  config.use_env = true
+  config.env_separator = "__"
 end


config/initializers/rails_config.rb
```diff
@@ -1,3 +1,4 @@
 Config.setup do |config|
   config.const_name = "Settings"
+  config.use_env = true
```
jcoyne, line 3: Do you also need to set:
```
config.env_separator = '__'
config.env_converter = :downcase
```

jcoyne (approved):
Do you a

config/initializers/rails_config.rb
thatbudakguy: looks like `:downcase` is the default, but the default separator is `.`, so I do need to set that. good catch

thatbudakguy: updated

jcoyne (approved):

thatbudakguy:
Using `shared_configs` instead per @cbeer: https://github.com/sul-dlss/shared_configs/pull/1951/files

PR #3120, opened 2023-04-24
Use Hash#each_values

jcoyne:
This means that ruby doesn't have to allocate an array before iterating.


dnoneill:
@jcoyne can you put something in the description about why this was implemented? Is it just speed?

jcoyne:
@dnoneill yes, it's a micro optimization, but it's good practice to use it when possible.

dnoneill (approved):

PR #3126, opened 2023-05-02
Import Lane proxies from file they provide

### LINKED ISSUE #3114: Change source of LANE-MED's proxy prefix hosts lists
Lane maintains its proxy prefix hosts list here: https://lane.stanford.edu/eresources/ezproxy-servers.txt

The values at lines beginning with `T ` contain the hosts that should have the prefix applied. They would like it if we could have SearchWorks apply the values from this file automatically instead of having to maintain a separate file in SearchWorks.

See: https://docs.google.com/document/d/1IMWgP6i1gIJ8WlY-XOWMv-g1JaCeEnYfTzdh-ppRUdk/edit#

jcoyne:
Fixes #3114

Run this command with
```
bin/rake proxies:import_law
```

A cronjob is set up at https://sul-ci-prod.stanford.edu/job/Continuous%20Deployment/job/searchworks-ezproxy-lane-import/


corylown (changes requested):
@jcoyne the only changes needed are related to naming. This is Lane Medical Library's proxy file, not Law's. So the files, class, and job naming should be changed from `law` to `lane`.

corylown (approved):

PR #3132, opened 2023-05-09
Add a service for evaluating FOLIO circ rules

### LINKED ISSUE #3149: Retrieve the relevant circ policies for an item using FOLIO circ rules
spun off from #3129 as a prerequisite; copied analysis from that ticket.

- [x] download the circ rules file and commit it to the repo as plain text
- [x] create an index that can find the highest-priority rule that applies to an item
- [x] make it easy to pass a `Holdings::Item` to the index for lookup
- [x] make a FOLIO API query for policy information, save it to a file, and commit it to the repo
- [x] make the rule search function return all the relevant information about policies for an item
- [x] confirm that the indexing strategy doesn't cause performance problems in realistic scenarios (e.g. an instance with many different items)

thatbudakguy:
Closes #3149 

<s>Requires ruby 3.2 because of usage of [`Data`](https://ruby-doc.org/3.2.1/Data.html).</s>

This PR adds a service for evaluating circ rules to find the applicable one. The idea is that you provide a `Holdings::Item` and the service selects the applicable circ rule, which in turn gives you the set of policies for that item (loaning/borrowing it, requesting/paging it, overdue/fines for it, etc).

To do this, the `holdings_json_struct` field is used to make extra FOLIO-specific data (names, UUIDs) available on `Holdings::Item` so that it can be provided at query time to the service. That data is encapsulated in FOLIO-specific `Item` and `Location` classes.

Note that the tests use human-readable strings instead of UUIDs for both the rules and the item's properties — the service doesn't care. In practice we will use UUIDs, but I think it makes the tests easier to understand to use real names. They also simplify the actual policy objects and just use strings.

For an overview of some different approaches @cbeer and I tried, a script for profiling them, and a list of UUIDs that might be useful in testing, see https://gist.github.com/thatbudakguy/fb15100375d9c12ee469488021344adb.


cbeer:
> Requires ruby 3.2 because of usage of [Data](https://ruby-doc.org/3.2.1/Data.html).

Can we just use `Struct` for now, or is `Data` actually getting us something more?

thatbudakguy:
`Data` is immutable, which I like, but the main reason I used it is because it supports creation using both keyword and positional arguments where struct only does positional. I thought this was nice for the `#from_dynamic` methods, so you can do stuff like:
https://github.com/sul-dlss/SearchWorks/blob/b604460d4bdd6c4c0067c729ff3d5be766ac026f/app/models/folio/location.rb#L24-L26
where you deconstruct the parsed JSON directly into keyword arguments to `#new`. remembering the order of struct fields makes me a little nervous.

If it's a pain to go to 3.2 just for this, though, I don't think it's a deal breaker to go back to `Struct`. It just seems like `Data` is nicer in a lot of cases.

cbeer:
Ah - Ruby 3.2 `Struct` changed to make positional arguments the default, but for now you could just do:

```
Struct.new(:a, :b, keyword_init: true)
```

thatbudakguy:
oh!! well that's nice to know about — I'll just use that

PR #3158, opened 2023-05-22
Remove view context from the model context

jcoyne:
Now the view calls the component, rather than the view calling the model calling the component


jcoyne:
We're sort of abusing the component by calling `render?` externally, so this won't work.

Issue #3147, opened 2023-05-16
Location level and Item level link in search result


saseestone:
Searching for Fixture items for our FOLIO Request Triggers spreadsheet, I came across this [search result ](https://searchworks.stanford.edu/?f%5Baccess_facet%5D%5B%5D=At+the+Library&f%5Bbuilding_facet%5D%5B%5D=Business&f%5Bformat_main_ssim%5D%5B%5D=Book) - the fifth item in the results list is showing both a location level request link AND an item level request link. We should be displaying only the Location level request link. 

<img width="785" alt="image" src="https://github.com/sul-dlss/SearchWorks/assets/14897546/925dc7b1-3632-4687-887f-25486b8d2b2d">

When you click through to the record (https://searchworks.stanford.edu/view/14655229) we are correctly displaying only the location level request link.

jcoyne:
I'm not currently able to reproduce this as the status changed to "checked out"
![Image](https://github.com/sul-dlss/SearchWorks/assets/92044/9a4e1761-8293-40a2-a9fa-c62c02d6a586)

cbeer:
Here's more examples:
https://searchworks.stanford.edu/?search_field=search&q=13178633
https://searchworks.stanford.edu/?q=14171461&search_field=search
https://searchworks.stanford.edu/?q=14171463&search_field=search

PR #3159, opened 2023-05-22
Remove unused logic around unavailable_current_location per library

jcoyne:
It is only ever used globally.  This simplifies the lookup and makes the code easier to read


jcoyne:
@cbeer I'm finding the current code impenetrable, and removing the unneeded bits makes it a bit easier to follow.  This is related to my investigation on https://github.com/sul-dlss/SearchWorks/issues/3147

PR #3160, opened 2023-05-22
Extract logic about whether to display an item request button to a separate class

jcoyne:
We don't want the model to be aware of view layer (component) concerns.  We also don't want external consumers calling `ItemComponent#render?`

### DIFF
--- a/lib/holdings/item.rb
+++ b/lib/holdings/item.rb
@@ -150,8 +150,8 @@ class Holdings
       }
     end
 
-    def request_link
-      @request_link ||= ItemRequestLinkComponent.new(item: self)
+    def request_link_policy
+      @request_link_policy ||= ItemRequestLinkPolicy.new(item: self)
     end
 
     # create sorting key for spine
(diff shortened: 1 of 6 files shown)


lib/holdings/item.rb
```diff
@@ -154,6 +154,34 @@ def request_link
...
+      Settings.aeon_locations[library] == '*' ||
+        Settings.aeon_locations.dig(library, home_location) == "*" ||
+        Settings.aeon_locations.dig(library, home_location)&.include?(type)
+    end
+
+    def in_nonrequestable_location?
+      (Settings.nonrequestable_locations[library] || Settings.nonrequestable_locations.default).include?(home_location)
+    end
+
+    def requestable?
```
cbeer, line 179: If we're going to do this refactor, maybe `#requestable?` isn't a good name for the method. We're certainly confusing item properties (like a non-requestable location or being on reserve), but an item is still requestable aeon-paged or mediated-paged, we just don't show an item-level link for it because the location-level link takes care of it.

jcoyne: Alright, I had another stab at this, extracting the business logic about requests link display to a separate class.

PR #3163, opened 2023-05-23
Create models for Folio Holdings and Items

jcoyne:


cbeer:
> having a bunch of logic around matching codes, etc.

I think I'm worried some of that logic may be important? I know our indexer's notion of a library/location doesn't neatly translate to Symphony's library/location codes (e.g. we collapse locations that translate to the same string). I'm not sure where FOLIO's holdings records fall in that.

I wonder too whether we actually want to expose those big, long, ugly UUIDs in the URLs. There's something nice about keeping them human-readable, isn't there?

It'll also make it more difficult to do side-by-side testing. Being able to copy query parameters from e.g. requests.stanford.edu and throw them at the folio-dev instance seems like a good feature for migration.If we can avoid doing this until post-migration, I think we'll be happier 🤷‍♂️

app/models/folio/holding.rb
```diff
@@ -0,0 +1,15 @@
+module Folio
+  class Holding
+    def initialize(id:, location:)
+      @id = id
+      @location = location
+    end
+
+    def self.from_dynamic(json)
```
thatbudakguy, line 8: this is a minor thing, but why `dynamic`? I'd just call these methods `from_json` (unless there's some convention here that I don't know about)

thatbudakguy (commented):
idk if we want to make the architectural change to send FOLIO UUIDs to requests yet, but I like the way the new FOLIO models in this PR are structured and I think they're needed for/blocking other work, so could we merge just those changes (in `app/models`) separately?

app/models/folio/holding.rb
jcoyne: That isn't json, it's a ruby hash.

thatbudakguy: That's what I said about `#as_json`, but apparently that's an ActiveRecord thing for "the hash that you _then_ turn into JSON"...

thatbudakguy: See: https://github.com/sul-dlss/searchworks_traject_indexer/blob/462a9b5f94b83c5ae1ef03548bf502432c6b7a67/lib/folio_record.rb#L137C1-L140

thatbudakguy: i'd be fine with `#from_hash` too, fwiw

jcoyne: That's fine, this is just copying how [quicktype](https://github.com/quicktype/quicktype/blob/2b8ab70650ec37fcd97e660dde62bf3d70470763/packages/quicktype-core/src/language/ruby/index.ts#L271) generates these sort of data classes.

app/models/folio/holding_location.rb
```diff
@@ -0,0 +1,13 @@
...
+      @effective_location = effective_location
+    end
+
+    def self.from_dynamic(json)
+      new(effective_location: Location.from_dynamic(json.fetch('effectiveLocation')))
+    end
+
+    attr_reader :effective_location
+  end
+end
```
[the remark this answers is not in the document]
jcoyne: That structure exists, but if we're not using it, it would just add bloat without offering anything.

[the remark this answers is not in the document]
thatbudakguy: since we know for sure we need effective location, why not just make `effective_location` a `Folio::Location` on `Folio::Holding` and add the rest later if we need them?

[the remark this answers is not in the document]
jcoyne: I wanted to follow the json structure, like `holding.location.effective_location`, which is what this does.  We can add `holding.location.permanent_location`, etc later if needed.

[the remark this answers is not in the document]
thatbudakguy: isn't the JSON structure just an artifact of how we wrote the postgres query? I'm saying `HoldingLocation` will only ever have `Location`s as properties; there's no unique logic or values that justify it having a class. Let's just put the `Location` directly on `Holding`. I don't think there will ever be data associated with the relationship itself, right?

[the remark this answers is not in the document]
thatbudakguy: we confirmed the RAMLs don't have a top-level `location` key; I'm gonna change the postgres query to also flatten out the various permanent/temporary/effective location keys.

thatbudakguy (approved):

PR #3174, opened 2023-05-26
Add translation map for folio location codes

jcoyne:
This will allow us to pass the Folio code to sul-requests, so it can find the holdings


cbeer:
I'm not sure I understand why searchworks should know about the FOLIO location codes, or why we'd want to pass it over to sul-requests (for the same reason we don't want to pass the folio holdings uuids).

jcoyne:
@cbeer This will help us in the long run as sul-requests can look at the folio_code to look up the location.  After we're fully on folio we can drop the old library/location codes.  You mentioned not having UUIDs in the URIs as that made them ugly looking, so I though this was better looking.

jcoyne:
Moved to https://github.com/sul-dlss/sul-requests/pull/1492

Issue #3135, opened 2023-05-11
Provide real-time availability lookups in Searchworks with FOLIO


cbeer:
We have two obvious choices:
- continue to implement an asynchronous, AJAX driven availability lookup (replacing vufind.pl with... something that queries FOLIO APIs) for real-time lookups
- (measure & tweak to) determine that FOLIO indexing latency is small enough that we can index the current item status quickly enough that the index is as reliable as any real-world race condition (e.g. a patron browsing stacks for an item that someone already took to go check out)

The second option is kinda appealing architecturally (and sets us up for a "and I can check it out today" facet), if we can get away with it. 

- [ ] figure out what information the vufind.pl script returns (and that we're using in searchworks)
- [ ] make sure we're indexing the right information (or figure out which FOLIO APIs return it for option 1)
- [ ] display (near)-real time current location(/availability) information for parity with Symphony-driven displays

Some work on this was possibility done (or considered) in the MVP work cycle.

jcoyne:
Our current latency is something like 1 hour (cron runs hourly) + 30 min (for querying) + 15 min (for solr commit). This latency seems pretty large to me. Who can say whether this is small enough?

cbeer:
If we don't think there's a way to get the latency under 10-15 minutes, I think we'll need to make requests to the FOLIO APIs. Perhaps we should figure out the async pattern first (for parity) and then figure out how to improve our indexing latency.

PR #3178, opened 2023-05-31
Add a FolioClient service

corylown:
Set up for #3135

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -32,14 +32,6 @@ class FolioClient
     @tenant = tenant
   end
 
-  def post(path, **kwargs)
-    authenticated_request(path, method: :post, **kwargs)
-  end
-
-  def post_json(path, **kwargs)
-    parse(post(path, **kwargs))
-  end
-
   # FOLIO Edge API - Real Time Availability Check
   # https://s3.amazonaws.com/foliodocs/api/edge-rtac/p/edge-rtac.html
   def real_time_availability(instance_ids:, full_periodicals: true)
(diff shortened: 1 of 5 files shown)


app/services/folio_client.rb
```diff
@@ -0,0 +1,95 @@
...
+    authenticated_request(path, method: :post, **kwargs)
+  end
+
+  def post_json(path, **kwargs)
+    parse(post(path, **kwargs))
+  end
+
+  # FOLIO Edge API - Real Time Availability Check
+  # https://s3.amazonaws.com/foliodocs/api/edge-rtac/p/edge-rtac.html
+  def rtac(instance_ids:, full_periodicals: true)
```
jcoyne, line 53: Can we call this `real_time_availability`?

app/services/folio_client.rb
```diff
@@ -0,0 +1,95 @@
...
+    parse(get(path, **kwargs))
+  end
+
+  def post(path, **kwargs)
+    authenticated_request(path, method: :post, **kwargs)
+  end
+
+  def post_json(path, **kwargs)
+    parse(post(path, **kwargs))
+  end
```
jcoyne, line 49: Can these be private?

jcoyne (changes requested):

app/services/folio_client.rb
corylown: yes, sounds good

app/services/folio_client.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require 'http'
+
+class FolioClient
+  DEFAULT_HEADERS = {
+    accept: 'application/json, text/plain',
```
jcoyne, line 7: If we're going to parse JSON from the response, then `text/plain` isn't really acceptable to us.

jcoyne (approved):

spec/services/folio_client_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe FolioClient do
+  subject(:client) { described_class.new(url:) }
+
+  let(:url) { 'https://example.com' }
```
jcoyne, line 8: Should a different host be considered, because example.com is in the allowed list?

PR #3186, opened 2023-06-01
Use LiveLookup::Folio to update availability information when enabled

### LINKED ISSUE #3135: Provide real-time availability lookups in Searchworks with FOLIO

corylown:
Closes #3135

### DIFF
--- a/app/assets/javascripts/jquery.live-lookup.js
+++ b/app/assets/javascripts/jquery.live-lookup.js
@@ -17,18 +17,29 @@
       $.getJSON(live_lookup_url, function(data){
           for (var i=0, length=data.length; i < length; i++) {
             var live_data = data[i];
-            var dom_item = $("[data-barcode='" + live_data.barcode + "']");
+
+            if (live_data.service == 'FOLIO') {
+              item_id = live_data.item_uuid
+              item_status = live_data.status
+              item_is_available = live_data.is_available
+            } else {
+              item_id = live_data.barcode
+              item_status = live_data.current_location
+              item_is_available = !live_data.due_date
+            }
+
+            var dom_item = $("[data-item-id='" + item_id + "']");
             var target = $(dom_item.data('status-target'), dom_item);
             var current_location = $('.current-location', dom_item)
             var status_text = target.next('.status-text');
-            if ( live_data.current_location ) {
-              current_location.html(live_data.current_location);
+            if ( item_status ) {
+              current_location.html(item_status);
             }
             if ( live_data.due_date ) {
               current_location.append(' Due ' + live_data.due_date);
             }
 
-            if ( live_data.due_date && (target.hasClass('unknown') || target.hasClass('page')) ) {
+            if ( !item_is_available && (target.hasClass('unknown') || target.hasClass('page')) ) {
               target.removeClass('unknown');
               target.removeClass('page');
               target.addClass('unavailable');
@@ -43,7 +54,7 @@
                 $('.request-link', dom_item).html(link);
               }
             }
-            if ( !live_data.due_date && dom_item.length > 0  && target.hasClass('unknown')) {
+            if ( item_is_available && dom_item.length > 0  && target.hasClass('unknown')) {
               target.removeClass('unknown');
               target.addClass('available');
               status_text.text(status_text.data('available-text'));
--- a/app/assets/javascripts/jquery.live-lookup.js
+++ b/app/assets/javascripts/jquery.live-lookup.js
@@ -18,28 +18,18 @@
           for (var i=0, length=data.length; i < length; i++) {
             var live_data = data[i];
 
-            if (live_data.service == 'FOLIO') {
-              item_id = live_data.item_uuid
-              item_status = live_data.status
-              item_is_available = live_data.is_available
-            } else {
-              item_id = live_data.barcode
-              item_status = live_data.current_location
-              item_is_available = !live_data.due_date
-            }
-
-            var dom_item = $("[data-item-id='" + item_id + "']");
+            var dom_item = $("[data-item-id='" + live_data.item_id + "']");
             var target = $(dom_item.data('status-target'), dom_item);
             var current_location = $('.current-location', dom_item)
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/jquery.live-lookup.js
```diff
@@ -17,18 +17,29 @@
       $.getJSON(live_lookup_url, function(data){
           for (var i=0, length=data.length; i < length; i++) {
             var live_data = data[i];
-            var dom_item = $("[data-barcode='" + live_data.barcode + "']");
+
+            if (live_data.service == 'FOLIO') {
```
cbeer, line 21: Maybe I should have asked on #3180, but I'm wondering if we want to reconcile the two APIs so we don't need to do this? If Sirsi returned item_id, status, and is_available fields we wouldn't have to care?

app/components/access_panels/library_location_component.html.erb
```diff
@@ -38,7 +38,7 @@
           <% end %>
         </td>
         <% if item.present? %>
-          <td class="item-availability" data-live-lookup-id="<%= document[:id] %>" data-status-target=".availability-icon" data-barcode="<%= item.barcode if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if !location_request_link.render? && item.circulates? && !item.request_link.render? %>>
+          <td class="item-availability" data-live-lookup-id="<%= document.live_lookup_id %>" data-status-target=".availability-icon" data-item-id="<%= document.barcode_to_uuid_mapping.fetch(item.barcode, item.barcode) if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if !location_request_link.render? && item.circulates? && !item.request_link.render? %>>
```
cbeer, line 41: Is `item` here a `Holdings::Item`? Maybe wecould give it a `live_lookup_item_id` accessor and bury the complexity in the model?

app/models/concerns/live_lookup_ids.rb
```diff
@@ -0,0 +1,24 @@
+# Mixin that adds an abstraction layer for bib/instance and item ids needed for live lookup
+# of availability to account for differences between Sirsi and FOLIO
+
+module LiveLookupIds
+  # For Sirsi use the catalog key for availability lookup; for FOLIO use the Instance UUID
+  def live_lookup_id
+    return self[:uuid_ssi] if Settings.FOLIO_LIVE_LOOKUP && fetch(:uuid_ssi, nil)
+
+    self[:id]
+  end
```
cbeer, line 10: Is there a case where we want to use FOLIO data, but hit Sirsi for availability? I'd be comfortable with
```suggestion
  def live_lookup_id
   self[:uuid_ssi] || self[:id]
  end
```

unless there's a good reason not to?

app/models/concerns/live_lookup_ids.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  # TODO: We can remove this and adjust how we get the item id for live lookup after we add
+  #   a mapping in searchworks_traject_indexer between barcodes and uuids in each solr document,
+  #   either in its own field or as part of the item_display field.
+  #
+  # @return [Hash] maps item barcodes to item UUIDs or an empty hash if FOLIO_LIVE_LOOKUP is false
+  def barcode_to_uuid_mapping
+    return {} unless Settings.FOLIO_LIVE_LOOKUP && fetch(:holdings_json_struct, nil)
+
+    JSON.parse(self[:holdings_json_struct].first).fetch('items', []).filter_map do |item|
```
cbeer, line 20: I think Nick developed something similar in parallel:
https://github.com/sul-dlss/SearchWorks/blob/ad7210b74a67600db98ebd82b0d4aa45218b2a3a/lib/holdings/item.rb#L228-L230

app/assets/javascripts/jquery.live-lookup.js
corylown: I struggled with what to do about this. I opted to align the `LiveStatus::Folio` response more with Folio's view of the world, thinking it will be less confusing in the long run if we describe the response accurately. I think we're going to need to be clear about what's a location and what's a status with Folio and this was an attempt to try to do that.

app/components/access_panels/library_location_component.html.erb
corylown: Hmm...maybe, I'll take a look.

app/assets/javascripts/jquery.live-lookup.js
corylown: Sorry, I misread. I'll see if changing the Sirsi response to look more like Folio would break anything.

app/services/live_lookup/sirsi.rb
```diff
@@ -63,13 +63,14 @@ def initialize(record)
 
       def as_json
         {
-          barcode:,
+          item_id:,
```
cbeer, line 66: Are we definitely the only consumers of this response, or should we keep barcode and current_location around (duplicating the data) for backwards compatibility?

corylown: Either way, I guess we'll find out when we switch over to folio 😬

PR #3188, opened 2023-06-02
Update Lane Library EZProxy config

sul-devops-team:


corylown:
Irina is checking Lane on whether this change is intentional (removing 200 hosts)

corylown:
Lane did not intend to remove 200 hosts from their list. They've since fixed it. Closing this.

PR #3189, opened 2023-06-02
Add a final sorter to ensure a consistent sort among rules of equal FOLIO criteria weighting for human readability

cbeer:


jcoyne (commented):
Can a test be added for this change?

thatbudakguy:
I'm going to merge it; this PR doesn't seem like the right place to bring in tests related to parsing the circ rules file...we don't have them because the code started as a spike, but we should probably talk about if we need them and how to write them at some point.

thatbudakguy (approved):

PR #3226, opened 2023-06-21
Conditionally display warning about FOLIO migration data

### LINKED ISSUE #3224: Display disclaimer about FOLIO data
The text should read: "<b>MIGRATION ERROR</b>: FOLIO is missing source record data; this bibliographic data was derived from the FOLIO instance metadata."

Screenshot 2023-06-21 at 11 08 04

Conditions to display the disclaimer:
- FOLIO json says `"source": "MARC"`
- in `marc_json_struct` the leader has first 8 characters blank

Test with:
(should have disclaimer)
https://searchworks-folio-dev.stanford.edu/view/9968376

https://searchworks-folio-dev.stanford.edu/view/9999944

https://searchworks-folio-dev.stanford.edu/view/5718135

(should not have disclaimer)
https://searchworks-folio-dev.stanford.edu/view/14665307

https://searchworks-folio-dev.stanford.edu/view/1

marlo-longley:
Closes #3224 

Screenshot 2023-06-21 at 12 58 14

Screenshot 2023-06-21 at 12 58 51

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -139,4 +139,15 @@ module ApplicationHelper
   def search_type_name
     t("searchworks.search_dropdown.#{controller_name}.label")
   end
+
+  # TODO: remove after migration is complete
+  def folio_disclaimer(document)
+    # this flag is set in searchworks-folio-dev shared_configs
+    return false unless Settings.folio.disclaimer
+    return false unless document['folio_json_struct'] && document['marc_json_struct']
+
+    marc_source = JSON.parse(document['folio_json_struct']&.first)&.dig('instance', 'source') == 'MARC'
+    # currently FOLIO MARC records have a leader that starts with 8 spaces
+    marc_source && document.to_marc&.leader&.match(/^ {8}/)
+  end
 end
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -3,6 +3,13 @@
 
     <%= render "shared/record/record_toolbar" %>
 
+    <!-- TODO: remove after migration is complete -->
+    <% if folio_disclaimer(@document) %>
+      <div class="alert alert-warning text-center" role="alert">
+       <span class="glyphicon glyphicon-warning-sign"></span>&nbsp;<b>MIGRATION ERROR</b>: FOLIO is missing source record data; this bibliographic data was derived from the FOLIO instance metadata.
+      </div>
+    <% end %>
+
     <% @page_title = t 'blacklight.search.show.title_html', :document_title => document_presenter(@document).html_title, :application_name => "#{I18n.t('blacklight.application_name')} catalog" -%>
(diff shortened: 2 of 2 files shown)


app/views/catalog/show.html.erb
```diff
@@ -3,6 +3,13 @@
 
     <%= render "shared/record/record_toolbar" %>
 
+    <!-- TODO: remove after migration is complete -->
+    <% if folio_disclaimer(@document) %>
+      <div class="alert alert-warning text-center" role="alert">
+       <b>MIGRATION ERROR</b>: FOLIO is missing source record data; this bibliographic data was derived from the FOLIO instance metadata.
```
thatbudakguy, line 9: tiny thing, but could we throw an icon of some sort in here? maybe a "caution" or "danger" indicator? i think it'd help draw the eye to it even more.

marlo-longley: How's this? If it looks OK I will push up the code.
Screenshot 2023-06-21 at 14 46 47

marlo-longley: I could even put them on both sides of the text haha

marlo-longley: I pushed up the icon.

app/helpers/application_helper.rb
```diff
@@ -139,4 +139,14 @@ def link_to_library_website_search
   def search_type_name
     t("searchworks.search_dropdown.#{controller_name}.label")
   end
+
+  # TODO: remove after migration is complete
+  def folio_disclaimer(document)
+    return false unless document['folio_json_struct'] && document['marc_json_struct']
```
cbeer, line 145: I wonder if we should add a feature flag in Settings, so we can make sure this doesn't show up in -prod?

app/helpers/application_helper.rb
```diff
@@ -139,4 +139,14 @@ def link_to_library_website_search
...
+
+    marc_source = JSON.parse(document['folio_json_struct']&.first)&.dig('instance', 'source') == 'MARC'
+    leader = JSON.parse(document['marc_json_struct']&.first)&.dig('leader')
+    # currently FOLIO MARC records have a leader that starts with 8 spaces
+    marc_source && leader&.match(/^ {8}/)
```
cbeer, line 150: I think we could use the already-parsed MARC record instead of re-parsing it?

```suggestion
    marc_source  && document.to_marc&.leader&.match(/^ {8}/)
```

app/helpers/application_helper.rb
marlo-longley: so it would be `return false unless document['folio_json_struct'] && document['marc_json_struct'] && Settings.the_feature_flag` ?

thatbudakguy: i might even split it into two separate `return` statements, for clarity

app/views/catalog/show.html.erb
thatbudakguy: looks great, thanks!

Issue #3154, opened 2023-05-22
Set Item::Status.availability_class using data available from FOLIO


corylown:
Possibly relates to: https://github.com/sul-dlss/searchworks_traject_indexer/issues/937
Spreadsheet of legacy locations to status codes: https://docs.google.com/spreadsheets/d/1TCWHj45Yb7_7kHst0Cg0Wrk9vGcRX86qYeAqhX1lYvA/edit#gid=0

Or, otherwise figure out what to display about item status/availability with an item's location/status information.

With Sirsi data SearchWorks sets the value of `Holdings::Status.availability_class` to one of:
- `unavailable cdl`
- `in_process`
- `unavailable`
- `noncirc_page`
- `noncirc`
- `page`
- `available`
- `unknown`

The matching case is determined largely by the library/location codes of the item from values in `Settings` or `Constants`. Many of the location codes used to set `Status` in SearchWorks will no longer be used (need to verify if this is true -- the mapping file suggests they are "legacy" codes) in FOLIO and migrated records will map these locations codes to various statuses in FOLIO. The migration mapping is available here: https://github.com/sul-dlss/folio_migration/blob/main/mapping_files/item_statuses.tsv.

Below I've done some analysis for each `Holdings::Status` type defined in SearchWorks to show places where we're relying on codes that will no longer be used in FOLIO, how these codes will be mapped to a FOLIO status, and what we might need to do to set this `Holdings::Status` with FOLIO data.

# `unavailable cdl`

<img width="375" alt="unavailable-cdl" src="https://github.com/sul-dlss/SearchWorks/assets/458247/99b1d914-696a-4b53-bed6-d04823a25bd4">

Item is set to `unavailable cdl` when the home_location is 'CDL'.

We should still be able to use this logic with FOLIO data.

# `in_process`

<img width="375" alt="in-process" src="https://github.com/sul-dlss/SearchWorks/assets/458247/b64cfe05-9979-4164-91f3-7095ebef4340">

Item is set to `in_process` if the library code is 'HV-ARCHIVE'.

We should still be able to use this library code to determine whether something is `in_process`. However, FOLIO also includes a status with the value of `In process` that is mapped to other Sirsi location codes and we need to determine whether items with this FOLIO status should also be set to `in_process` in SearchWorks.

# `unavailable`

[...]

PR #3265, opened 2023-07-13
Check to see if a folio item is pageable?

jcoyne:
Ref #3154


jcoyne:
Abandoning as it seems to have been fixed by #3284

PR #3268, opened 2023-07-14
Rename pageable to deliver_from_offsite

jcoyne:
This is just one subcategory of pageable.


lib/holdings/status.rb
```diff
@@ -24,8 +24,8 @@ def availability_class
         'noncirc_page'
       when noncirc?
         'noncirc'
-      when pageable?
-        'page'
+      when deliver_from_offsite??
```
corylown, line 27:
```suggestion
      when deliver_from_offsite?
```

lib/holdings/status/deliver_from_offsite.rb
```diff
@@ -1,17 +1,17 @@
...
   class Status
-    class Pageable
+    class DeliverFromOffsite
       def initialize(item)
         @item = item
       end
 
-      def pageable?
-        pageable_home_location? || pageable_library?
+      def deliver_from_offsite??
```
corylown, line 8:
```suggestion
      def deliver_from_offsite?
```

jcoyne:
@corylown fixed those.

lib/holdings/status/deliver_from_offsite.rb
```diff
@@ -1,23 +1,23 @@
...
-    class Pageable
+    class DeliverFromOffsite
       def initialize(item)
         @item = item
       end
 
-      def pageable?
-        pageable_home_location? || pageable_library?
+      def deliver_from_offsite?
+        pageable_home_location? || offsite_library?
```
corylown, line 9: Should `pageable_home_location?` and `Constants::PAGE_LOCS` (below) also be renamed? I'd guess something like `offsite_home_location?` given that they also get the truck icon, but I don't know where these locations are.

corylown (changes requested):

lib/holdings/status/deliver_from_offsite.rb
jcoyne: I think that *could* be something we do, but I didn't really understand that part, so I scoped the changes to avoid that part. Since that's a private method, I'm not too concerned about it leading people astray.

jcoyne:
@cbeer good thought. I only see this used in sul-requests and Searchworks. https://github.com/search?q=org%3Asul-dlss%20availability_class&type=code

Here is a PR to accommodate this change: https://github.com/sul-dlss/sul-requests/pull/1611

lib/holdings/status/deliver_from_offsite.rb
jcoyne: @corylown I changed it to `deliverable_home_location?` Does that work for you?

corylown (approved):

PR #3302, opened 2023-07-21
Handle the case where there are no items

### LINKED ISSUE #3301: [SearchWorks/prod] NoMethodError: undefined method `folio_item?' for nil:NilClass
## Backtrace

    line 68 of [PROJECT_ROOT]/app/policies/request_link_policy.rb: folio_items?
    line 21 of [PROJECT_ROOT]/app/policies/request_link_policy.rb: aeon_pageable?
    line 70 of [PROJECT_ROOT]/app/components/location_request_link_component.rb: link_text

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/98014493)

jcoyne:
For prod error. 

Fixes #3301

### DIFF
--- a/spec/policies/request_link_policy_spec.rb
+++ b/spec/policies/request_link_policy_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+RSpec.describe RequestLinkPolicy do
+  subject(:policy) { described_class.new(location:, library:, items:) }
+
+  let(:location) { 'HOOVER' }
+  let(:library) { 'STACKS' }
+
+  describe '#aeon_pageable?' do
+    subject { policy.aeon_pageable? }
+
+    context 'when there are no items' do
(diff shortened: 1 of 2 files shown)


spec/policies/request_link_policy_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  describe '#aeon_pageable?' do
+    subject { policy.aeon_pageable? }
+
+    context 'when there are no items' do
+      let(:items) { [] }
+
+      it { is_expected.to be true }
+    end
+  end
```
thatbudakguy, line 17: out of context, this isn't intuitive to me: something is aeon pageable if there are no items?

jcoyne: I agree, but I think that was the previous behavior.

jcoyne: It seems that @cbeer has confirmed this was a data problem and has worked to resolve it.

PR #3310, opened 2023-07-24
Use Folio discovery labels for locations

### LINKED ISSUE #3271: Use discoveryDisplayLabels from FOLIO
FOLIO locations have a `discoveryDisplayLabel` that is/should be appropriate for us to use instead of `Constants::LOCS`.

Before making this change, we should do some due diligence to make sure the FOLIO labels are generally appropriate and substantially similar to the existing constants (CB: I wouldn't be surprised if we end up taking the FOLIO labels for 90% of the locations, and still having to override some in Searchworks. If we assemble a list of the questionable differences, Sarah et al can review.)

jcoyne:
Stop using Constant::LOCS (except as a fallback for non-migrated locations)

Fixes #3271 

This adds a Honeybadger notification if we don't find the code in the folio locations (and if we don't ignore it e.g. `INPROCESS`, etc.)


corylown (commented):
@jcoyne since this would switch us over to using Folio's display labels in production, I think we should wait until Sarah updates the display labels in Folio (or otherwise modify this so it only applies to Folio sourced records). They're currently not labels we would want to display in production. For example, the current label "US Federal Documents" would display as "Green Library US Federal Government Documents."

jcoyne:
@corylown that makes sense. Sarah says she'll update the labels tonight.

jcoyne:
@corylown this has been rebased with Sarah's label changes pulled in.

app/services/folio/locations.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Folio
+  # This reads with the locations.json file exported from Folio.
+  class Locations
```
thatbudakguy, line 7:
```suggestion
  # This reads the locations.json file exported from Folio.
```

app/services/folio/locations_map.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Folio
+  # This starts with the location.tsv file provided by libsys. We make any modifications to that data here.
+  # This is intended to be a temporary class, which will only be necessary until we complete the Folio migration.
+  class LocationsMap
```
thatbudakguy, line 8:
```suggestion
  # This is intended to be a temporary class, which will only be necessary until we complete the Folio migration.
  # TODO: remove after migration
  class LocationsMap
```
makes it easier to find later?

thatbudakguy (approved):

Issue #3129, opened 2023-05-05
Display the request button in Searchworks using the FOLIO circ rules


cbeer:
Searchworks currently displays two different kinds of request buttons:
- location-level
- item-level

We display location-level request links if the library/location is "pageable" with circulating items (e.g. SAL3), or "mediated" (or "aeon") for library's like SPEC.

We display item-level request links if the item's current location is requestable (e.g. CHECKEDOUT, LOST, etc).

We're hoping [some of the data for determining which request button](https://github.com/sul-dlss/SearchWorks/blob/bc855902d11d1e573a9466b23e7083851bfbee7e/config/settings.yml#L133-L589) to show can be determined by [evaluating FOLIO's circ rules](https://github.com/sul-dlss/SearchWorks/issues/3100); see #3149.

At a high-level, we generally show a location-level request link if a patron can have the item retrieved from off-site or an non-public area (e.g. locked stacks, SPEC) in order to at least see it. Item-level request links are used if a specific item is unavailable when it would be accessible (e.g. in the stacks somewhere) but is temporarily unavailable (e.g. because it is checked out). The actual rules are [even more nuanced](https://docs.google.com/spreadsheets/d/1n1n_DxHI1nDZC2DecjiMeiafu9VqpM2jJcS5l1-azYA/edit#gid=1279865253).

We should show a [location-level request link](https://github.com/sul-dlss/SearchWorks/blob/master/app/components/location_request_link_component.rb) if:

- [x] the effective library/location has a request policy that allows paging for any of the material types/loan types in that location (for any patron group).
- [x] TODO: we're currently filtering by current location to remove location-level links if the current location doesn't allow paging... is this appropriately captured in the FOLIO circ rules?
  - Look in SW code and make sure all locations there are in circ rules.
  - I believe this means `Settings.disabled_current_locations` which means https://github.com/sul-dlss/SearchWorks/blob/master/config/settings.yml#L771-L776
  - Those locations were not translated to folio locations https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/translation_maps/locations.tsv and they look like they are probably item statuses, but I'm not sure which statuses those items would have been given. `SPEC-INPRO` is in the temporary locations.
  - So, no these aren't captured because we need to ensure "SPEC-COLL" items with certain statuses "Missing", "In-Process" are not requestable. 
[...]

PR #3323, opened 2023-07-25
The item request link looks at folio data

jcoyne:
Ref #3129


app/policies/item_request_link_policy.rb
```diff
@@ -5,6 +5,8 @@ def initialize(item:)
   end
 
   def show?
+    return folio_pageable? if item.folio_item?
```
cbeer, line 8: This doesn't match my intuition about when we show item-level request links. I'm not sure it's actually wrong, but whoever reviews it should take this out for a good spin.

I believe we show item-level request links when:
- there isn't a location-level request link (which.. is kinda the opposite of `folio_pageable?`, hence my confusion)
- when we know from the it's current location that the item is pageable without even needing to look at its current availability (because it's in some state that we know isn't temporary but the item is unavailable).

I don't know if the circ rules adequately capture that latter case, and it seems like it'd more likely be a `Hold` request type (because something `Page`-able seems like it's likely pageable at the item level?).. but it's possible it all works out in the end?

cbeer:
Fixed by #3375

PR #3336, opened 2023-07-27
Convert ItemRequestLinkComponent to use Folio locations 

jcoyne:
Ref #3129

### DIFF
--- a/spec/components/item_request_link_component_spec.rb
+++ b/spec/components/item_request_link_component_spec.rb
@@ -9,7 +9,8 @@ RSpec.describe ItemRequestLinkComponent, type: :component do
   let(:library) { 'GREEN' }
   let(:location) { 'LOCKED-STK' }
   let(:item) do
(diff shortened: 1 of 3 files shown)


spec/components/item_request_link_component_spec.rb
```diff
@@ -107,4 +108,44 @@
...
+    let(:policy) { instance_double(ItemRequestLinkPolicy, show?: true) }
+    let(:item) do
+      instance_double(Holdings::Item, document:, library:, effective_location:, home_location: location, barcode: 13_333_333_333_333, folio_item?: true)
+    end
+
+    before do
+      allow(ItemRequestLinkPolicy).to receive(:new).and_return(policy)
+    end
+
+    context 'with an onsite page locaion' do
```
thatbudakguy, line 122:
```suggestion
    context 'with an onsite page location' do
```

PR #3339, opened 2023-07-27
Parse the due date using the application's time zone

cbeer:
FOLIO due dates are stored as UTC.

### DIFF
--- a/app/services/live_lookup/folio.rb
+++ b/app/services/live_lookup/folio.rb
@@ -55,7 +55,7 @@ class LiveLookup
 
     def due_date(item)
       date = item.fetch('dueDate', nil)
-      Date.parse(date).strftime('%m/%d/%Y') if date
+      date&.to_time&.in_time_zone&.strftime('%m/%d/%Y')
     end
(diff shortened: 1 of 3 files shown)


app/services/live_lookup/folio.rb
```diff
@@ -55,7 +55,7 @@ def status(item)
 
     def due_date(item)
       date = item.fetch('dueDate', nil)
-      Date.parse(date).strftime('%m/%d/%Y') if date
+      Time.zone.parse(date).strftime('%m/%d/%Y') if date
```
jcoyne, line 58: The due date already has a time zone, right? I find that this better expresses the intention:
```suggestion
     DateTime.parse(date).in_time_zone.strftime('%m/%d/%Y') if date
```

jcoyne: Or alternately:
```suggestion
      date.to_time.in_time_zone.strftime('%m/%d/%Y') if date
```

jcoyne (approved):

PR #3343, opened 2023-07-27
Treat items with the type 'ONLINE' as internet resources

cbeer:
Lane has item-level resources for e-resources in a variety of locations.

### DIFF
--- a/lib/holdings/item.rb
+++ b/lib/holdings/item.rb
@@ -214,7 +214,7 @@ class Holdings
     private
 
     def internet_resource?
-      home_location == 'INTERNET'
+      home_location == 'INTERNET' || type == 'ONLINE'
     end
 
     def eresv?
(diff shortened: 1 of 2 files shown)


lib/holdings/item.rb
```diff
@@ -214,7 +214,7 @@ def folio_item?
     private
 
     def internet_resource?
-      home_location == 'INTERNET'
+      home_location == 'INTERNET' || type == 'ONLINE'
```
jcoyne, line 217: Can you add the comment here about this being for Lane resources?

jcoyne (changes requested):

lib/holdings/item.rb
cbeer: I'm not sure why that's appropriate; it's not just for Lane resources (although because none of Lane's eresources have the home location INTERNET, it's easy to find examples)

PR #3332, opened 2023-07-27
Refactor Holdings::Item to use a pre-parsed hash for its incoming data

cbeer:

PR #3351, opened 2023-07-28
Handle nil values in the Holding::Item better

### LINKED ISSUE #3350: [SearchWorks/morison] NoMethodError: undefined method `end_with?' for nil:NilClass
## Backtrace

    line 42 of [PROJECT_ROOT]/app/policies/item_request_link_policy.rb: current_location_is_always_requestable?
    line 10 of [PROJECT_ROOT]/app/policies/item_request_link_policy.rb: show?
    line 22 of [PROJECT_ROOT]/app/components/item_request_link_component.rb: render?

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/98618870)

cbeer:
Regression from #3332; nil values used to be impossible for (most) item fields.

Fixes #3350

### DIFF
--- a/app/components/access_panels/library_location_component.html.erb
+++ b/app/components/access_panels/library_location_component.html.erb
@@ -42,7 +42,7 @@
             <span class="availability-icon-wrapper">
               <i class="availability-icon <%= item.status.availability_class %>"></i>
               <span data-available-text="<%= t('searchworks.availability.available') %>" data-unavailable-text="<%= t('searchworks.availability.unavailable') %>" class='status-text'>
-                <%= item.status.status_text unless item.current_location && item.current_location.name.present? %>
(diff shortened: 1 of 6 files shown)


jcoyne (changes requested):
So current_location can be nil?  When does that happen? Can we add some tests to illustrate this scenario?

cbeer:
Any of the item_display fields can be nil, if they're derived from `item_display_struct`. I think [the change](https://github.com/sul-dlss/SearchWorks/pull/3351/files#diff-9667b67ea2894add0e62bbb7d311ee07ce7d972b03e551c88ee2a5c7e1fcccfcR33) to `.from_item_display_string` means our whole test suite is testing that now.

thatbudakguy (approved):

PR #3362, opened 2023-07-31
Remove support for item_display field, use item_display_struct exclusively

cbeer:
The full dump from this weekend populated this field for Symphony.

TODO:
- [x] full reindex of searchworks-folio-dev
- [x] full reindex of searchworks-morison (and other non-prod Symphony collections?)


corylown:
Needs rebasing

corylown (approved):

PR #3368, opened 2023-08-01
Update _top_navbar.html.erb link

cbeer:


jcoyne:
I believe this is a duplicate of #3365

PR #3384, opened 2023-08-04
Use Settings.LIVE_LOOKUP_SERVICE to select LiveLookup service class

corylown:
Prepares us to have more than two LiveLookup services in case we need to use Solr as a backup service while Symphony is shut down. Should get merged around the same time as: https://github.com/sul-dlss/shared_configs/pull/2026

### DIFF
--- a/app/services/live_lookup.rb
+++ b/app/services/live_lookup.rb
@@ -1,5 +1,5 @@
 # LiveLookup serves as a wrapper around specific ILS implementations
-#   for requesting live information about item availability.
+# for requesting live information about item availability.
 class LiveLookup
   delegate :as_json, :to_json, to: :records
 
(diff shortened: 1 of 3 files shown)


app/services/live_lookup.rb
```diff
@@ -1,19 +1,21 @@
...
-    else
-      LiveLookup::Sirsi.new(@ids).records
-    end
+    live_lookup_service.new(@ids).records
+  end
+
+  private
+
+  def live_lookup_service
+    (Settings.LIVE_LOOKUP_SERVICE || 'LiveLookup::Sirsi').constantize
```
jcoyne, line 19: Can this be downcase?  This makes it look like a constant, but it's actually a method.

jcoyne (changes requested):

app/services/live_lookup.rb
jcoyne: Also, can you put `live_lookup_service: LiveLookup::Sirsi` in `config/settings.yml` rather than a default here?

corylown: Sounds good. Updated with both suggestions.

PR #3395, opened 2023-08-07
Allow global alert paragraphs to fill the whole alert section

cbeer:
Screenshot 2023-08-07 at 10 19 55


jcoyne (approved):
Looks good. This appears to override https://github.com/sul-dlss/SearchWorks/blob/be76d2d1c14370d4974586aeb16997d3a6783552/app/assets/stylesheets/base.scss#L64 Do you know why we need that in the first place?

PR #3396, opened 2023-08-07
Add LiveLookup::Solr service for use when ILS service is unavailable

### LINKED ISSUE #3248: Disable real-time availability lookups during Symphony shutoff
We're planning to turn off symphony web services during the FOLIO cutover to prevent changes to the underlying ILS data. During that time, real-time availability lookups won't work.

In order to prevent patrons from seeing lots of "unknown" statuses for items in SearchWorks, we want to just use the existing MARC 999 status data and not overlay it with our real-time lookup.

For some items, the status will still show as "unknown" — this is expected for items in locations like "ask at X", "see other", etc.

corylown:
Closes #3248

### DIFF
--- a/app/services/live_lookup/solr.rb
+++ b/app/services/live_lookup/solr.rb
@@ -18,7 +18,7 @@ class LiveLookup
     def records
       @records ||= response.dig('response', 'docs').flat_map do |doc|
         doc.fetch('item_display_struct', []).map do |item|
-          LiveLookup::Solr::Record.new(item).as_json
+          LiveLookup::Solr::Record.new(JSON.parse(item)).as_json
         end
       end
     end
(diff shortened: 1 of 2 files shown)


app/services/live_lookup/solr.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      def status
+        return unless valid_current_location?
+
+        current_location.name
+      end
+
+      def item_as_json
+        JSON.parse(item)
```
cbeer, line 85: Instead of repeatedly parsing the string, should we push this up to the constructor (or higher? `LiveLookup::Solr::Record.new(JSON.parse(item)).as_json`)

PR #3421, opened 2023-08-15
Use preferred item for browse nearby if instance has no preferred barcode

### LINKED ISSUE #3405: Browse nearby for items without callnumbers shouldn't be broken.
E.g. https://searchworks-folio-dev.stanford.edu/view/in00000066916

Image

I'm not sure whether we're indexing these wrong and failing to suppress it, or if there's a check we're missing.

marlo-longley:
Fixes #3405

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -48,8 +48,12 @@ class BrowseController < ApplicationController
 
       @items = service.items(params[:before] || params[:after])
     else
-      barcode = params[:barcode] || @original_doc[:preferred_barcode]
-      item = @original_doc.items.find { |c| c.barcode&.starts_with?(barcode) }
+
+      item = if params[:barcode]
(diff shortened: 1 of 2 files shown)


cbeer:
I guess I'm unclear why `Item#browsable?` is true for these records. It feels like we did something wrong... somewhere... along the way.

thatbudakguy (approved):

thatbudakguy:
Hmm. Yeah, it seems...not great to have a nonexistent call number be typed as "LC"

cbeer (changes requested):
That's (presumably?) just the data though. I don't think we want to go second-guessing it too much. The mystery remains, why do we have all the pieces that `#browseable?` thinks we need, but we end up with nothing to browse?

It seems to me that either `#browseable?` is missing a condition, or we've failed to inject the right data in our FOLIO mappings. I think we should look into the underlying issue instead of patching over it in `#callnumber_type`.

cbeer (approved):

PR #3431, opened 2023-08-18
Extract location item component

cbeer:

### DIFF
--- a/app/components/access_panels/location_item_component.html.erb
+++ b/app/components/access_panels/location_item_component.html.erb
@@ -1,0 +1,35 @@
+<%= tag.tr class: classes, data: { barcode: item.barcode } do %>
+  <td class="callnumber">
+    <%= item.callnumber %>
+  </td>
+  <% if render_item_note? %>
+    <td class="public-note">
+      <% if item.public_note.present? %>
+        <div class='public-note note-highlight'>Note: <%= inject_line_break_opportunities item.public_note %></div>
+      <% end %>
+    </td>
+  <% end %>
+  <% if render_item_details? %>
+    <td class="item-availability" data-live-lookup-id="<%= document.live_lookup_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
+      <span class="availability-icon-wrapper">
+        <i class="availability-icon <%= item.status.availability_class %>"></i>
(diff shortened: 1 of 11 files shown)


app/components/access_panels/location_item_component.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def render_real_time_availability_request_link?
+      # we're not rendering item-level request links (because e.g. there's already alocation level request link)
+      return false unless render_item_level_request_link?
+
+      # non-circulating items don't need real time availability
+      return false unless item.circulates?
+
+      # items that definitely have at render time don't need real time availability
+      return false if item.request_link.render?
```
thatbudakguy, line 48: definitely have...?

thatbudakguy (approved):

app/components/access_panels/location_item_component.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def render_real_time_availability_request_link?
+      # we're not rendering item-level request links (because e.g. there's already alocation level request link)
+      return false unless render_item_level_request_link?
+
+      # non-circulating items don't need real time availability
+      return false unless item.circulates?
+
+      # items that definitely have at render time don't need real time availability
```
cbeer, line 47:
```suggestion
      # items that definitely have an item-level request link at render time don't need real time availability
```

app/components/access_panels/location_item_component.html.erb
```diff
@@ -0,0 +1,35 @@
...
+        <%= item.loan_period %>
+      <% end %>
+      <% if render_item_level_request_link? %>
+        <span class="request-link">
+          <%= render item.request_link %>
+        </span>
+      <% end %>
+    </td>
+  <% else %>
+    <td></td>
```
cbeer, line 33: I hope this is the only markup difference between index/show pages; we currently show availability for bound-withs on the index page but not the show page (weird), but Sarah's ok with consistency.

app/components/access_panels/location_item_component.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= tag.tr class: classes, data: { barcode: item.barcode } do %>
+  <td class="callnumber">
```
cbeer, line 2: This PR adds css classes to the `td` elements on the index page, but they appear to be unused on the index pages. The styles seem to come from the `availability-item` class applied exclusively by the show page.

app/components/access_panels/location_item_component.html.erb
thatbudakguy: Agree that that's weird (if anything seems like it'd be the reverse) but +1 for consistency

PR #3491, opened 2023-09-12
Remove LANE-MED location links

cbeer:
Fixes VUF-9327

### DIFF
--- a/lib/holdings/location.rb
+++ b/lib/holdings/location.rb
@@ -26,16 +26,8 @@ class Holdings
       @code && Constants::BOUND_WITH_LOCS.include?(@code)
     end
 
-    def location_link
-      return unless library == 'LANE-MED'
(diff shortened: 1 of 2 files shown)


lib/holdings/location.rb
```diff
@@ -26,16 +26,8 @@ def bound_with?
...
-      return if items.any?(&:folio_item?)
-
-      if items.first.try(:barcode)
-        "https://lane.stanford.edu/view/bib/#{items.first.barcode.gsub(/^LL/, '')}"
-      else
-        'https://lane.stanford.edu'
-      end
-    end
+    # Intentionally left blank
+    def location_link; end
```
jcoyne, line 30: Aren't there tests covering this method that need to be updated?

cbeer: I hope so 🤷‍♂️

Issue #3494, opened 2023-09-13
Fix "unknown" availability status for Bound-with children


marlo-longley:
We want the parent's availability to show up for the BW-child.
In Symphony (https://searchworks-symphony.stanford.edu) for example these items show as available while in FOLIO/production they do not.
Availability status is based on catkey and BW children don’t have catkeys in FOLIO.

a36242 — availability differences for green and classics libraries
a87219 — unknown availability
a78957 — unknown availability
a136521 — unknown availability
a80032  — unknown availability
a133575 — availability is correct but there is no request button unlike in symphony
a105784 — unknown availability
a76911 — unknown availability
a87777 — unknown availability for 1 item

PR #3503, opened 2023-09-15
Display parent availability for FOLIO bound-with children

marlo-longley:
Closes https://github.com/sul-dlss/SearchWorks/issues/3494

### DIFF
--- a/app/components/access_panels/location_item_component.html.erb
+++ b/app/components/access_panels/location_item_component.html.erb
@@ -10,7 +10,7 @@
     </td>
   <% end %>
   <% if render_item_details? %>
-    <td class="item-availability" data-live-lookup-id="<%= document.live_lookup_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
+    <td class="item-availability" data-live-lookup-id="<%= item.live_lookup_instance_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
       <span class="availability-icon-wrapper">
         <i class="availability-icon <%= item.status.availability_class %>"></i>
         <span data-available-text="<%= t('searchworks.availability.available') %>" data-unavailable-text="<%= t('searchworks.availability.unavailable') %>" class='status-text'>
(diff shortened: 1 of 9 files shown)


app/components/access_panels/location_item_component.html.erb
```diff
@@ -10,7 +10,7 @@
     </td>
   <% end %>
   <% if render_item_details? %>
-    <td class="item-availability" data-live-lookup-id="<%= document.live_lookup_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
+    <td class="item-availability" data-live-lookup-id="<%= item.live_lookup_instance_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id if item.live_status? %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
```
jcoyne, line 13: Do we also need to change this line? https://github.com/sul-dlss/SearchWorks/blob/master/app/components/access_panels/at_the_library_component.html.erb#L2 ?

cbeer: I don't think so (although maybe it's redundant? Or we should be omitting item-level attributes if it's the same as that?)

lib/holdings/item.rb
```diff
@@ -134,8 +133,17 @@ def stackmapable?
       stackmapable_library? && stackmapable_location?
     end
 
+    def bound_with_parent
+      return nil unless document && document['holdings_json_struct']
+
+      match = document['holdings_json_struct'].first['holdings']&.find do |item|
+        item.dig('boundWith', 'item', 'barcode') == barcode
```
cbeer, line 140:
```suggestion
        item.dig('boundWith', 'item', 'id') == id
```

I think we've seen items without barcodes or duplicated barcodes (😬). `id` ought to always be available though.

lib/holdings/item.rb
```diff
@@ -134,8 +133,17 @@ def stackmapable?
       stackmapable_library? && stackmapable_location?
     end
 
+    def bound_with_parent
+      return nil unless document && document['holdings_json_struct']
```
cbeer, line 137:
```suggestion
      return nil unless document && document['holdings_json_struct'].any?
```

I think an empty array would also evaluate to `true` and break with a NoMethod error below, right?

lib/holdings/item.rb
```diff
@@ -134,8 +133,17 @@ def stackmapable?
...
+
+      match = document['holdings_json_struct'].first['holdings']&.find do |item|
```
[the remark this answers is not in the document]
marlo-longley: @cbeer not sure I did what you meant, but I moved some login into `folio_holdings`

PR #3521, opened 2023-09-19
Remove the xs only navigation

### LINKED ISSUE #3509: Duplicated element IDs in mobile version of search results header
We render two separate versions of this header and its controls — one shown for small & up screens, and one just for xs (mobile) screens. Since a lot of the markup is duplicated, the IDs ended up getting duplicated too.

Screenshot 2023-09-15 at 15 34 02

The IDs that are duplicated are basically everything except the "next" button:
- `<div id="search-results-toolbar">` (the containing bar itself)
- `<div id="view-type-dropdown">` (the "view" dropdown)
- `<div id="sort-dropdown">` (the sort dropdown)
- `<div id="per_page-dropdown">` (the "per page" dropdown)
- `<button id="select_all-dropdown">` (the "select all" checkbox...oddly named?)

We need to make the IDs for either the mobile version or the non-mobile version different so that they're unique. We probably need to do at least a cursory search to make sure nothing else (e.g. javascript) is looking for those ID values.

Having duplicated IDs on a page is a [WCAG Level A violation](https://www.w3.org/TR/WCAG20-TECHS/F77.html), which is something SODA has asked us to remediate whenever time allows.

[See this issue in SiteImprove](https://my2.siteimprove.com/Accessibility/1059585/NextGen/Issue/1?ruleId=3&pageSegments=&issueKind=1&tagFilterType=2&conformance=&siteTargetIssueKinds=1,2&wcagVersion=21&siteTarget.wcagVersion=21&siteTarget.conformanceLevels=0,1,3,4&siteTarget.issueKinds=1,2)

jcoyne:
Because this has duplicate ids which is an accessability problem.  We don't expect to serve devices that can't handle small navigation.

Fixes #3509

### DIFF
--- a/app/views/catalog/_sort_and_per_page.html.erb
+++ b/app/views/catalog/_sort_and_per_page.html.erb
@@ -1,6 +1,6 @@
 <% ivar_to_paginate =  article_selections? ? @bookmarks : @response %>
 <!-- Markup for sm screens and larger -->
-<div class="sort-and-per-page sul-toolbar selection-list-toolbar hidden-xs">
+<div class="sort-and-per-page sul-toolbar selection-list-toolbar">
   <div id="search-results-toolbar" class="search-widgets">
     <%= render partial: "paginate_compact", object: ivar_to_paginate %>
     <% if bookmarks? %>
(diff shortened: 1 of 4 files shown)


thatbudakguy:
lol, it appears we were well aware of the duplicate IDs:
https://github.com/sul-dlss/SearchWorks/blob/ee67df16b4ed3eddb6016b3a569445380ede5fbe/spec/features/access_points/databases_spec.rb#L46

i think this'll pass once the three instances of that in tests are fixed. gotta love it

thatbudakguy (approved):

PR #3522, opened 2023-09-20
Update FOLIO types

sul-devops-team:


jcoyne:
@cbeer is this an example of one of the "weird" deltas we're trying to prevent?  Any idea where these extra carriage returns come from?

cbeer:
This looks fine to me? Staff edit the plain text directly and probably added new lines for clarity.

Issue #3470, opened 2023-09-01
Use location/details to link to stackmaps


thatbudakguy:
see e.g. in the config: https://github.com/sul-dlss/SearchWorks/blob/55c521d5dba09b1e1a19e44b59603f26748dd6f4/config/settings.yml#L121

instead of keeping this around, we can just define on the library/location where you go to get its stackmap.

note that LAW is soon to remove stackmaps entirely (2weeks?)

See https://github.com/sul-dlss/folio-graphql/pull/128

marlo-longley:
Most of the code is written in #3528 but this is blocked until we can get the location details for `stackmapBaseUrl` out of Folio; requires communication on the product side.

thatbudakguy:
potentially unblocked now since data has been added to location details per @saseestone ?

PR #3528, opened 2023-09-20
Get stackmap API url from FOLIO Location details

marlo-longley:
Closes https://github.com/sul-dlss/SearchWorks/issues/3470

This work should not be merged until we confirm this data has been populated in FOLIO. Currently this has not been done. For the changed `location_spec` test to pass, we'd need the populated data in `locations.json` here, but everything passes testing with locally modified data.

This PR assumes that a `stackmapBaseUrl` will be present in location details if the location is stackmapable. For example:
```json
    "details": {
      "stackmapBaseUrl": "https://stanford.stackmap.com/json/"
    }
```

If this assumption is correct it removes the need for lists of `ignored_stackmap_locations` per library, as this information is now stored on the location itself. But I'm happy to revise if this isn't the case.


app/views/catalog/stackmap.html.erb
```diff
@@ -7,7 +7,7 @@
   <h3 class="modal-title"><%= params[:title] %></h3>
 </div>
 
-<div class="stackmap col-md-12" data-behavior="stackmap" data-callnumber="<%= params[:callnumber] %>"  data-library="<%= params[:library] %>" data-location="<%= params[:location] %>" data-stackmap-url="<%= stackmap_api_url(params[:library]) %>">
+<div class="stackmap col-md-12" data-behavior="stackmap" data-callnumber="<%= params[:callnumber] %>"  data-library="<%= params[:library] %>" data-location="<%= params[:location] %>" data-stackmap-url="<%= params[:api_url] %>">
```
jcoyne, line 10: I don't see where `api_url` gets in the params. Can you point that out for me?

app/helpers/catalog_helper.rb
```diff
@@ -22,7 +22,7 @@ def current_view
 
   def stackmap_link(document, location)
     item = location.items.first
-    stackmap_path(title: (document['title_display'] || '').html_safe, id: document.id, callnumber: item.callnumber, library: item.library, location: item.home_location)
+    stackmap_path(title: (document['title_display'] || '').html_safe, id: document.id, callnumber: item.callnumber, library: item.library, location: item.home_location, api_url: location.stackmap_api_url)
```
marlo-longley, line 25: @jcoyne here is where I added the param

thatbudakguy (approved):

jcoyne (approved):

PR #3536, opened 2023-09-22
Hide fake cover when Google Books cover is present

marlo-longley:
Related to https://github.com/sul-dlss/SearchWorks/commit/e5a10ba85b9471f4fd55461eb87b8f92dc938c7e


corylown (commented):
Thanks @marlo-longley -- this does fix the gallery view, but it looks like the cover images are still missing from the list view.
Screenshot 2023-09-22 at 8 34 11 PM

marlo-longley:
I didn't realize that was an issue -- thanks for catching and will address!

PR #3551, opened 2023-09-26
Fix collection membership display

thatbudakguy:
- Strip leading 'a' from catkeys when finding parent collections
- Make digital collection preview search use a-prefixed catkeys
- Permit some params for CollectionMembersController to silence warnings

### DIFF
--- a/app/helpers/collection_helper.rb
+++ b/app/helpers/collection_helper.rb
@@ -1,10 +1,10 @@
 module CollectionHelper
   def link_to_collection_members(link_text, document, options = {})
-    link_to(link_text, search_catalog_path(f: { collection: [document[:id]] }), options)
+    link_to(link_text, collection_members_path(document), options)
   end
 
   def collection_members_path(document, options = {})
-    search_catalog_path(f: { collection: [document[:id]] })
+    search_catalog_path(f: { collection: [document.prefixed_id] })
   end
 
   def collection_members_enumeration(document)
--- a/app/helpers/collection_helper.rb
+++ b/app/helpers/collection_helper.rb
@@ -27,7 +27,8 @@ module CollectionHelper
   def collection_breadcrumb_value(collection_id)
     if @response.documents.first&.index_parent_collections.present?
       collection = @response.documents.first.index_parent_collections.find do |coll|
-        coll[:id] == collection_id
+        # strip leading 'a' from collection catkeys
+        coll[:id] == collection_id.sub(/^a(\d+)$/, '\1')
       end
       return document_presenter(collection).heading if collection.present?
     end
--- a/app/models/concerns/collection_member.rb
+++ b/app/models/concerns/collection_member.rb
@@ -19,7 +19,7 @@ module CollectionMember
     unless @index_parent_collections
       @index_parent_collections = self[:collection_with_title].map do |collection_with_title|
         id, title = collection_with_title.split('-|-').map(&:strip)
-        SolrDocument.new(id:, title_display: title)
+        SolrDocument.new(id: strip_a(id), title_display: title)
       end
     end
     @index_parent_collections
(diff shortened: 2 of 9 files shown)


app/models/concerns/collection_member.rb
```diff
@@ -36,7 +36,7 @@ def online_label
 
   def parent_collection_params
     self["collection"].map do |collection_id|
-      "id:#{collection_id}"
+      "id:#{collection_id.gsub(/a?(\d+)/, '\1')}"
```
corylown, line 39: Should we left & right anchor this regex?
```suggestion
      "id:#{collection_id.sub(/^a(\d+)$/, '\1')}"
```

app/models/concerns/digital_collection.rb
```diff
@@ -96,5 +96,18 @@ def number_of_documents_without_image_urls
...
+    # search for items with collection set to both prefixed and non-prefixed ids
+    def collection_member_params
+      [document[:id], prefixed_document_id].uniq.map do |collection_id|
+        "collection:\"#{collection_id}\""
+      end.join(" OR ")
+    end
+
+    # @return [String] the document id, prefixed with an 'a' if it's a numeric catkey
+    def prefixed_document_id
+      @document[:id].gsub(/(\d+)/, 'a\1')
```
corylown, line 110:
```suggestion
      @document[:id].sub(/^(\d+)$/, 'a\1')
```

corylown (changes requested):
@thatbudakguy I think some changes are needed in `CollectionHelper` as well where we're generating queries on the collection field using the document id. We'll need to prefix the document id with `a` in these cases as well so the queries work. Without this change the "Explore this collection (437 digital items)" link will produce a search with no results.
Screenshot 2023-09-27 at 11 00 09 AM

thatbudakguy:
thanks for catching that. I moved the prefixed ID generation to a method directly on `SolrDocument` so that it was easier to access in the helper (and wherever else we may need it later).

corylown:
Thanks, @thatbudakguy. I think I found another place where the id needs adjustment for things to link correctly.

In CollectionMember#index_parent_collections we're plucking the collection id out of `:collection_with_title` where it will have the `a` prefix. This id is then used to link to the digital collection in the index view:
Screenshot 2023-09-27 at 3 51 52 PM

The link with the `a` prefix will work, but it seems preferable not to publish these and to prefer the link without the a.

If we change this behavior then we also need to add the `a` back when we compare the collection id in `CollectionHelper#collection_breadcrumb_value`. Happy to hop on a huddle to explain.

app/helpers/collection_helper.rb
```diff
@@ -27,7 +27,8 @@ def collections_search_params
   def collection_breadcrumb_value(collection_id)
     if @response.documents.first&.index_parent_collections.present?
       collection = @response.documents.first.index_parent_collections.find do |coll|
-        coll[:id] == collection_id
+        # re-add leading 'a' to collection catkeys since that's how we index it
+        coll[:id] == collection_id.sub(/^(\d+)$/, 'a\1')
```
corylown, line 31: My fault. I think I got this backwards. And should get the test passing.

```suggestion
        # removing leading 'a' from the collection_id passed to the helper because
        # we remove it from the ids in index_parent_collections
        coll[:id] == collection_id.sub(/^a(\d+)$/, '\1')
```

corylown (changes requested):

thatbudakguy:
I got really lost trying to figure out the right behavior and when to change tests...hope this is all right.

corylown (changes requested):
@thatbudakguy I think the linking and display are correct now. A couple of suggestions for test descriptions. Once those are resolved we can ship it. Thanks for working on this.

thatbudakguy:
@corylown ok, fixed the test descriptions and rebased

corylown (approved):
🎉

PR #3555, opened 2023-09-27
Convert feedback form to use fetch

jcoyne:
This helps to move us off Jquery 1.x, so that we can upgrade to Bootstrap 4.

### DIFF
--- a/app/assets/javascripts/feedback_form.js
+++ b/app/assets/javascripts/feedback_form.js
@@ -43,18 +43,17 @@ Blacklight.onLoad(function(){
           $thisform.find('#last_search').val(lastSearch);
           $thisform.on('submit', function(e){
             e.preventDefault();
-            var valuesToSubmit = $(this).serialize();
-            $.ajax({
-              url: form.action,
-              data: valuesToSubmit,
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/feedback_form.js
```diff
@@ -44,17 +44,17 @@ Blacklight.onLoad(function(){
...
-                $($el).collapse('hide');
-                $($form)[0].reset();
-              }
-              renderFlashMessages(response);
-            });
+            fetch(form.action, {
+              method: 'POST',
+              headers: { "Content-Type": "application/x-www-form-urlencoded" },
+              body: valuesToSubmit
+            }).then((resp) => resp.json())
```
thatbudakguy, line 51: I think if you make `body` a `URLSearchParams` it'll automatically send the content-type as `x-www-form-urlencoded`. That might let you skip using jQuery's `serialize()` too; I think it would just be `new URLSearchParams(new FormData($thisform[0]))`

thatbudakguy (approved):

thatbudakguy (approved):

PR #3597, opened 2023-10-03
Remove unnecessary check

jcoyne:
Floating-point equality is defined such that NaN is never equal to anything, so "x === NaN" always returns false.

### DIFF
--- a/app/assets/javascripts/jquery.preview-gallery.js
+++ b/app/assets/javascripts/jquery.preview-gallery.js
@@ -115,7 +115,7 @@
 
       function itemsPerRow() {
         var width = $('#documents').width();
-        return  Math.floor(width/$itemWidth);
+        return Math.floor(width/$itemWidth);
       }
 
       function reorderPreviewDivs() {
@@ -123,12 +123,12 @@
         var previewIndex = previewDivs.index($previewTarget) + 1;
         $itemsPerRow = itemsPerRow();
         /* 
-        / If $itemsPerRow is NaN or 0 we should return here. If not we are going 
+        / If $itemsPerRow is 0 we should return here. If not we are going 
         / to have a bad time with an infinite while loop. This only manifests
         / on the show page when using the "back" button to get back to a show
         / page using the browse nearby feature.
         */
-        if ($itemsPerRow === NaN || $itemsPerRow === 0) {
+        if ($itemsPerRow === 0) {
           return;
         }
 


app/assets/javascripts/jquery.preview-gallery.js
```diff
@@ -115,20 +115,20 @@
...
         $itemsPerRow = itemsPerRow();
         /* 
-        / If $itemsPerRow is NaN or 0 we should return here. If not we are going 
+        / If $itemsPerRow is 0 we should return here. If not we are going 
         / to have a bad time with an infinite while loop. This only manifests
         / on the show page when using the "back" button to get back to a show
         / page using the browse nearby feature.
         */
-        if ($itemsPerRow === NaN || $itemsPerRow === 0) {
+        if ($itemsPerRow === 0) {
```
cbeer, line 131:
```suggestion
        if (Number.isNaN($itemsPerRow) || $itemsPerRow === 0) {
```

It looks like Jack was pretty sure NaN was also a problem in https://github.com/sul-dlss/SearchWorks/pull/1863. Have you confirmed that it isn't a problem, or would it be better to keep checking NaN the right way?

thatbudakguy: `Number.isNaN` is the right way to do this, and imo there's no reason not to do it – let's just make the check actually work.

jcoyne: Right, but it's never NaN. It's always zero in this case.  You can try this by going to https://searchworks.stanford.edu/view/14639881 then click on "[The 2023-24 budget. Multiyear assessment](https://searchworks.stanford.edu/view/14640023)" and then click the back button. The result is always `0`

thatbudakguy: That seems hard to prove conclusively...the value relies on a bunch of jQuery `.width()` calls, which can return NaN if the DOM isn't setup right, and maybe could also end up dividing by zero? Idk why we wouldn't just trust Jack since there's no consequence to doing so.

jcoyne: I assert that maintaining code that isn't necessary is an ongoing cost. So saying "Just do this" is not without consequences.

jcoyne: a divide by zero would result in `Infinity` which is not `NaN`:
```
Number.isNaN(1/0)
false 
```

An invalid DOM would return `null` which is not `NaN`:
```
Number.isNaN($('.fakething').width())
false 
```

thatbudakguy: doing math with `null` creates `NaN`.

```js
$('.fakething').width() / $('.fakething2').width()
NaN
```

but if you really want to remove this, I guess I can live with assuming this'll never happen.

Issue #3613, opened 2023-10-10
[SearchWorks/prod] NoMethodError: undefined method `hours_api_url' for nil:NilClass


cbeer:
## Backtrace

    line 11 of [PROJECT_ROOT]/lib/hours_request.rb: find_library
    line 15 of [PROJECT_ROOT]/lib/hours_request.rb: get
    line 3 of [PROJECT_ROOT]/app/controllers/hours_controller.rb: show

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/96773507)

marlo-longley:
Closed by #3615

PR #3615, opened 2023-10-12
Support folio library codes

cbeer:
This PR adds FOLIO library codes in places we're using Symphony codes. This should allow us to change the indexer to populate the `item_display` `library` field with the FOLIO code instead.

This also removes tests for the Symphony library code `HV-ARCHIVE` which does not exist in the FOLIO data (because it is mapped to `HOOVER` instead)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -34,6 +34,7 @@ Metrics/BlockLength:
 
 Metrics/ModuleLength:
   Exclude:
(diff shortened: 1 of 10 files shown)


thatbudakguy (approved):

marlo-longley:
This should close https://github.com/sul-dlss/SearchWorks/issues/3613 by adding `MEDIA-CENTER`

PR #3647, opened 2023-10-18
Map location codes to Symphony codes for stackmaps

cbeer:
At least until stackmaps is ready to handle FOLIO codes.


jcoyne (approved):

jcoyne (commented):
Do we need to add any tests for this?

cbeer:
I hope not. All this can be removed just as soon as stackmaps switches things over to use FOLIO codes.

PR #3653, opened 2023-10-20
Use the status from the item_display field to determine on-order-ness

cbeer:

### DIFF
--- a/lib/holdings/item.rb
+++ b/lib/holdings/item.rb
@@ -9,7 +9,6 @@ class Holdings
     attr_accessor :due_date
 
     delegate :loan_type, :material_type, :effective_location, :permanent_location, to: :folio_item, allow_nil: true
-    delegate :status, to: :folio_item, prefix: :folio, allow_nil: true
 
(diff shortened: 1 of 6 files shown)


lib/holdings/item.rb
```diff
@@ -34,7 +33,8 @@ def browsable?
     end
 
     def on_order?
-      current_location.code == 'ON-ORDER'
+      folio_status == 'On order' ||
+        current_location.code == 'ON-ORDER'
```
jcoyne, line 37: Is there a reason we still need to support the location code?

cbeer: At least until we do a full re-index with https://github.com/sul-dlss/searchworks_traject_indexer/pull/1275, on-order items without a FOLIO item don't have the status yet.

jcoyne: Are we going to be able to remember that? Should we drop a comment here or create a new ticket that points to that?

jcoyne (approved):
Looks good. Feel free to merge after considering a comment or ticket for tracking the removal of the old method.

PR #3672, opened 2023-11-03
Encode html &#13 as \r for display

### LINKED ISSUE #3671: Collection description not properly formatting in catalog view
When viewing a collection catalog list in searchworks that includes newlines and links in the description, the formatting is not displayed properly, see: https://searchworks.stanford.edu/catalog?f%5Bcollection%5D%5B%5D=sn695pn3494

Which presents as:
Screenshot 2023-11-03 at 8 28 41 AM

Whereas the object view (https://searchworks.stanford.edu/view/sn695pn3494) is formatting properly:

Screenshot 2023-11-03 at 8 30 32 AM

Related work in H2/Argo: https://github.com/sul-dlss/happy-heron/issues/3380

aaron-collier:
Closes #3671 

This may not be the best or only place to fix this display issue. In both H2 and Argo we've made sure that when `\r\n` is included they are represented properly and do not use the `&#13` html encoding by reverting it back to `\r`. This seems to still be converted to `&#13` when reading the solr document for display in searchworks. 

Happy for any suggestions on how else to approach this.

### DIFF
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -34,4 +34,11 @@ module RecordHelper
       end
     end.join('').html_safe
   end
+
+  ##
+  # Expose format_mods_html from mods_display to properly encode content as needed
+  # RecordHelper behavior
+  def format_record_html(...)
+    format_mods_html(...)
+  end
 end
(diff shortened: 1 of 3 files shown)


cbeer:
What about using format_mods_html (https://github.com/sul-dlss/mods_display/blob/116daefe1d095ee0db4354c0fde46ea2518707fb/app/helpers/mods_display/record_helper.rb#L83)?

aaron-collier:
@cbeer thank you, I was not aware of that function. I'm not seeing any examples of how it may be used in this context - if you have any pointers, that would be great. Thank you.

corylown (approved):

PR #3673, opened 2023-11-03
Remove Folio::LocationsMap.for now that the indexed locations use the…

cbeer:
… FOLIO codes

### DIFF
--- a/app/services/folio/locations_map.rb
+++ b/app/services/folio/locations_map.rb
@@ -9,23 +9,14 @@ module Folio
   class LocationsMap
     include Singleton
 
-    # @param [String] library_code the Symphony library
-    # @param [String] location_code the Symphony location
-    # @return [String] the Folio location code
-    def self.for(library_code:, location_code:)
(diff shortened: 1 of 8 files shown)


app/services/folio/locations_map.rb
```diff
@@ -9,23 +9,14 @@ module Folio
...
-    # @param [String] library_code the Symphony library
-    # @param [String] location_code the Symphony location
-    # @return [String] the Folio location code
-    def self.for(library_code:, location_code:)
-      instance.data.dig(library_code, location_code) || location_code
-    end
-
+    # @param [String] location_code the FOLIO location
+    # @return [String] the Symphony location code
     def self.symphony_code_for(location_code:)
```
jcoyne, line 14: Can't we rip this method out too?

jcoyne: Oh, it looks like it's still used by stackmap_link

jcoyne (approved):

PR #3753, opened 2024-01-18
update record helper to include correct class for mods gem

### LINKED ISSUE #3699: Collection description should display with line breaks if they are present
We previously fixed formatting issues with encodings such as `&#13;` displaying in the collection description, but the fix did not address displaying the line breaks. See https://github.com/sul-dlss/SearchWorks/issues/3671 and https://github.com/sul-dlss/SearchWorks/pull/3672

We would like the line breaks to display.

For an example, see: https://searchworks.stanford.edu/catalog?f%5Bcollection%5D%5B%5D=sn695pn3494

The collection description currently displays like:
Screenshot 2023-11-29 at 10 17 50 AM

We would like it to display more like it does on the full record view with line breaks (https://searchworks.stanford.edu/view/sn695pn3494):

Screenshot 2023-11-03 at 8 30 32 AM

dnoneill:
closes #3699

### DIFF
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -38,7 +38,7 @@ module RecordHelper
   ##
   # Expose format_mods_html from mods_display to properly encode content as needed
   # RecordHelper behavior
-  def format_record_html(...)
-    format_mods_html(...)
+  def format_record_html(record_content)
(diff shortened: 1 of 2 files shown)


marlo-longley:
Looking good @dnoneill. One suggestion is to combine these commits into a history that will read clearly in the main SearchWorks branch after it's merged.

jcoyne (approved):

PR #3762, opened 2024-01-19
Patch Blacklight.modal.receiveAjax to execute recaptcha

jcoyne:

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -52,6 +52,39 @@ import "./tooltip";
 import "./truncate";
 import "./update-hidden-inputs-by-checkbox";
 
+
+// TODO: Remove this whole method when we upgrade to Blacklight 8, provided that 
+//       https://github.com/projectblacklight/blacklight/pull/3133 is merged
+//
+// Add the passed in contents to the modal and display it.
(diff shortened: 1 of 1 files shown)


app/javascript/application.js
```diff
@@ -52,6 +52,27 @@ import "./tooltip";
 import "./truncate";
 import "./update-hidden-inputs-by-checkbox";
 
+// Patched this method so that scripts returned from the ajax call are executed.
+// Specifically the script for recaptcha
+Blacklight.modal.receiveAjax = function (contents) {
+  // does it have a data- selector for container?
+  // important we don't execute script tags, we shouldn't.
```
taylor-steve, line 59:
```suggestion
```
Can we remove this one comment to avoid confusion?

taylor-steve (changes requested):

app/javascript/application.js
taylor-steve: This is great and works perfectly for the recaptcha issue. This jquery-less version would strip other attributes from script tags (e.g., integrity, crossorigin) I believe? Not a problem here, but do you think it could be a concern in more general use?

taylor-steve (approved):

PR #3801, opened 2024-01-25
Remove unused conditional

jcoyne:
This makes the code easier to understand

### DIFF
--- a/app/helpers/preview_helper.rb
+++ b/app/helpers/preview_helper.rb
@@ -1,8 +1,5 @@
 module PreviewHelper
-  def preview_data_attrs(show_preview = true, preview_type, id, target)
-    if show_preview
-      attrs = "data-behavior=\"#{preview_type}\" data-preview-url=\"#{preview_path(id)}\" data-preview-target=\"#{target}\""
-    end
-    (attrs ||= '').html_safe
+  def preview_data_attrs(preview_type, id, target)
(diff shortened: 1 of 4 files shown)


app/helpers/preview_helper.rb
```diff
@@ -1,8 +1,5 @@
 module PreviewHelper
-  def preview_data_attrs(show_preview = true, preview_type, id, target)
-    if show_preview
-      attrs = "data-behavior=\"#{preview_type}\" data-preview-url=\"#{preview_path(id)}\" data-preview-target=\"#{target}\""
-    end
-    (attrs ||= '').html_safe
+  def preview_data_attrs(preview_type, id, target)
```
marlo-longley, line 2: It may be helpful orientation to mention in a comment here that the possible values for `data-behavior` passed through `preview_type` (`preview-gallery`, `preview-filmstrip`, and `preview-brief`) are added so that javascript can pick them, and that there are corresponding javascript files such as `jquery.preview-filmstrip.js` where this happens.

jcoyne: That is being removed in another PR (https://github.com/sul-dlss/SearchWorks/pull/3803) , so I'm not sure it's worthwhile

marlo-longley (approved):

corylown (approved):

PR #3803, opened 2024-01-25
Decouple preview_filmstrip from async_collection members

jcoyne:
Accomplished by using stimulus.  Stimulus automatically binds to 'data-controller' when a node is added to a document. This allows us to remove the coupling with async_collection_members.


corylown (approved):

corylown:
@jcoyne approved, but needs rebase.

PR #3805, opened 2024-01-25
Restore blacklight-hierarchy javascript

### LINKED ISSUE #3802: Call number facet hierarchy is broken
Prior to [v3.21.14](https://github.com/sul-dlss/SearchWorks/releases/tag/v3.21.14) the call number facet displayed a tree structure to browse the call number hierarchy. This is currently broken and only displays a flat list of call number types.

Current behavior:
Screenshot 2024-01-25 at 9 03 36 AM

Expected behavior:
Screenshot 2024-01-25 at 9 03 28 AM

jcoyne:
Fixes #3802


corylown (changes requested):
@jcoyne in the previous implementation the hierarchy would stay open to the selected node after the facet is applied:
Screenshot 2024-01-25 at 11 25 55 AM

With this implementation it is closed after the facet the selected:
Screenshot 2024-01-25 at 11 26 04 AM

jcoyne:
@corylown I think that is a separate issue due to the Bootstrap 4 upgrade.  I can look at that too, but it looks like blacklight-hierarchy has an un-listed dependency on bootstrap 3.

corylown (approved):

PR #3806, opened 2024-01-25
wrap collection summary in helpers.format_record_html

### LINKED ISSUE #3795: Line break and ampersand work in Collection description needs to extend to "Context" box in item record view
Related to: https://github.com/sul-dlss/SearchWorks/issues/3699

I just noticed that we have encodings (i.e. &#13) and issues with line breaks in the Collection description that appears in the "Context" box on the item record view.

Example record: https://searchworks.stanford.edu/view/sy691hy9213
image

dnoneill:
closes #3795

### DIFF
--- a/spec/components/access_panels/in_collection_component_spec.rb
+++ b/spec/components/access_panels/in_collection_component_spec.rb
@@ -16,4 +16,20 @@ RSpec.describe AccessPanels::InCollectionComponent, type: :component do
       expect(page).to have_no_content("Digital content")
     end
   end
+
+  describe 'merged records with collection members that have encoded summaries' do
+    let(:document) { SolrDocument.new(id: '1') }
+    let(:parent) { SolrDocument.new(id: '2', summary_display: ['Summary &amp; Display']) }
+
+    before do
+      allow(document).to receive_messages(parent_collections: [parent])
+      render_inline(described_class.new(document:))
+    end
+
+    it 'renders summary display with summary display encoded properly' do
+      expect(page).to have_css("h3", text: 'Item belongs to a collection')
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Would you mind adding a component unit test for this?

spec/components/access_panels/in_collection_component_spec.rb
```diff
@@ -16,4 +16,20 @@
       expect(page).to have_no_content("Digital content")
     end
   end
+
+  describe 'merged records with collection members' do
```
jcoyne, line 20: would you mind updating this to reflect that we're testing escaped characters in the summary field?

dnoneill: not at all. good catch.

spec/components/access_panels/in_collection_component_spec.rb
```diff
@@ -16,4 +16,20 @@
       expect(page).to have_no_content("Digital content")
     end
   end
+
+  describe 'merged records with collection members that have encoded summaries' do
```
jcoyne, line 20: This test is for the AccessPanels::InCollectionComponent, but the change you made was to the AccessPanels::CollectionComponent. Can you create a test file for that component and move your test there?

jcoyne (changes requested):

spec/components/access_panels/in_collection_component_spec.rb
jcoyne: Actually, it appears that these two components wrap one another, so we can keep it like this.

jcoyne (approved):

PR #3814, opened 2024-01-26
update render_real_time_availability_request_link? to not render

dnoneill:
request link when item not lendable

Not sure about this. My gut reaction is that `return false if item.request_link.render?` should be `return false if !item.request_link.render?` due to the conditions of render but I am not familiar enough with edge cases to be sure. Please advise if this is the right route, my gut was or if there is a better route I did not think of.


corylown (approved):
Some of the logic around determining whether to display an item level request link is confusing. And it would good to add a test. But there is some urgency to fix this so I am going to merge this and write up an issue for us to take a closer look.

PR #3822, opened 2024-01-29
update livereload to work correctly in interval between index update,…

### LINKED ISSUE #3815: clean up live lookup request button
cleanup from #3810. Requires index and page to update to allow request button to be shown if status of item changes.

dnoneill:
… update tests

closes #3815

### DIFF
--- a/app/policies/item_request_link_policy.rb
+++ b/app/policies/item_request_link_policy.rb
@@ -12,6 +12,13 @@ class ItemRequestLinkPolicy
     item.on_order?
   end
 
+  # Check to see if request types include hold or recall.
+  # This is used to make sure live lookup can't create a request link for items that aren't eligible for hold/recall.
+  def item_allows_hold_recall?
+    item.allowed_request_types.include?('Hold') ||
+      item.allowed_request_types.include?('Recall')
+  end
+
   private
 
   attr_reader :item
--- a/app/services/live_lookup/folio.rb
+++ b/app/services/live_lookup/folio.rb
@@ -21,7 +21,8 @@ class LiveLookup
             item_id: item.fetch('id', nil),
             due_date: due_date(item),
             status: status(item),
-            is_available: available?(item)
+            is_available: available?(item),
+            is_requestable_status: requestable_status?(item)
           }
         end
       end
@@ -35,6 +36,13 @@ class LiveLookup
 
     def folio_client
       @folio_client ||= FolioClient.new
+    end
+
+    # Check to see if we can add a request button based on the status in folio
+    # and predetermined list of status that can be recalled
+    def requestable_status?(item)
+      status = item.fetch('status', nil)
+      Settings.folio_hold_recall_statuses.include?(status)
     end
 
     def status(item)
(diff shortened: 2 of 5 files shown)


app/services/live_lookup/folio.rb
```diff
@@ -37,6 +38,13 @@ def folio_client
       @folio_client ||= FolioClient.new
     end
 
+    # Check to see if we can add a request button based on the status in folio
+    # and predetermined list of status that can be recalled
+    def folio_status_check(item)
```
corylown, line 43: It's a private method, so 🤷 , but I think we usually prefer methods that return true/false to be named with a question mark.
```suggestion
    def requestable_status?(item)
```

app/services/live_lookup/folio.rb
```diff
@@ -21,7 +21,8 @@ def records
             item_id: item.fetch('id', nil),
             due_date: due_date(item),
             status: status(item),
-            is_available: available?(item)
+            is_available: available?(item),
+            is_requestable_status: folio_status_check(item)
```
corylown, line 25:
```suggestion
            is_requestable_status: requestable_status?(item)
```

app/policies/item_request_link_policy.rb
```diff
@@ -12,6 +12,13 @@ def show?
     item.on_order?
   end
 
+  # Check to see if request types include hold or recall.
+  # This is used to make sure live reload can't create a request link for items that aren't holdable/requestable.
```
corylown, line 16: I think we want to be clear this is about hold/recall and not other kinds of requests.

```suggestion
  # This is used to make sure live lookup can't create a request link for items that aren't eligible for hold/recall.
```

lib/holdings/item.rb
```diff
@@ -164,8 +164,9 @@ def request_link
       @request_link ||= ItemRequestLinkComponent.new(item: self)
     end
 
+    # Check to see if they item can be requested for live reloading.
```
corylown, line 167:
```suggestion
    # Check to see if the item is ever eligible for hold/recall. Used to determine whether a request link could be added by live lookup. 
```

lib/holdings/item.rb
```diff
@@ -164,8 +164,9 @@ def request_link
...
     def requestable?
-      @requestable ||= ItemRequestLinkPolicy.new(item: self).show?
+      @requestable ||= ItemRequestLinkPolicy.new(item: self).item_allows_request_recall?
```
corylown, line 169:
```suggestion
      @requestable ||= ItemRequestLinkPolicy.new(item: self).item_allows_hold_recall?
```

app/policies/item_request_link_policy.rb
```diff
@@ -12,6 +12,13 @@ def show?
...
+  def item_allows_request_recall?
```
corylown, line 17:
```suggestion
  def item_allows_hold_recall?
```

corylown (changes requested):
@dnoneill this seems to work well. My suggestions are about naming and trying to make sure the comments are clear enough so we/I can understand what's going on whenever we have to revisit this. Happy to discuss.

corylown (approved):

PR #3832, opened 2024-01-31
Add check for additional Shibboleth attribute to handle cases not covered by suAffiliation alone

### LINKED ISSUE #3690: Add privilege group to auth check when determining access to articles+
Reported by James Harris in Technical Services, E-Resources division.

Two users with sponsored SUNet IDs reported that they can't access articles+, stanford-only resources even though the sponsoring faculty member indicated during sponsorship that they should be eligible to access online library resources. Looking at their Regadmin accounts, the users were showing a suAffiliation =`stanford:student:recent` despite the active sponsorship.  Based on the ranking in the [UIT Registry Affiliation table](https://uit.stanford.edu/service/registry/data-affiliation-types-qualifiers), we would have expected `stanford:affilate:sponsored` to outrank the `stanford:student:recent` affliation. 

Helen Clark provided the following explanation, via email, when asked specifically about one of the users - You Jia Zhu (SUNet ID = youjia):

> 
> - You are correct that the affiliate:sponsored affiliation ranks higher than the student:recent affiliation.
> - For a person to have the affiliate:sponsored affiliation, their SUNet ID sponsorship must be in 'active' status. 
> - If a recent student is sponsored for a SUNet ID but is still eligible for that level of SUNet ID services due to the recent student affiliation, the SUNet ID sponsorship will stay in 'created' status.   Once that level of SUNet ID services is not provided by the student recent affiliation,  the sponsorship status will change from 'created' to 'active', and the affiliation will become affiliate:sponsored.     
> - A recent student with a sponsorship in 'created' status though will be added to the  stanford:administrative and stanford:affiliate-sponsored privilege groups on the start date of the sponsorship.    (Please recall our 'created status' work in 2016.)   
> If the recent student is sponsored for Library e-resources, the student is added to the stanford:library-eresources-eligible privilege group.   
> 
> In the case of You Jia Zhu. SUNet ID: youjia: 
> 
> - The SUNet ID is currently active at the base service level due to the student:recent affiliation.
> - The base service level SUNet ID sponsorship is in 'created' status so the affiliation continues to be student:recent.
> - The library e-resources sponsorship is in 'active' status. 
> - You Jia's privilege groups include: stanford:administrative, stanford:affiliate-sponsored, stanford:library-eresources-eligible and stanford:student-recent.    
> - If You Jia needs the affiliate:sponsored affiliation, the department could sponsor You Jia at the full-service level.  Since You Jia has base SUNet ID services currently , a SUNet ID sponsorship at the full-service level would provide additional services and therefore the sponsorship would activate ('active' status).   The affiliation would then become affiliate:sponsored.   
> 

In my reading of Helen's response, the `stanford:student:recent` will outweigh the `stanford:affilate:sponsored` always.

We need to add a check for an active privgroup = `stanford:library-eresources-eligible` to our affiliation check, and we may need to include 'created' as an eligible status (in addition to 'active')

hudajkhan:
Closes #3690 

Background:
As stated in the original issue, we can have a case where the Shibboleth attribute "suAffiliation" brings back "stanford:student:recent" which does not provide the user the kind of access they need to articles.  We can give this person access if we read in the value "member" in the "unscoped-affiliation" attribute, which is a mapping from "eduPersonAffiliation" and which has a one to one correspondence with the privilege group "stanford:library-eresources-eligible".  Please read original ticket for discussion around the reason for using this field and value. 

What this pull request does:

- Retrieve the "unscoped-affiliation" Shibboleth attribute in devise so that these values are read in and written to the user model. 
- Allow the additional "person_affiliations" attribute to be writable in the user model, so that information can be written to the model in the devise code on initialization.
- Add methods to the user model to retrieve the person_affiliations (i.e. unscoped-affiliation) info. 
- Change the stanford affiliation check for unscoped-affiliation first, as this should in fact cover all the suAffiliation categories as well.  If that fails, we still check against suAffiliation and send a honeybadger notification if that returns true when unscoped-affiliation returns false.   In that case, the user will still be granted stanford affiliated access.


app/models/user.rb
```diff
@@ -23,17 +23,41 @@ def affiliations
     @affiliations ||= ENV['suAffiliation']
   end
 
+  def person_affiliations
+    @person_affiliations ||= ENV['unscoped-affiliation']
+  end
+
+  # If a user has proper Stanford affiliation to allow access to resources now depends
+  # on two kinds of affiliation information: the 'su_affiliated' method which returns true
```
cbeer, line 31: Do we need to check both places (e.g. are there people with an `su_affiliation` group that grants access that `unscoped-affiliation` member doesn't), or is this a temporary thing for migration?

app/models/user.rb
```diff
@@ -23,17 +23,41 @@ def affiliations
...
     affiliations.split(';').any? do |affiliation|
       Settings.SU_AFFILIATIONS.include?(affiliation.strip)
     end
   end
 
+  def person_affiliated?
+    return false if person_affiliations.blank?
+
+    person_affiliations.split(';').any? do |person_affiliation|
+      person_affiliation.strip === Settings.UNSCOPED_AFFILIATION # rubocop:disable Style/CaseEquality
```
mjgiarlo, line 54: If you're doing a straight-up string comparison, I think you could consider `==` instead of `===` here since the latter is typically used to gauge case equality. (In this way, Ruby and JavaScript are annoyingly different.)

hudajkhan: Ok, I can change that

app/models/user.rb
hudajkhan: This is a good question. I'll follow up with Sarah and Irina.  I know the reverse is true (i.e. unscoped-affiliation grants access su_affiliation doesn't) but not sure about this question.

mjgiarlo (approved):
Other than the triple-equals tweak, this LGTM. Happy to approve pending a tiny `===` change.

app/models/user.rb
hudajkhan: I also didn't want to mess with that part of the logic because the su_affiliation allowed list is quite long.

jcoyne (dismissed):
Can we write a test for this?

jcoyne:
@hudajkhan would you mind squashing these commits?

hudajkhan:
Updating the approach in this fashion: Relying on unscoped-affiliation at first, and if that doesn't go through but suAffiliation would allow the user to get access, we'll give them access and possibly log when that happened.

hudajkhan:
@jcoyne We can squash commits on merge as well, which would also work.

mjgiarlo (approved):

PR #3834, opened 2024-01-31
Use passed argument rather than ivar for add_purl_embed_header

jcoyne:
This will allow it to be called by components


jcoyne:
Superseded by #3826

PR #3844, opened 2024-02-01
update index_parent_collections to use solr doc url if exists

### LINKED ISSUE #3840: broken links to digital collections
https://searchworks.stanford.edu/?search_field=search&q=biface links under the digital collection link are broken and do not work.

dnoneill:
closes #3840

### DIFF
--- a/app/models/concerns/collection_member.rb
+++ b/app/models/concerns/collection_member.rb
@@ -14,7 +14,7 @@ module CollectionMember
   end
 
   def index_parent_collections
-    return nil if !is_a_collection_member? || self[:collection_with_title].blank?
+    return nil if self[:collection_with_title].blank? || parent_collections.blank?
 
     unless @index_parent_collections
       @index_parent_collections = self[:collection_with_title].map do |collection_with_title|


corylown:
@dnoneill I don't think this does what we want. Some collections are released to searchworks and the desired behavior is that if the collection is in searchworks it links to the collection record in searchworks. For example, this https://searchworks.stanford.edu/catalog?f%5Bcollection%5D%5B%5D=bc677qs7597 links to https://searchworks.stanford.edu/view/bc677qs7597. With the change you've proposed in this PR it would link to PURL instead at https://purl.stanford.edu/bc677qs7597. On the record page I believe SearchWorks queries the SearchWorks Solr index to determine that the collection record exists before linking to it. That probably isn't a scalable approach from the index page where there could be 100 records with collection links to look up, but illustrates the problem. I think we'd need some information on the object record to tell us if the parent collection has been released so SearchWorks knows what to do.

jcoyne (changes requested):
There should probably be a test added around `index_parent_collections` for this change.

Issue #3736, opened 2024-01-16
Use `courses_json_struct` instead of `crez_course_info`


cbeer:
See https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/folio_config.rb#L2517

1) document the shape of `courses_json_struct` in the traject config
2) confirm it has all the necessary data as `crez_course_info` (and add anything that's missing)
3) update `CourseReserves` concern to pull data from `courses_json_struct`
4) figure out what, if anything, to do with the `CourseReservesController` course info faceting

hudajkhan:
[Also see: ](https://github.com/sul-dlss/searchworks_traject_indexer/blob/e0491c887170813888862756b2aa291304fc728c/lib/traject/config/folio_config.rb#L2599)

hudajkhan:
This may require more clarification from Chris, especially regarding the first bullet.

hudajkhan:
Chris says (1) means we should understand what is actually there in the FOLIO data that is coming back.

jcoyne:
There are three locations in the code where this is being used:

1. In the CourseReservesController, where it's using it as a facet.  https://github.com/sul-dlss/SearchWorks/blob/main/app/controllers/course_reserves_controller.rb#L8
1. In `CourseReserveAccessPointHelper#create_course` https://github.com/sul-dlss/SearchWorks/blob/eceef766c1db35c6f1898cfe94fa3607c5e9cc0a/app/helpers/course_reserve_access_point_helper.rb#L2
1. In `CourseReserves::Processor` (https://github.com/sul-dlss/SearchWorks/blob/eceef766c1db35c6f1898cfe94fa3607c5e9cc0a/app/models/concerns/course_reserves.rb#L12)

jcoyne:
because we do not index `*_struct` fields. I don't believe we are able to replace the usage in `CourseReservesController`.

cbeer:
I don't think it makes sense to index all of `courses_json_struct` for faceting (and I don't think it'd do what we want either?). I wouldn't be surprised if we add a new field to support `CourseReservesController` or do something without solr to pull the course information.

jcoyne:
I believe we can proceed on this when the `response.numFound` for the first query has an identical number as the latter query (currently 1101)
http://sul-solr-prod-b.stanford.edu/solr/#/searchworks-folio-dev/query?q=courses_folio_id_ssim:*&q.op=OR&indent=true
[...]

PR #3853, opened 2024-02-01
Course reserves are now driven from the Folio data

jcoyne:
Ref #3736


hudajkhan:
Trying this out locally and on staging, the number of courses on the reserves page with FOLIO is 687 while it is 503 in current production. This isn't an error, but just making a note.  

_When clicking on an individual course reserve link_
Current production: 
Screenshot 2024-02-16 at 1 13 58 PM

Staging with branch:
Screenshot 2024-02-16 at 1 15 05 PM

Also of note, in production, the URL also includes the instructor facet "&f[instructor][]=Alexander+Nemerov" whereas staging has just the course number. (This may also be fine. Just noting the difference)

In staging, the course reserve red button does not seem to be visible in the search results view as it is in production. 

_When clicking on item_
Current production:
Screenshot 2024-02-16 at 1 17 21 PM

Staging with branch:
Screenshot 2024-02-16 at 1 17 49 PM

hudajkhan:
Noting a few more differences (these aren't necessarily errors, just differences):

- The facet label for courses changed in this PR from "course" to "courses_folio_id_ssim" . It may be useful for consistency to keep the same label, but that can be up for discussion. 
- In the example above, the ordering of the courses in an individual page "course reserves" box is different between staging and production, but the same set is represented in both places.  This may be a minor difference.  In production, in the example above, it seems the list is ordered alphabetically by course number.

jcoyne:
@hudajkhan I believe the missing red button is because we need this config change to be deployed: https://github.com/sul-dlss/sul-solr-configs/pull/310

Aside from that I believe all the changes were intentional.

jcoyne:
@dbranchini @saseestone can you speak about what we want to do with the course reserve panel? 
Huda correctly pointed out we don't have it sorted by course name, however, in this case this is a single cross listed course. (course listing 46f8e5a3-a407-4277-9cfc-ce9ab0307c8b).  Is there some change we want to make to visualize this cross-lising?
Screenshot 2024-02-16 at 3 32 20 PM

cbeer:
I've deployed the solrconfig changes (after figuring out that we were using a special Argo branch of sul-solr-configs....).

I'm pretty sure we don't want to use the facet label "Courses (FOLIO)". Our patrons don't know what FOLIO is. @dbranchini ?

jcoyne:
@hudajkhan it's now sorted by course number:

Screenshot 2024-02-21 at 1 41 13 PM

PR #3857, opened 2024-02-02
strip leading a from collection search

### LINKED ISSUE #3854: Restore "Appears in" context for virtual objects
When an item is part of a virtual object, SearchWorks used to show a link to the record for the whole virtual object under the heading "Appears in" within the context widget on the item page. I believe this is the original ticket for this feature: https://github.com/sul-dlss/SearchWorks/issues/1051 

At some point in the last year or two, this context link has stopped appearing.

**Example**

This map is a page from a larger atlas: https://searchworks.stanford.edu/view/sd322dt2118

In the purl page display, you can see that it has a link to the purl for the whole atlas:

> Screenshot 2024-02-01 at 3 27 38 PM

In the context sidebar, there should be a link to the [SW record for the whole atlas](https://searchworks.stanford.edu/view/10448742) (not the purl page), but currently there's only a link to the collection:

> Screenshot 2024-02-01 at 3 28 40 PM

dnoneill:
closes #3854 

The code in app/models/concerns/solr_set.rb is what fixes this issue. The rest is consolidating the stripping out the a from the catkey.

### DIFF
--- a/app/helpers/collection_helper.rb
+++ b/app/helpers/collection_helper.rb
@@ -1,4 +1,8 @@
 module CollectionHelper
+  def self.strip_leading_a(collection_id)
+    collection_id.sub(/^a(\d+)$/, '\1')
+  end
+
   def link_to_collection_members(link_text, document, options = {})
     link_to(link_text, collection_members_path(document), options)
   end
@@ -28,7 +32,7 @@ module CollectionHelper
     if @response.documents.first&.index_parent_collections.present?
       collection = @response.documents.first.index_parent_collections.find do |coll|
         # strip leading 'a' from collection catkeys
-        coll[:id] == collection_id.sub(/^a(\d+)$/, '\1')
+        coll[:id] == CollectionHelper.strip_leading_a(collection_id)
       end
       return document_presenter(collection).heading if collection.present?
     end
--- a/app/models/concerns/collection_member.rb
+++ b/app/models/concerns/collection_member.rb
@@ -19,7 +19,7 @@ module CollectionMember
     unless @index_parent_collections
       @index_parent_collections = self[:collection_with_title].map do |collection_with_title|
         id, title = collection_with_title.split('-|-').map(&:strip)
-        SolrDocument.new(id: strip_a(id), title_display: title)
+        SolrDocument.new(id: CollectionHelper.strip_leading_a(id), title_display: title)
       end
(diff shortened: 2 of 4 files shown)


app/helpers/collection_helper.rb
```diff
@@ -1,4 +1,9 @@
 module CollectionHelper
+  # strip leading 'a' from collection catkeys
+  def self.strip_leading_a(collection_id)
+    collection_id.sub(/^a(\d+)$/, '\1')
```
mjgiarlo, line 4: Not functionally different but worth considering a string-based implementation?

```suggestion
    collection_id.delete_prefix('a')
```

app/helpers/collection_helper.rb
```diff
@@ -1,4 +1,9 @@
 module CollectionHelper
+  # strip leading 'a' from collection catkeys
```
mjgiarlo, line 2: Especially with the string-based implementation and given the very clear name you've given the method, I think we could also consider dropping the comment.

app/helpers/collection_helper.rb
```diff
@@ -1,4 +1,9 @@
...
+  def self.strip_leading_a(collection_id)
```
mjgiarlo, line 3: Can you say a little bit about using a class method here vs. an instance method? It looks like most of the current helper methods are instance methods.

dnoneill: From what I can tell you can't use an instance method in concerns without importing the entire helper method. So I went with a class method so that we could just call the class method in the concern files.

app/helpers/collection_helper.rb
dnoneill: I used the pattern that was previously used but I think this probably makes more sense.

cbeer: We've been using `sub(/^a(\d+)$/, '\1')` to avoid accidentally stripping an `a` from catkey that wasn't migrated over from symphony (where a leading `a` might happen.. I'm not sure we have any like this yet, but...)

dnoneill: Okay. Changed it back.

app/helpers/collection_helper.rb
mjgiarlo: Gotcha, thank you!

Issue #3467, opened 2023-09-01
Replace/update Searchworks fixture data with FOLIO-flavored data


thatbudakguy:
- [x] uuid_ssi
- [x] item_display_struct

hudajkhan:
Fields in checkboxes are new fields in FOLIO.

hudajkhan:
To begin: review what needs to be cleaned up still, and if there is fixture data (like uuid) that is being generated in the tests that we could just put in the fixture data directly.

hudajkhan:
Even if the values in fixture data are not real, we want the structure of the fixture data to mirror the structure coming back from FOLIO.  This ticket is about ensuring we take the time to review this area.

PR #3870, opened 2024-02-05
Remove references to unused title_uniform_display solr field

marlo-longley:
Relates to #3467 
See https://github.com/sul-dlss/searchworks_traject_indexer/blob/cdad30e01b4e44880af5c68923464adad8a74e2e/lib/traject/config/folio_config.rb#L286

There seem to be no documents with `title_uniform_display` in our current production index, and all tests pass without it.
The indexer note says to remove the field in 2010!

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -214,7 +214,6 @@ class CatalogController < ApplicationController
     #   The ordering of the field names is the order of the display
     config.add_show_field "title_full_display", label: "Title"
     config.add_show_field "vern_title_full_display", label: "Title"
-    config.add_show_field "vern_title_uniform_display", label: "Uniform Title"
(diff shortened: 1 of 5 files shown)


jcoyne (approved):

cbeer:
Isn't this used in e.g. [13553090](https://searchworks.stanford.edu/view/13553090)?

jcoyne:
It looks like title_uniform_display can go, but vern_title_uniform_display has to stay.

marlo-longley:
I reviewed this more carefully and found that any needs we have for a uniform title are served by by `uniform_title_display_struct`. With this branch checked out, and using the indexer from this branch: https://github.com/sul-dlss/searchworks_traject_indexer/pull/1324/files -- I locally indexed a record and found the display was still correct for Uniform Title.


Screenshot 2024-02-07 at 1 25 37 PM

jcoyne (approved):

Issue #3876, opened 2024-02-06
Gap disappears on "brief" view after opening and closing preview


jcoyne:
Before doing anything there is some spacing between documents
<img width="785" alt="Screenshot 2024-02-05 at 10 09 00 PM" src="https://github.com/sul-dlss/SearchWorks/assets/92044/10d483d2-782a-4e60-a7aa-450296185891">

Then open the preview
<img width="785" alt="Screenshot 2024-02-05 at 10 09 08 PM" src="https://github.com/sul-dlss/SearchWorks/assets/92044/bc2d35c8-8990-46ad-8fee-166f078277aa">

After closing the preview the gap disappears.
<img width="789" alt="Screenshot 2024-02-05 at 10 09 15 PM" src="https://github.com/sul-dlss/SearchWorks/assets/92044/9756d020-0722-4221-a265-c79dcad3c1ca">

Before it is open, it appears on the doc
<img width="819" alt="Screenshot 2024-02-05 at 10 13 46 PM" src="https://github.com/sul-dlss/SearchWorks/assets/92044/fde8c7df-ef34-45aa-8b2c-7ffe747779ae">

This is what gives us the margin.

After it is hidden it has `display:none`
<img width="911" alt="Screenshot 2024-02-05 at 10 14 46 PM" src="https://github.com/sul-dlss/SearchWorks/assets/92044/5a4f287a-a8ed-491b-8ce6-3f5bf28d94dd">

I think we can fix this by moving the margin from `.preview-container` to `.brief-document`

PR #3877, opened 2024-02-06
Remove unnecessary class

jcoyne:
The 'hidden' class isn't confering any styles here

### DIFF
--- a/app/javascript/jquery.preview-brief.js
+++ b/app/javascript/jquery.preview-brief.js
@@ -39,9 +39,7 @@ import PreviewContent from './preview-content'
           .html('Close');
 
         $briefTarget.hide();
-        $previewTarget.removeClass('hidden');
         $previewTarget.show();
-
       }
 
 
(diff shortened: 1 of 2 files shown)


cbeer:
It isn't, but shouldn't it? That seems like the problem with #3876.

jcoyne:
@cbeer the problem there is that show() and hide() method in jquery set the display property.  They don't set any classes.

cbeer:
I'm getting a little confused about all the PRs around this functionality. I've put up https://github.com/sul-dlss/SearchWorks/pull/3888 which restores the previous styling + behavior. I suspect we want to move away from the `.hidden` class (from Bootstrap 3), but it'd really help me if we start from the previous state and create a single PR to update it and restyle the preview component as necessary.

I'd suggest we merge #3888 and you can create a new PR (replacing #3877 + #3887?) that does that.

jcoyne:
Okay, I merged #3888 and moved this into #3887

PR #3883, opened 2024-02-06
Add a feature spec for live-lookup

### LINKED ISSUE #3882: Add a feature test for live-lookup
Currently there is no test of the Javascript.

jcoyne:
Fixes #3882


app/models/solr_document.rb
```diff
@@ -116,6 +115,7 @@ def to_semantic_values
   use_extension(Blacklight::Document::DublinCore)
 
   attribute :document_formats, :array, FORMAT_KEY
+  attribute :live_lookup_id, :string, 'uuid_ssi'
```
jcoyne, line 118: I replaced the `LiveLookupIds` module with this one-liner.

corylown (approved):

PR #3885, opened 2024-02-06
Restore the pre-bootstrap-upgrade body-color

cbeer:
It looks like bootstrap changed their variable name from 3 -> 4.

### DIFF
--- a/app/assets/stylesheets/bootstrap-variables.scss
+++ b/app/assets/stylesheets/bootstrap-variables.scss
@@ -9,6 +9,7 @@ $danger: $bright-red;
 
 // Body
 $body-bg: $sul-body-bg;
+$body-color: $sul-text-color;
 
 // Text / Links
 $text-color: $sul-text-color;


app/assets/stylesheets/bootstrap-variables.scss
```diff
@@ -9,6 +9,7 @@ $danger: $bright-red;
 
 // Body
 $body-bg: $sul-body-bg;
+$body-color: $sul-text-color;
```
jcoyne, line 12: The problem here is that `$sul-text-color` is set to `#000` which is not the color specified by the stanford identity guide and neither is it consistent with the styles on the library website.

jcoyne (changes requested):

app/assets/stylesheets/bootstrap-variables.scss
cbeer: I'm not sure why Stanford's current branding colors matter. This was how the site was originally designed and took into account things like color contrast. It's fine if we want to adjust colors later, but getting back to the original design seems like it fixes some regressions.

corylown: I propose we merge this change and consider a branding/palate update as a separate issue to be addressed with the help of a designer.

jcoyne:
@corylown I've already engaged the designers and they indicated that they want us to follow the Stanford brand guidelines.

PR #3887, opened 2024-02-06
Remove unnecessary hidden class

### LINKED ISSUE #3876: Gap disappears on "brief" view after opening and closing preview

jcoyne:
Fixes #3876

### DIFF
--- a/app/assets/stylesheets/modules/brief-view.scss
+++ b/app/assets/stylesheets/modules/brief-view.scss
@@ -1,18 +1,11 @@
 .brief-document {
   @include document;
-  margin-bottom: 0px;
+  margin-bottom: -1px;
   padding-left: 20px;
 
   h3.index_title {
     @include h3-index-title-small-font;
     word-wrap: break-word;
-  }
-
-  .preview {
-    border-radius: 0;
-    border-width: 0;
-    margin-top: 0;
-    padding: 0;
   }
 
   .sul-icon {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/modules/brief-view.scss
```diff
@@ -1,20 +1,13 @@
...
-  margin-bottom: 0px;
+  margin-bottom: -1px;
   padding-left: 20px;
 
   h3.index_title {
     @include h3-index-title-small-font;
     word-wrap: break-word;
   }
 
-  .preview {
```
jcoyne, line 11: the `preview` class is not nested under `brief-document` in the DOM, so these rules are never applied.

app/javascript/jquery.preview-brief.js
```diff
@@ -62,7 +56,6 @@ import PreviewContent from './preview-content'
 
       function init() {
         $triggerBtn = $item.find('*[data-behavior="preview-button-trigger"]');
-        $briefTarget = $item.find('.brief-container .col-md-8');
```
jcoyne, line 65: There are no DOM elements matching this selector, so this isn't doing anything.

mjgiarlo (approved):

PR #3892, opened 2024-02-06
remove LOC constants, update tests, update live lookup

dnoneill:
Requires https://github.com/sul-dlss/searchworks_traject_indexer/pull/1326 to work.


corylown:
@dnoneill this is looking pretty good to me after doing some testing. It appears to fix #3912 as you mentioned on that issue. And the nice thing is we could deploy this to searchworks at any point and then let the indexer catch up with the changes in [searchworks_traject_indexer/pull/1326](https://github.com/sul-dlss/searchworks_traject_indexer/pull/1326), at least it seems so in my testing.

There were a few style and display differences I thought I'd point out in case @cbeer or others want to take a look:

***

**https://searchworks.stanford.edu/view/10632848**
`Checked out` item
"Checked out" text displays as black instead of red

Prod
Screenshot 2024-02-13 at 4 21 49 PM

This branch
Screenshot 2024-02-13 at 4 22 24 PM

***

**https://searchworks.stanford.edu/view/in00000014511**
`On order` item
Same as checked out in that the expected red color is not applied to the text

Prod
Screenshot 2024-02-13 at 4 27 37 PM

This branch
Screenshot 2024-02-13 at 4 27 42 PM

***

**https://searchworks.stanford.edu/view/in00000009317**
`SUL-TS-PROCESSING` location
Maybe an improvement over prod in that we're not displaying both "in process" and "in transit", but I'm not sure if we prefer one over the other or if we want to display both for some reason.

Prod
Screenshot 2024-02-13 at 4 29 49 PM

This branch
Screenshot 2024-02-13 at 4 29 55 PM

***

**https://searchworks.stanford.edu/view/4415720**
`LAW-BINDERYPREP` (which lacks `availabilityClass": "In_process"` even though it should probably have it)

Prod
Screenshot 2024-02-13 at 4 35 17 PM

This branch
Screenshot 2024-02-13 at 4 35 23 PM

***

**https://searchworks.stanford.edu/view/14818612**
`SUL-TS-CC-BINDERYPREP` location
Has `availabilityClass": "In_process"` and seems like an improvement over prod in that we're not displaying both in process and checked out.

Prod
Screenshot 2024-02-13 at 4 45 40 PM

This branch
Screenshot 2024-02-13 at 4 45 44 PM

***
**https://searchworks.stanford.edu/view/13968683**
Fixes duplicate `in process` display

Prod
Screenshot 2024-02-13 at 4 53 41 PM

This branch
Screenshot 2024-02-13 at 4 53 53 PM

saseestone:
I hope this doesn't complicate things, but if we have an `availabilityClass": "In_process"`, we'll want to prioritize displaying that over displaying a folio status (e.g. "In transit" or "Checked out" with due date added)

So taking Cory's examples from above:

***

https://searchworks.stanford.edu/view/14818612
SUL-TS-CC-BINDERYPREP location
Has `availabilityClass": "In_process" `and seems like an improvement over prod in that we're not displaying both in process and checked out.

Prod:
image

We'd want to emphasis "In process" and not display "Checked out" + due date

***
https://searchworks.stanford.edu/view/in00000009317
SUL-TS-PROCESSING location
Maybe an improvement over prod in that we're not displaying both "in process" and "in transit", but I'm not sure if we prefer one over the other or if we want to display both for some reason.

I'll note it also has `availabilityClass": "In_process" ` 

Prod
image

We'd want to emphasize "In process" and not display "In transit"

***

In the case of: https://searchworks.stanford.edu/view/4415720
LAW-BINDERYPREP (which lacks `availabilityClass": "In_process"` even though it should probably have it)

Yes, it should have `availabilityClass": "In_process"`. I'm working with Law admins to get it added correctly. But I think the pattern would still hold - with an emphasis on "In process" and suppressing any FOLIO item status or temporary location discovery display name.

If the location does NOT have an `availabilityClass": "In_process"`, then I think it's fine to display it as you have in the this branch. 

***

dnoneill:
Couple of questions:
1. Are there other availability classes we want to preference over status or is it just "In transit"?
2. Do we want to include this fix on this branch or is this a separate issue? I have a fix I can push but I want to make sure this isn't something we don't want to include with this PR.

dnoneill:
see https://github.com/sul-dlss/SearchWorks/pull/3951 to fix problem Sarah outlined above.

app/services/live_lookup/solr.rb
```diff
@@ -71,14 +71,11 @@ def item_id
       end
 
       def available?
-        current_location_code != 'CHECKEDOUT' &&
-          Constants::UNAVAILABLE_LOCS.exclude?(current_location_code)
+        status && status == 'Available'
```
jcoyne, line 74:
```suggestion
        status == 'Available'
```

jcoyne (changes requested):

app/services/live_lookup/solr.rb
```diff
@@ -90,20 +87,11 @@ def home_location_code
...
-                        current_location_same_as_home_location?
-
-        true
+        @current_location_code ||= item['current_location']
       end
 
-      def current_location_same_as_home_location?
-        current_location.name == home_location.name ||
-          Constants::CURRENT_HOME_LOCS.include?(current_location_code)
+      def location_lookup
```
jcoyne, line 93: What is the purpose of this method? I'm not seeing anything calling it.

dnoneill: Looks like it is a function I missed when removing old code. Thanks for the catch.

jcoyne (approved):

PR #3895, opened 2024-02-06
Disable noscript on feedback form recaptcha

### LINKED ISSUE #3873: Feedback form reCAPTCHA fails intermittently
Sometimes, for unauthenticated users, the Feedback form Captcha will show the green success checkmark but the form will fail to submit with the error "You must pass the reCAPTCHA challenge".

Screenshot 2024-02-05 at 2 59 27 PM

At first glance this seems similar to https://github.com/sul-dlss/SearchWorks/issues/3808 in the sense that sometimes the noscript tags are processed by the browser and sometimes they are not, causing the 'g-recaptcha-response' to be empty for the form. The reason for this was clear in #3808, less so here as it is very intermittent and can vary from reload to reload on the same page.

I can recreate fairly reliably by:
* Viewing, not submitting, the feedback form at the SearchWorks homepage
* Selecting the "Online" Access facet and viewing the feedback form on that page. This will tend not to work.

Inspecting the recaptcha source, when you see this, the form will work:
Screenshot 2024-02-05 at 2 27 08 PM

When seeing this, the form will fail:
Screenshot 2024-02-05 at 2 28 41 PM

Turning off noscript like in #3808 appears to work, but I'm less clear on why we would need to do this here.

taylor-steve:
Fixes the feedback form reCAPTCHA issue in #3873.

In [SearchWorks v3.21.14](https://github.com/sul-dlss/SearchWorks/releases/tag/v3.21.14) and after, noscript tags in the form containing this reCAPTCHA sometimes do not get processed by the browser. When this happens, it breaks the reCAPTCHA auth.

### DIFF
--- a/app/views/shared/feedback_forms/_form_fields.html.erb
+++ b/app/views/shared/feedback_forms/_form_fields.html.erb
@@ -50,7 +50,8 @@
     <% if current_user.blank? %>
       <div class="form-group row">
         <div class="col-sm-9 offset-sm-3">
-          <%= recaptcha_tags id: "#{type}-recaptcha" %>
+          <%# Remove noscript for recaptcha. See https://github.com/sul-dlss/SearchWorks/issues/3873 %>
+          <%= recaptcha_tags id: "#{type}-recaptcha", noscript: false %>
 
           <p>(Stanford users can avoid this Captcha by logging in.)</p>
         </div>


cbeer:
This seems like the same problem + solution from https://github.com/sul-dlss/SearchWorks/pull/3812, right?

taylor-steve:
Same solution, same class of problem, but different specifics?

In #3812 it was clear that noscripts were left unprocessed as part of the blacklight modal build process (receiveAjax) and would always happen.

This isn't using that specific modal system as far as I can tell, and it does seem to be at least somewhat intermittent.

I feel like this solves the immediate problem, but doesn't address:
* Do we expect to be able to use noscript tags in that form (not sure if you read the other issue, this doesn't seem specific to reCAPTCHA, some static noscript content fails in the same way)
* If we do, I wasn't sure where to open that issue yet and with what details. I could open another more general issue in SearchWorks?

PR #3897, opened 2024-02-07
Organize the color palette

jcoyne:
Use the names from the identity guide


cbeer:
The current names are from the last revision of the identity guide:
https://web.archive.org/web/20200311081005/https://identity.stanford.edu/color.html#digital-color

My preference would be to stick with the current names until Darcy et al want to fully switch palettes. I'm worried mixing names between the different palettes will cause some confusion.

PR #3900, opened 2024-02-07
Remove hover style for managed purl

### LINKED ISSUE #3899: Link styling is wrong for managed purls
See https://searchworks.stanford.edu/view/8112988

The underline is at the bottom when hovering.
Screenshot 2024-02-07 at 9 18 18 AM

jcoyne:
Fixes #3899


jcoyne:
There still is a hover style, but it applies to the box you are hovering over.  It does not include underlining the text as the button is the whole box, not just the text.

PR #3903, opened 2024-02-07
Remove Blacklight config of show and index fields

jcoyne:
Blacklight is not creating these views for us


marlo-longley (commented):
I'm assuming you are commenting rather than deleting on purpose?
I checked this branch out and the displays seem to function as expected locally.

jcoyne:
Yes. I think this may help us understand which fields are used and which we can search for to find if they can be removed.

mjgiarlo (approved):

PR #3904, opened 2024-02-07
Leaflet doesn't seem to be used anymore

jcoyne:


corylown (commented):
I believe leaflet is used by `jquery.stackmap.js` to display stack maps.

jcoyne:
Thanks @corylown. I'll add a comment about that.

PR #3914, opened 2024-02-08
Turn off annotate_rendered_view_with_filenames in development

corylown:
This was enabled here: https://github.com/sul-dlss/SearchWorks/commit/cf1658fc3e6d2d53ca9e953a416716b80f9bac43#diff-d3c4b3f41072daa416f1920511e9b2e26caea8c5cec0a14cb9508589a4dafa47R69

But seems to cause some wonky display and link behavior problems in development, such as:

Extraneous "Summary" label:
Screenshot 2024-02-08 at 10 52 46 AM

Brief view record links that reload the current page instead of going to the full record view:
Screenshot 2024-02-08 at 10 53 19 AM

### DIFF
--- a/app/views/catalog/_home_page.html.erb
+++ b/app/views/catalog/_home_page.html.erb
@@ -83,7 +83,7 @@
             <h3>IIIF resources</h3>
           <% end %>
           <p>Find IIIF-compatible items in the Stanford Digital Repository.</p>
-          <p><a href="https://library.stanford.edu/projects/international-image-interoperability-framework/viewers">What is IIIF?</a></p>
+          <p><a href="https://iiif.io">What is IIIF?</a></p>
         </div>
       </div>
       </div>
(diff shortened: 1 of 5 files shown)


jcoyne:
Ah, I didn't restart my server. Yes it does resolve the problem.  This appears to affect how some of our javascript is working.

cbeer:
I've never heard of `annotate_rendered_view_with_filenames`, but it seems useful. Should we merge this and ticket further investigation?

corylown:
@cbeer yes, I agree it seems like a useful feature and would be worth seeing if we can make changes that would let us re-enable it without breaking stuff. As is it's making it difficult to figure out what's broken because of code changes vs this setting interfering with functionality.

jcoyne:
I have a fix in progress...https://github.com/sul-dlss/SearchWorks/pull/3916, but we need https://github.com/sul-dlss/SearchWorks/pull/3915 merged first.

jcoyne:
@corylown I think there's a different problem around "Brief view record links that reload the current page instead of going to the full record view:" in development mode. I'm seeing that even with this feature disabled.  This is because the tracking page is returning a 422 because `ActionController::InvalidAuthenticityToken (Can't verify CSRF token authenticity.)`

PR #3925, opened 2024-02-09
Add a docker compose file for starting solr

jcoyne:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,29 @@ The installation script will
 
 You will need to update the configuration in `config/settings.yml` for various parts of the app to work.  Please check that file for more information.
 
-Start Solr
+### Start Solr
 
+There are at least two options for starting Solr.
+
+#### Docker compose
+
+```shell
+docker compose up
+```
+
+#### solr_wrapper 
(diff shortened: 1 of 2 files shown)


taylor-steve:
Got docker working, so I checked this out!

You set the solr container environment variables but then reference host environment variables later. Is this to provide defaults or will most people have these variables already set locally/in a .env file? On a fresh checkout, running `docker compose up` fails for me with those two variables being unset.

If you're trying to set defaults, I think you can do something like this:
```yaml
version: "3.7"

services:
  solr:
    image: "solr:${SOLR_VERSION:-8.11.1}"
    volumes:
      - $PWD/config/solr_configs:/opt/solr/conf
      - $PWD/config/solr_configs/CJKFilterUtils-v3.0.jar:/opt/solr/contrib/CJKFilterUtils-v3.0.jar
    ports:
      - "${SOLR_PORT:-8983}:8983"
    entrypoint:
      - docker-entrypoint.sh
      - solr-precreate
      - blacklight-core
      - /opt/solr/conf
      - "-Xms256m"
      - "-Xmx512m"
```

In the case where those variables are needing to come from somewhere else, maybe we call that out in the readme.

This is going to be nice to have, thanks.

jcoyne:
@taylor-steve thanks for the feedback. I've made those changes.

dnoneill:
Works for me as well.

dnoneill (approved):

PR #3930, opened 2024-02-09
Remove unnecessary leading colon from yaml fixtures

jcoyne:
And use safe_load


jcoyne:
It's from the Psych library (for loading YAML) https://www.rubydoc.info/stdlib/psych/Psych.safe_load

marlo-longley:
I am about to edit a bunch of these fixture files so it would be good to merge this first.
Looks like it needs a rebase.

dnoneill (approved):

PR #3937, opened 2024-02-12
Update librarian view to have items

### LINKED ISSUE #3761: Revise librarian view on an item detail page
Public services support staff have long used the MARC 999, visible in Librarian view, to quickly view item level information needed for a variety of tasks. This view was especially useful to volunteers & temporary staff in special collections units, who do not have access to the ILS. When we moved to FOLIO, the Symphony-constructed 999 field ceased. This has negatively impacted staff, who now struggle to interpret the FOLIO json. Additionally, the FOLIO UI makes it difficult to quickly see item information if the Instance record has more than 50 items. 

Most desired is a “Quick look” that would contain Call number, Item ID (barcode), FOLIO permanent and effective location. If that’s not possible, having the FOLIO json structured in a tree format may help staff inexperienced with reading json - particularly when viewing records with large numbers of items.

UX Doc - https://docs.google.com/document/d/1NSgaHmcatoo8oH-dcB0e5cpNE9888I--v24sN-JGPMM/edit?usp=sharing

jcoyne:
Fixes #3761 

On https://searchworks-stage.stanford.edu/view/2800276
Before:
Screenshot 2024-02-14 at 3 15 29 PM

After:
Screenshot 2024-02-14 at 3 16 36 PM

After clicking items:
Screenshot 2024-02-14 at 3 16 48 PM

After for https://searchworks-stage.stanford.edu/view/2279186  (no items)
Screenshot 2024-02-14 at 3 19 37 PM

After for https://searchworks-stage.stanford.edu/view/14859222 (items in a different location)
Screenshot 2024-02-14 at 3 29 25 PM

After for https://searchworks-stage.stanford.edu/view/in00000008968 (on order item)

Screenshot 2024-02-15 at 2 17 37 PM

### DIFF
--- a/app/assets/stylesheets/bootstrap-variables.scss
+++ b/app/assets/stylesheets/bootstrap-variables.scss
@@ -53,7 +53,9 @@ $pagination-active-bg: $sul-pagination-active-bg;
 $pagination-active-border: $sul-pagination-active-bg;
 
 // Tables
-$table-accent-bg: $light-sandstone; // background for .table-striped
+$table-cell-padding: 0.5rem;
+$table-accent-bg: $secondary; // background for .table-striped (This becomes $table-striped-color in BS 5)
+$table-caption-color: $blackish;
 
 // Tooltip
 $tooltip-bg: $sul-tooltip-bg;
--- a/app/assets/stylesheets/modules/librarian-view.scss
+++ b/app/assets/stylesheets/modules/librarian-view.scss
@@ -1,8 +1,20 @@
-#marc_view .field {
-  word-break: break-all;
+.last-updated { 
+  margin-left: auto;
+  align-self: center;
+}
+.last-updated + .blacklight-modal-close {
+  margin-left: 0;
 }
 
-#folio-json-view summary {
-  @extend .h4;
-  @extend .my-3;
+#marc_view {
+  .field {
+    word-break: break-all;
+  }
+
+  .item-details {
+    caption {
+      caption-side: top;
+      background-color: $secondary;
+    }
+  }
 }
--- a/app/components/librarian/item_details/table_component.html.erb
+++ b/app/components/librarian/item_details/table_component.html.erb
@@ -1,0 +1,24 @@
+<details class="item-details mb-3">
+    <summary class="h3">Item details</summary>
+    <% locations.each do |location| %>
+      <table class="table">
+          <caption class="px-3 h4 mb-0"><strong>Permanent Location</strong> <%= location.label %></caption>
+          <thead>
+              <tr>
+                  <th>Effective Call Number</th>
+                  <th>Barcode</th>
+                  <th>Effective Location</th>
+              </tr>
+          </thead>
+          <tbody>
+            <% location.items.each do |item| %>
+                <tr>
+                  <td class="pl-5"><%= item.call_number %></td>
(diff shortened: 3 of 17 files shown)


cbeer:
Have you tried this with an electronic resource, bound-with, or something else without a FOLIO item?

With e.g. `a14136548`, I'm getting:

```
  
ActionView::Template::Error (undefined method `name' for nil:NilClass):
    30:           </div>
    31:         </div>
    32:       <%- end-%>
    33:     <%- end -%>
    34:   </details>
    35:   <%= render Librarian::ItemDetails::TableComponent.new(document: @document) %>
    36:   <%= render "folio_json_view" %>
  
app/components/librarian/item_details/table_component.html.erb:5:in `block in call'
app/components/librarian/item_details/table_component.html.erb:3:in `each'
app/components/librarian/item_details/table_component.html.erb:3:in `call'
```

dbranchini:
- In my prototype/comps, I used the same background color for the Permanent location heading as background color already used in FOLIO JSON, Holdings JSON and Circ Rules for consistency. It looks like you selected Fog (#DAD7CB) from the Identity Guide. Can you change that to Light Fog (#F4F4F4) for the new heading and all of the backgrounds for consistency and to align with the Identity Guide? 

- The detailed rows under the Permanent Location are indented by 40px. 

- It would also help to condense the spacing between rows. For comparison, on my comps, the rows are 24px (font is 14px) with spacing of 4px on both the top and bottom of the divider lines.

jcoyne:
@dbranchini that image is actually from the comps.  I made a mistake in which one I attached.

dbranchini:
@jcoyne , huh, that's strange. If you look at the original ticket, I have screenshots of the comps and you can see the differences that I mention above.

cbeer:
If there's no FOLIO items, I think we probably want to suppress the header too?

app/assets/stylesheets/bootstrap-variables.scss
```diff
@@ -53,7 +53,9 @@ $pagination-active-bg: $sul-pagination-active-bg;
 $pagination-active-border: $sul-pagination-active-bg;
 
 // Tables
-$table-accent-bg: $light-sandstone; // background for .table-striped
+$table-cell-padding: 0.5rem;
```
cbeer, line 56: Is this a bootstrap variable? Is this going to change the padding of every table in Searchworks (e.g. course reserves, _index_locations.html.erb)?

cbeer:
I'm not sure what call number we're using, but it seems like we're missing key information in the table.
Screenshot 2024-02-16 at 09 12 09

It looks like this might be just the base call number, and not the effective call number FOLIO (and we) normally construct, e.g.:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/6c8c060302c51008f7dbc7a598f6e4ffb79141ca/lib/folio_holding.rb#L32

app/assets/stylesheets/bootstrap-variables.scss
jcoyne: @cbeer yes, this is a bs variable. It changes every table. I asked @dbranchini about this and she agreed that this should be a site-wide change for consistency.

cbeer:
Thanks.

Screenshot 2024-02-20 at 09 05 39

PR #3942, opened 2024-02-13
Add a way to try interesting objects in development and test

jcoyne:
This allows us to get data in the same way it's expected in production, eliminating the problem of testing on non-realistic fixtures and not having fixtures that express all of the features.


marlo-longley (changes requested):
Nice thanks Justin. Could we add something about this to the "Getting data in development" section of the README? I see your comments in the file but it could be good to expose that this exists up front, since it's not automatically run. Something like?

"We have included fixtures files in the this repo that will be indexed with the searchworks:fixtures task. That task indexes what is in the fixtures/solr_documents directory. If you want to add other documents you can use the harvester which will download more json to that folder, and the run searchworks:fixtures again to reindex"

jcoyne:
@marlo-longley thank you for your review. I've added to the Readme. Good suggestion!

marlo-longley (approved):

PR #3949, opened 2024-02-15
Avoid using sass extend unnecessarily

jcoyne:
we can move away from sass if we stop using extend


marlo-longley (approved):

jcoyne:
@marlo-longley I think sass filled a need that we had 10 years ago, but now browsers natively support most of what we need.  Why continue to depend on a tool that we don't really need?

PR #3955, opened 2024-02-16
Remove unused style for btn-sul-toolbar and pagination

jcoyne:


marlo-longley (approved):
I did a visual review of these parts of the UI and all looks good.

PR #3959, opened 2024-02-16
Fix background for dropdown on small screens

jcoyne:
I believe this was broken in the BS4 upgrade

before
Screenshot 2024-02-16 at 11 23 15 AM

after
Screenshot 2024-02-16 at 11 22 48 AM


dnoneill:
I also don't think this great for accessibility. Due to the small font size I feel like contrast is an issue.

jcoyne:
Screenshot 2024-02-16 at 12 00 15 PM
Here's what it looks like on https://searchworks-morison.stanford.edu/view/14995717

jcoyne:
Will wait on this until @dbranchini can create a design.

PR #3960, opened 2024-02-16
Change stackmap controls into buttons

jcoyne:
Using an anchor isn't appropriate since they don't have an href.  Also increased the spacing to comply with WCAG AA recommended target size

Screenshot 2024-02-16 at 12 46 39 PM

### DIFF
--- a/app/assets/stylesheets/modules/stackmap.scss
+++ b/app/assets/stylesheets/modules/stackmap.scss
@@ -20,22 +20,8 @@
       overflow: auto;
       padding: 20px 0 10px 0;
 
-      .zoom-controls {
-        a {
-          border-bottom: 0;
-        }
-
-        .zoom-in, .zoom-out, .zoom-fit {
-          margin-right: 10px;
-        }
-      }
-
       .show-description {
         text-align: right;
-
-        a {
-          border-bottom: 0;
-        }
       }
     }
 
--- a/app/javascript/jquery.stackmap.js
+++ b/app/javascript/jquery.stackmap.js
@@ -27,7 +27,7 @@ import L from "leaflet";
         if (response.stat === "OK" && response.results.maps.length > 0) {
           plugContent(response);
         } else {
-          $container.html('').html('<p>No map available for this item</p>');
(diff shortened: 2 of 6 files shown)


cbeer:
It looks like this PR also changed the padding and alignment of those buttons. Here's what it used to look like:

Screenshot 2024-02-16 at 09 42 44


I'm not sure the old padding is a requirement, but they look pretty spaced out to me, but "Show text directions" should be aligned with the controls.

jcoyne:
@cbeer they are spaced more because this makes it easier to activate the controls without accidentally hitting an adjacent target. Is there a reason we wouldn't want this spacing?

cbeer:
I guess I'll defer to @dbranchini here too. I think the close(r) spacing makes it clear the 3 actions are grouped together.

jcoyne:
The old version is very challenging to use on mobile. I think it may not be WCAG AA compliant.
Old
Screenshot 2024-02-16 at 12 07 32 PM


New
Screenshot 2024-02-16 at 12 08 13 PM

marlo-longley:
I checked out this branch and can confirm it works as expected and matches the screenshots. I am happy to approve it code-wise, but this should probably have a final designer review.

dbranchini:
I looked up target size for buttons, icons, etc., and WCAG 2.2 AA says a minimum of 24px by 24px and WCAG 2.1 AAA says a minimum of 44px by 44px. Per these guidelines, spacing between buttons is less of an issue than the target size itself. This is true for all viewports and/or input devices. Where did we land on the mobile view, @jcoyne ? (Also note, I can't find any guidelines on this topic for WCAG 2.0 or 2.1 AA.)

dbranchini:
I reviewed these with Justin, and I'm good with his latest version. Justin added padding to comply with the target size to the existing icon buttons. It should be noticed that we may want to address link target size site-wide to comply with the guidelines listed above. It is not required at this time, but it will be in the future as Stanford increases its compliance versions.

marlo-longley (approved):

PR #3964, opened 2024-02-16
Stop using response from SearchService#fetch

jcoyne:
This is deprecated in Blacklight

### DIFF
--- a/app/components/view_type_dropdown_component.rb
+++ b/app/components/view_type_dropdown_component.rb
@@ -10,4 +10,9 @@ class ViewTypeDropdownComponent < Blacklight::Response::ViewTypeComponent
   def active_icon
     helpers.render_view_type_group_icon(@selected)
   end
+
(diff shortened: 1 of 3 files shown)


app/views/browse/index.html.erb
```diff
@@ -27,7 +27,11 @@
         <i class="fa fa-arrow-right" aria-hidden="true"></i><span class="sr-only">Next</span>
       <% end %>
       <div class='pull-right'>
-        <%= render 'catalog/view_type_group' %>
+        <%= render(Blacklight::Response::ViewTypeComponent.new(
```
cbeer, line 30: We seem to have our own `view_type_group` partial. Should we use it here, instead of that upstream one?

https://github.com/sul-dlss/SearchWorks/blob/main/app/views/catalog/_view_type_group.html.erb#L1-L5

jcoyne: I'm up for deleting that local partial if you agree. It's just calling the component, which I'm doing here.

jcoyne: Sure, done.

corylown (approved):

PR #3971, opened 2024-02-20
update results pages to check for missing fields

### LINKED ISSUE #3970: Suppress results metadata if there's no content to display
In Symphony, we were guaranteed a complete MARC leader so we could always display e.g. "Description: Book". With FOLIO, we're showing a blank section for many on-order records:

Image

dnoneill:
closes #3970 

Moves everything in `results-metadata-section` into a separate component. This section is used in 4 components:
1. document_metadata_component
2. search_result/collection/marc/metadata_component
3. search_result/collection/mods/metadata_component
4. search_result/item/marc/metadata_component
5. search_result/item/mods/metadata_component

Each of these components have separate fields they want to display so each component had a new field called fields added to their component rb files. The fields are then searched in the new component located in app/components/search_result/results_metadata_component.rb to see if they exist in the document. If none of the fields exist in the document the component doesn't render. If the document does have a field then the partial will check to see if that field exists. If it does then it will render that field into html.


jcoyne:
Consider breaking this section out to a ViewComponent.  Then you can use the `render?` method to determine whether or not anything should be drawn.  I try to avoid helper methods as these tend to pollute the view context (they have no encapsulation).

app/components/document_metadata_component.html.erb
```diff
@@ -1,7 +1,8 @@
 <div class="results-metadata-sections">
   <%= render "marc_search_results_document_fields", document: document %>
 
-  <dl class="dl-horizontal dl-invert results-metadata-section col-md-8">
+  <% has_metadata_fields = helpers.has_metadata_fields?(document, fields=['summary_data', 'extent', 'is_a_database?', 'index_parent_collections','index_links', 'index_parent_collections', 'collection_titles']) %>
+  <dl class="dl-horizontal dl-invert results-metadata-section col-md-8 <%= ' empty-metadata-section' unless has_metadata_fields %>">
```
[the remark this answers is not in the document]
dnoneill: I considered it but I thought it might more messy because of the extra tabbing.

app/components/search_result/results_metadata_component.html.erb
```diff
@@ -0,0 +1,55 @@
+<dl class="dl-horizontal dl-invert results-metadata-section col-md-8">
+    <% if present_fields['extent'] %>
+      <dt><%= document.extent_label %></dt>
+      <dd><%= document.extent %></dd>
+    <% end %>
+
+    <% if document.is_a_database? && present_fields['db_az_subject'] %>
+        <dt>Database topics</dt>
+        <dd>
+            <%= safe_join(document[:db_az_subject].map { |subject| helpers.link_to_database_search(subject) }, '; ') %>
```
dnoneill, line 10: There were two ways this was rendered in the old templates. This way and the way below. From what I can tell they do the same thing but if whoever is reviewing the PR could confirm that it would be appreciated.
```
<%= document[:db_az_subject].map do |subject| %>
            <% link_to(subject, search_catalog_path(f: {db_az_subject: [subject], document.format_key => ['Database']})) %>
          <% end.join('; ').html_safe %>
```

app/components/search_result/item/marc/metadata_component.rb
```diff
@@ -6,10 +6,11 @@ module Marc
       class MetadataComponent < Blacklight::Component
         def initialize(document:)
           @document = document
+          @fields = ['extent', 'db_az_subject', 'summary_data', 'index_links.finding_aid', 'index_parent_collections', 'collection_titles']
```
jcoyne, line 9: Can this be a constant?

app/components/search_result/collection/mods/metadata_component.rb
```diff
@@ -6,10 +6,12 @@ module Mods
       class MetadataComponent < Blacklight::Component
         def initialize(document:)
           @document = document
+          @fields = ['extent', 'summary_data']
```
jcoyne, line 9: Can this be a constant?

app/components/search_result/collection/marc/metadata_component.rb
```diff
@@ -6,10 +6,12 @@ module Marc
       class MetadataComponent < Blacklight::Component
         def initialize(document:)
           @document = document
+          @fields = ['extent', 'summary_data', 'index_links.finding_aid']
```
jcoyne, line 9: Can this be a constant?

jcoyne (changes requested):

app/components/document_metadata_component.html.erb
```diff
@@ -1,54 +1,7 @@
 <div class="results-metadata-sections">
```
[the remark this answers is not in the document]
dnoneill: I think that is a question for you. It got added in this PR.

app/components/search_result/collection/mods/metadata_component.rb
dnoneill: Yes but they are different across components so where/who are you suggesting it get defined?

app/components/search_result/results_metadata_component.rb
```diff
@@ -0,0 +1,55 @@
...
+    def present_fields
+      @present_fields = {}
+      @fields.each do |field|
+        if field == 'summary_data'
+          summary_data = SummaryDataComponent.new(document:)
+          if summary_data.render?
+            @present_fields[field] = summary_data
+          end
+        elsif field.include?(".")
+          # this is for fields like index_links.finding_aid
```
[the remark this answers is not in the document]
dnoneill: No but I wasn't sure if there might be the possibility of wanting to add another field with this same structure.

dnoneill:
made updates based on suggestions. Also moved summary_data into a solrdocument model. I also updated the app/components/search_result/results_metadata_component.html.erb template to pull data from the present_fields hash instead of the document model.

app/components/search_result/results_metadata_component.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    def present_fields
+      @present_fields = {}
+      @template_fields.each do |field|
+        field_value = document.public_send(field)
+        if field_value.present?
+          @present_fields[field] = field_value
+        end
+      end
+      @present_fields
```
jcoyne, line 19:
```suggestion
      @present_fields ||= @template_fields.to_h do |field|
        [field, document.public_send(field)]
      end.compact_blank
```

app/models/solr_document.rb
```diff
@@ -117,6 +117,19 @@ def to_semantic_values
...
+  def db_az_subject
+    self[:db_az_subject] if is_a_database?
+  end
+
+  def finding_aid
+    index_links&.finding_aid if has_finding_aid?
+  end
+
+  def summary_data
+    @summary_data ||= SummaryDataComponent.new(document: self)
```
jcoyne, line 129: I don't think we should initialize components in the model layer.  I think this might be a good one to move to a special case in `ResultsMetadataComponent`

app/components/search_result/results_metadata_component.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module SearchResult
+  class ResultsMetadataComponent < ViewComponent::Base
+    def initialize(document:, template_fields: [])
```
jcoyne, line 5: Do we need a default for template_fields?  Isn't it always required?
```suggestion
    def initialize(document:, template_fields:)
```

jcoyne (changes requested):

app/components/search_result/results_metadata_component.rb
dnoneill: This didn't work because Array.to_h doesn't work but I was able to do what you were indicating with @template_fields.map

jcoyne (approved):

PR #3974, opened 2024-02-20
Update "at the library" and "online" view for eresources 

### LINKED ISSUE #3944: Remove “At the library” section for coordinate online records; Add coordinate library to Available online card
## Problem Statement:
We have an inconsistent display for eResource records without a physical volume associated with the record. 

- If the eResource-only record is “owned” by SUL, we do NOT display the “At the Library” section. 
- If the eResource-only record is “owned” by Business, Lane Medical & Law, we DO display the “At the Library” section as well as the "Online" section. It displays correctly in the "Available online" card, but it displays a vague eResource with an item status of "Unknown" in the "At the Library" section which is confusing to users. The record is also added to the Library facet. 

### SUL - **_At the Library is not displayed_**:

Screenshot follows:

---

### Coordinate libraries - **_At the Library is displayed_**:
Libraries/records are: [Business: The emotional brain…](https://searchworks.stanford.edu/view/in00000011584); [Lane Medical: Brain communications](https://searchworks.stanford.edu/view/L354432); [Law: Breakthroughs in brain research… ](https://searchworks.stanford.edu/view/3305821) 

Screenshots from above records follow:
    

## Proposed Solution:
- Remove "At the library" heading and card from eResource-only records "owned" by a coordinate library.
- Maintain the relationship and facet (desired behavior by stakeholders).
- Add the "owning" coordinate library to the "Available online" card above the eResource link as shown on the comp.
- Specific styles:
-- Font size is the same as font used elsewhere - 14 px.
-- Spacing b/n lines - 4 px.
-- Coordinate library is a heavier font-weight - 600 or bold.

**NOTE: This is for eResource-only records. For example, it would not apply this record - https://searchworks.stanford.edu/view/15095727.**

See the screenshot and comp:

At the library/Unknown "Flow" on the Figma sidebar**
https://www.figma.com/proto/W3GIkPyGLeO5xMojmkZuTX/SearchWorks?page-id=0%3A1&type=design&node-id=308-1254&viewport=-62%2C-2146%2C0.25&t=iNqOsgsz6STHXHkf-1&scaling=min-zoom&starting-point-node-id=308%3A1254&show-proto-sidebar=1&mode=design

dnoneill:
closes #3944 

If the item is an eresource only item (there are no physical holdings), "At the library" will not render. This is done by updating suppressed? to work for all internet resources not just the SUL ones.

If the  view is an eresource only item there needs to be a library display in the online section if the library is not SUL. library_display looks to see if there are libraries associated with the document, if there are it checks if the record has physical copies or is a SUL location. If it isn't it will return the location name to display in the online section. See https://github.com/sul-dlss/searchWorks/issues/3944#issuecomment-1955173750 for screenshots.

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -120,6 +120,20 @@ class SolrDocument
     self[:img_info] || self[:file_id] || []
   end
 
+  # Checks holdings to see if there are any physical holdings
+  # If there aren't and the library isn't sul, return the library name
+  def eresources_library_display_name
+    library = holdings.libraries&.first
+    return unless library
+
+    has_physical_copies = holdings.libraries.any?(&:present?)
+    return if has_physical_copies
+
+    unless library.code == 'SUL'
+      library.name
+    end
+  end
+
   concerning :MarcOrganizationAndArrangement do
     def organization_and_arrangement
       @organization_and_arrangement ||= OrganizationAndArrangement.new(self)
--- a/spec/components/access_panels/at_the_library_component_spec.rb
+++ b/spec/components/access_panels/at_the_library_component_spec.rb
@@ -116,6 +116,10 @@ RSpec.describe AccessPanels::AtTheLibraryComponent, type: :component do
       render_inline(described_class.new(document:))
     end
 
+    it 'has solr document eresources_library_display_name of nil' do
+      expect(document.eresources_library_display_name).to be_nil
+    end
+
     it 'displays the MARC 590 as a bound with note (excluding subfield $c)' do
       expect(page).to have_css('.bound-with-note.note-highlight a', text: 'Copy 1 bound with v. 140')
       expect(page).to have_no_css('.bound-with-note.note-highlight', text: '55523 (parent record’s ckey)')
--- a/spec/models/solr_document_spec.rb
+++ b/spec/models/solr_document_spec.rb
@@ -124,6 +124,41 @@ RSpec.describe SolrDocument do
     end
   end
 
+  describe '#eresources_library_display_name' do
+    eresource_bus_library = { id: 30, item_display_struct:
+                [{ barcode: '9102385731231', callnumber: nil, current_location: nil,
+                   effective_location_id: '43c011f3-adeb-4c49-b357-06772aff11de',
+                   full_shelfkey: nil, home_location: 'BUS-ELECTRONIC', library: 'BUSINESS',
+                   permanent_location_code: 'BUS-ELECTRONIC', scheme: 'LC', type: 'ONLINE' }] }
+    context 'when it has holdings but no physical items' do
+      let(:solr_doc) { SolrDocument.new(eresource_bus_library) }
+
+      it 'is has a eresources_library_display_name' do
+        expect(solr_doc.eresources_library_display_name).to eq('Business Library')
+      end
+    end
+
+    context 'when it is an eresource and has physical copies' do
+      physical_location_hash = eresource_bus_library.deep_dup
+      physical_location_hash[:item_display_struct].push({ barcode: 123456, home_location: 'BUS', library: 'BUSINESS',
+                                                          permanent_location_code: 'BUS' })
+      let(:physical_location_doc) { SolrDocument.new(physical_location_hash) }
+
+      it 'has an eresource and physical location' do
+        expect(physical_location_doc.eresources_library_display_name).to be_nil
+      end
+    end
+
+    context 'when there are physical items and the library is SUL' do
+      solr_info = YAML.load_file(Rails.root.join('spec/fixtures/solr_documents/13553090.yml'))
+      let(:solr_doc) { SolrDocument.new(solr_info) }
+
+      it 'does not have a location display' do
+        expect(solr_doc.eresources_library_display_name).to be_nil
+      end
+    end
+  end
+
   describe 'EdsDocument' do
     let(:eds) { SolrDocument.new(eds_title: 'yup') }
     let(:non_eds) { SolrDocument.new }
--- a/app/components/access_panels/online_component.html.erb
+++ b/app/components/access_panels/online_component.html.erb
@@ -15,6 +15,9 @@
     <% end %>
 
     <% component.with_body do %>
+      <% if (sfx_links&.any? || links.any?) && document.eresources_library_display_name %>
+        <strong> <%= document.eresources_library_display_name %> </strong>
+      <% end %>
       <% if sfx_links&.any? %>
         <% sfx_url = sfx_links.first.href %>
         <% # Intentionally not encoding sfx_url here as the URLs from the catalog appear to already be encoded %>
--- a/app/views/catalog/_index_online_section.html.erb
+++ b/app/views/catalog/_index_online_section.html.erb
@@ -9,6 +9,9 @@
   </dt>
   <dd>
     <ul class="online-links" data-behavior="truncate-results-metadata-links">
+      <% if document.preferred_online_links.present? && document.eresources_library_display_name %>
+        <strong><%= document.eresources_library_display_name %></strong>
+      <% end %>
       <% document.preferred_online_links.each do |link| %>
         <li class="<%= 'stanford-only' if link.stanford_only? %>">
           <%= link.html.html_safe %>
--- a/lib/holdings/item.rb
+++ b/lib/holdings/item.rb
@@ -26,7 +26,7 @@ class Holdings
 
     def suppressed?
       @item_display.values.none?(&:present?) ||
-        (item_display[:library] == 'SUL' && internet_resource?)
+        internet_resource?
     end
 
     def browsable?


app/models/solr_document.rb
```diff
@@ -120,6 +120,18 @@ def file_ids
     self[:img_info] || self[:file_id] || []
   end
 
+  def location_display
+    @location = holdings.libraries&.first
```
jcoyne, line 124: Instead of an instance variable here, can you use a local?  How about using `library` instead as there are `Locations` within a library, so using the name `location` could be confusing.   Would it makes more sense to look at `folio_holdings` here as `holdings` also has `mhlds`

jcoyne (changes requested):

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,33 @@
     end
   end
 
+  describe 'check ersources location_display' do
+    solr_info = YAML.load_file(Rails.root.join('spec/fixtures/solr_documents/30.yml'))
```
jcoyne, line 128: We usually reserve fixtures for feature tests.  If we don't have a feature test here, it might just be easier to instantiate a `SolrDocument` inline here and skip the fixture.  Adding fixtures is something we try to be judicious with since they make the repository big and it's a bunch of indirection when you're trying to figure out a test.

jcoyne (changes requested):

app/models/solr_document.rb
dnoneill: Where are you thinking we should using folio_holdings? In this line and has_physical_copies or just this line?

spec/components/access_panels/at_the_library_component_spec.rb
```diff
@@ -116,6 +116,10 @@
       render_inline(described_class.new(document:))
     end
 
+    it 'has solr document location_display of nil' do
+      expect(document.location_display).to be_nil
```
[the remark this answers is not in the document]
dnoneill: The function is used in the component and that data used there is another good place to make sure everything is working correctly. The solr document has a physical location so that is why I put it in there.

app/models/solr_document.rb
```diff
@@ -120,6 +120,20 @@ def file_ids
     self[:img_info] || self[:file_id] || []
   end
 
+  # Checks holdings to see if there are any physical holdings
+  # If there aren't and the library isn't sul, return the library name
+  def library_display
```
cbeer, line 125: I think I'm struggling with the method name a little bit; maybe a super explicit name would help?

```suggestion
  def eresources_library_display_name
```

cbeer:
Thanks -- other than the small method naming quibble, I'm 👍.  It took me a little while to figure out that all SUL branch libraries keep their electronic resources under `SUL-ELECTRONIC` (vs `SCI-ELECTRONIC`, `MAR-ELECTRONIC`, etc). I briefly wondered whether we'd have to check the FOLIO campus to catch all those, but that feels like a "cross that bridge" kind of problem 🤷‍♂️

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,36 @@
...
+                   effective_location_id: '43c011f3-adeb-4c49-b357-06772aff11de',
+                   full_shelfkey: nil, home_location: 'BUS-ELECTRONIC', library: 'BUSINESS',
+                   permanent_location_code: 'BUS-ELECTRONIC', scheme: 'LC', type: 'ONLINE' }] }
+    physical_location_hash = solr_info.deep_dup
+    physical_location_hash[:item_display_struct].push({ barcode: 123456, home_location: 'BUS', library: 'BUSINESS',
+                                                        permanent_location_code: 'BUS' })
+    let(:solr_doc) { SolrDocument.new(solr_info) }
+    let(:physical_location_doc) { SolrDocument.new(physical_location_hash) }
+
+    it 'is has a library_display' do
```
jcoyne, line 139: can you make this one context block for `when it has holdings but no physical items`

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,36 @@
...
+      expect(solr_doc.library_display).to eq('Business Library')
+    end
+
+    it 'has an eresource and phsyical location' do
```
jcoyne, line 143: and this should be a context for `when it has physical copies`

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,36 @@
...
+      expect(physical_location_doc.library_display).to be_nil
+    end
+  end
+
+  describe 'check sul ersource library_display' do
```
jcoyne, line 148: And maybe you can come up with a context for this one? Is this testing the same paths as the test above?  If so, do we need it?

jcoyne (changes requested):
Looks good aside from a few testing suggestions and the method name as Chris pointed out.

spec/models/solr_document_spec.rb
dnoneill: It is testing one of the fixtures which is a sul eresouce. I came up with a context for this one too and added into the describe '#eresources_library_display_name' so they are all grouped together.

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,42 @@
...
+      end
+    end
+
+    context 'when it has an eresouce and physical copies' do
+      physical_location_hash = eresource_bus_library.deep_dup
+      physical_location_hash[:item_display_struct].push({ barcode: 123456, home_location: 'BUS', library: 'BUSINESS',
+                                                          permanent_location_code: 'BUS' })
+      let(:physical_location_doc) { SolrDocument.new(physical_location_hash) }
+
+      it 'has an eresource and phsyical location' do
```
jcoyne, line 148:
```suggestion
      it 'has an eresource and physical location' do
```

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,42 @@
...
+
+    context 'when it has holdings but no phsyical items' do
+      let(:solr_doc) { SolrDocument.new(eresource_bus_library) }
+
+      it 'is has a eresources_library_display_name' do
+        expect(solr_doc.eresources_library_display_name).to eq('Business Library')
+      end
+    end
+
+    context 'when it has an eresouce and physical copies' do
```
jcoyne, line 142:
```suggestion
    context 'when it is an eresource and has physical copies' do
```

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,42 @@
...
 
+  describe '#eresources_library_display_name' do
+    eresource_bus_library = { id: 30, item_display_struct:
+                [{ barcode: '9102385731231', callnumber: nil, current_location: nil,
+                   effective_location_id: '43c011f3-adeb-4c49-b357-06772aff11de',
+                   full_shelfkey: nil, home_location: 'BUS-ELECTRONIC', library: 'BUSINESS',
+                   permanent_location_code: 'BUS-ELECTRONIC', scheme: 'LC', type: 'ONLINE' }] }
+
+    context 'when it has holdings but no phsyical items' do
+      let(:solr_doc) { SolrDocument.new(eresource_bus_library) }
```
jcoyne, line 135:
```suggestion
    context 'when it has holdings but no physical items' do
      eresource_bus_library = { id: 30, item_display_struct:
                [{ barcode: '9102385731231', callnumber: nil, current_location: nil,
                   effective_location_id: '43c011f3-adeb-4c49-b357-06772aff11de',
                   full_shelfkey: nil, home_location: 'BUS-ELECTRONIC', library: 'BUSINESS',
                   permanent_location_code: 'BUS-ELECTRONIC', scheme: 'LC', type: 'ONLINE' }] }
      let(:solr_doc) { SolrDocument.new(eresource_bus_library) }
```

spec/models/solr_document_spec.rb
```diff
@@ -124,6 +124,42 @@
...
+      physical_location_hash[:item_display_struct].push({ barcode: 123456, home_location: 'BUS', library: 'BUSINESS',
+                                                          permanent_location_code: 'BUS' })
+      let(:physical_location_doc) { SolrDocument.new(physical_location_hash) }
+
+      it 'has an eresource and phsyical location' do
+        expect(physical_location_doc.eresources_library_display_name).to be_nil
+      end
+    end
+
+    context 'when it is a SUL-ERESOURCE' do
```
jcoyne, line 153: I'm still having trouble understanding which case this is.  
```suggestion
    context 'when there are physical items and the library is SUL' do
```

app/models/solr_document.rb
```diff
@@ -120,6 +120,20 @@ def file_ids
     self[:img_info] || self[:file_id] || []
   end
 
+  # Checks holdings to see if there are any physical holdings
+  # If there aren't and the library isn't sul, return the library name
+  def eresources_library_display_name
+    library = holdings.libraries&.first
```
jcoyne, line 126: Couldn't a mhld cause problems here?  the Holding::Items are derived from `item_display_struct`, which is created from the combination of holding statements and items. Its possible to have holding statements and no items, which might cause a problem here.  I think it might be better to look at `SolrDocument#folio_items` so that you don't have to worry about holding statements.

jcoyne (changes requested):

app/models/solr_document.rb
cbeer: I think `holdings` is the right place? We're going down this path because these are e-resources without an associated item in FOLIO, just a holdings record.

jcoyne: What about `folio_holdings` then?  I got the impression we're trying to move away from `item_display_struct`.   I'm also wondering if this method ought to move to OnlineComponent, as I don't see much use for this outside of that context.

dnoneill: I am just adding this so I don't forget Monday. I tested folio_holdings when Justin first commented on it, it will work with `folio_holdings.first.effective_location.library`. 

Also I am wondering if the mhld is an issue would we need to update this bit of code as well? Or is this not a problem in ezproxy? 
https://github.com/sul-dlss/SearchWorks/blob/afd93f43ee03dfbab769fb67da19d1ccde619af1/app/models/links/ezproxy.rb#L55-L57

spec/models/solr_document_spec.rb
dnoneill: Tried this. Broke the `when it is an eresource and has physical copies` test. I can do this and copy and paste the data, but that just makes the file longer.

app/models/solr_document.rb
jcoyne: Oh, yeah, that's probably something to look at too, but not for this PR.

jcoyne: I've made https://github.com/sul-dlss/SearchWorks/pull/4028 for that one

cbeer: > I got the impression we're trying to move away from item_display_struct

I don't know where you got that impression. I don't think that's the case in general.

jcoyne: Is the idea to keep two parallel data structures in Solr? I thought the idea was to migrate to the folio based one. Maybe this could be covered post standup next week. I don't think we have a shared vision of what we're working toward.

cbeer:
It seems like the conversation on this PR is over. I'm going to merge it  and if there's any outstanding TODOs, we can fix them up after.

PR #3982, opened 2024-02-21
Remove Symphony reserves locations

jcoyne:
None of these codes are used in FOLIO


dnoneill:
Not really. I didn't remove any of the reserves stuff because I didn't realize they were symphony related.

dnoneill (approved):

PR #3985, opened 2024-02-21
Remove unused code

jcoyne:
This is never used because on the search results marcxml is not in fl. Revert part of https://github.com/sul-dlss/SearchWorks/commit/bee03acebd8cc7dd32daabc6684ba18ee7596f8d

### DIFF
--- a/app/views/catalog/_index_location.html.erb
+++ b/app/views/catalog/_index_location.html.erb
@@ -8,20 +8,7 @@
           <th scope='col' class='col-xs-6'>Status</th>
         </tr>
       </thead>
-      <% if document.respond_to?(:to_marc) && document.bound_with? %>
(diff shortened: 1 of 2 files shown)


corylown:
@jcoyne @cbeer is this a feature we want (display bound-with messaging on search results)? I think this could be updated to work by changing the condition to something like `if library.items.any?(&:bound_with?)`.

cbeer:
I think we probably want this feature, at least until we do our bound-with work. It's definitely funny there's no indication the record is bound-with on the search results:
 
Screenshot 2024-02-21 at 07 39 58

jcoyne:
See https://github.com/sul-dlss/SearchWorks/pull/3991 for another alternative that uses @corylown's suggestion to restore the functionality.

PR #3987, opened 2024-02-21
Remove duplicate test

jcoyne:

### DIFF
--- a/spec/lib/holdings_spec.rb
+++ b/spec/lib/holdings_spec.rb
@@ -26,20 +26,14 @@ RSpec.describe Holdings do
       ).items
     )
   }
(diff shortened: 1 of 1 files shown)


spec/lib/holdings_spec.rb
```diff
@@ -26,20 +26,14 @@
       ).items
     )
   }
-  let(:no_holdings) { Holdings.new([]) }
+  let(:no_holdings) { Holdings.new }
 
   describe "#present?" do
-    let(:blank_callnumber) {
-      Holdings.new([])
```
jcoyne, line 33: This is the same as `no_holdings`

corylown (approved):

PR #3988, opened 2024-02-21
Simplify the LocationRequestLinkPolicy by looking at Folio data directly

jcoyne:


hudajkhan:
Holding off on changes to holdings (for the time being) until we have further reviewed how SearchWorks uses item_display_struct (derived from folio holdings, items, boundwiths, mholds, etc.) vs. folio holdings.

PR #3990, opened 2024-02-21
SEE-OTHER is no longer used in the index, so remove it from the code

jcoyne:

### DIFF
--- a/lib/holdings/location.rb
+++ b/lib/holdings/location.rb
@@ -41,7 +41,7 @@ class Holdings
     end
 
     def present_on_index?
-      any_items? || any_index_mhlds? || @code == 'SEE-OTHER'
+      any_items? || any_index_mhlds?
     end
 
     def sort
--- a/spec/components/access_panels/at_the_library_component_spec.rb
+++ b/spec/components/access_panels/at_the_library_component_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe AccessPanels::AtTheLibraryComponent, type: :component do
       SolrDocument.new(
         id: '1234',
         item_display_struct: [
-          { barcode: '1234', library: 'SAL3', effective_permanent_location_code: 'SEE-OTHER', callnumber: 'ABC 123' }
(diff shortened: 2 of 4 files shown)


cbeer:
In order to review this, it'd be a big help if you provided context for why we no longer need SEE-OTHER. Has it been replaced by something else, or is this a migration problem that we should be addressing? 

In particular, I still see location codes for e.g. `SAL3-SEE-OTHER` and `GRE-SEE-OTHER`. Do they no longer need this special handling to display ok?

jcoyne:
@cbeer this is about the usage of `SEE-OTHER` (and not `*-SEE-OTHER`).  This is not a valid Folio location code and is not used in `item_display`

cbeer:
I understand that. I'm asking whether those  `*-SEE-OTHER` locations are something we should be handling in this code (and failed to catch in the migration).

jcoyne:
@cbeer I don't know.  I don't understand what we would do with those.  They don't show up for most bound-with things I'm seeing in `item_display`, I'm not sure they should.

cbeer (changes requested):
Ok. Then I'd like to flag this for further analysis. It's quick to just throw away the code, but I think we need to figure out whether this is the correct solution or if there was an earlier oversight.

dnoneill:
I think some of this is out of date because of #3892, I think some of this should be pulled out if see-other is a symphony code only.

hudajkhan:
Questions to review:
a) What are analyzed serials and how are they represented in FOLIO data?
Analyzed serials appear similar to bound-withs because separate records have relationships with each other, but metadata folks point out that these two constructs are different.  
b) Will we need SEE-OTHER to handle analyzed serials? (In which case, we don't want to remove SEE-OTHER) 
c) Even if we don't use SEE-OTHER as a code, will we need similar behavior in other places?

See example: https://searchworks.stanford.edu/view/53602

hudajkhan:
Discussion resulted in new issue: https://github.com/orgs/sul-dlss/projects/60?pane=issue&itemId=54534154

corylown (approved):

PR #3993, opened 2024-02-21
Remove unused as_json methods

jcoyne:
We used to use these for availability in #1068, but we no longer have this approach


cbeer:
Could you update the commit + PR title to say `as_json`?

Also, I think `Holdings::Status#as_json` can also go away, right?

PR #3994, opened 2024-02-21
don't update live look up status if there is an avaliability class

dnoneill:
Based on Sarah's comments in [#1026](https://github.com/sul-dlss/searchworks_traject_indexer/issues/1026)

Doesn't allow live updating of status if there is an "In_process" availability class, otherwise it allows for updating of status.

Works on records like

### DIFF
--- a/app/components/access_panels/location_item_component.rb
+++ b/app/components/access_panels/location_item_component.rb
@@ -25,6 +25,11 @@ module AccessPanels
       item.temporary_location.name
     end
 
+    def has_in_process_availability_class?
+      availability_class = item.effective_location&.details&.dig('availabilityClass')
+      availability_class.present? && availability_class == 'In_process'
+    end
+
     def render_item_details?
       !item.suppressed?
     end
--- a/app/components/access_panels/location_item_component.html.erb
+++ b/app/components/access_panels/location_item_component.html.erb
@@ -10,7 +10,7 @@
     </td>
   <% end %>
   <% if render_item_details? %>
-    <td class="item-availability" data-live-lookup-id="<%= item.live_lookup_instance_id %>" data-status-target=".availability-icon" data-item-id="<%= item.live_lookup_item_id %>" <%= "data-request-url='#{helpers.request_url(document, library: item.library, location: item.home_location, barcode: item.barcode)}'".html_safe if render_real_time_availability_request_link? %>>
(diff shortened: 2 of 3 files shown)


app/components/access_panels/location_item_component.rb
```diff
@@ -18,11 +18,15 @@ def availability_text
...
 
     def current_location_text
-      return if item.effective_location&.details&.key?('availabilityClass') || item.effective_location&.details&.key?('searchworksTreatTemporaryLocationAsPermanentLocation')
+      return if has_avaliability_class? || item.effective_location&.details&.key?('searchworksTreatTemporaryLocationAsPermanentLocation')
 
       item.current_location.name
     end
 
+    def has_avaliability_class?
+      item.effective_location&.details&.key?('availabilityClass')
```
cbeer, line 27: Is this any availability class data, or should we limit it just to the in-process  value? I think we still want to do live lookups for e.g. stuff with "off-site", no?

dnoneill: I was understandings that if it had an availability class we did not live lookup. @saseestone do you know?

cbeer: The availability class values are:

```
Available
Unavailable
In_process
Offsite
```
(From https://github.com/sul-dlss/folio-graphql/blob/501e6a891e8f6b586ab4bfc8ded6e1ca2f8148e6/src/schema.graphql#L4365-L4370)

Offsite is not like the others and requires special handling (because it could be paged to campus and be checked out/lost/etc)

app/components/access_panels/location_item_component.rb
```diff
@@ -25,6 +25,12 @@ def temporary_location_text
       item.temporary_location.name
     end
 
+    def has_avaliability_class?
+      # exclude offsite since it is not like other avaliability classes
```
jcoyne, line 29: Do we know why this is?

app/components/access_panels/location_item_component.rb
```diff
@@ -25,6 +25,12 @@ def temporary_location_text
       item.temporary_location.name
     end
 
+    def has_avaliability_class?
```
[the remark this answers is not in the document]
jcoyne: Is anything using this method?

jcoyne (approved):

PR #3999, opened 2024-02-22
Remove noop location_link method

jcoyne:


cbeer:
I'm not sure we're ready to remove this method. While we're no longer using it to link out to Lane's catalog, they're using the location detail `discoveryDisplayUrl` (e.g. https://lane.stanford.edu/view/bib/312369 / https://searchworks.stanford.edu/view/L312369) to drive very similar behavior and it seems likely to me we'll resume using this behavior in the not-too-distant future.

jcoyne:
@cbeer How about now?

PR #4005, opened 2024-02-22
Remove current_location code check for on-orderness

cbeer:
Part of #3892; folio_status and the current location code are both populated consistently:
https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/folio_record.rb#L183-L213


jcoyne:
Do you want to remove "ON-ORDER" from item_request_link_component_spec.rb too?

PR #4011, opened 2024-02-22
Use temporary_location_code (FOLIO) instead of current_location (Symphony) consistently

cbeer:
This PR uses the FOLIO name (`temporary_location_code`) instead of the Symphony `current_location` (which was a mash-up of FOLIO's temporary location and legacy Symphony status codes). With https://github.com/sul-dlss/searchworks_traject_indexer/pull/1326, we're no longer including the status codes in this field, so `temporary_location_code` and `current_location` should contain identical data.

`home_location` (+ FOLIO's `permanent_location_cod`) are not quite the same, so this PR only handles the `temporary_location_cod`


app/components/access_panels/location_item_component.rb
```diff
@@ -14,13 +14,13 @@ def initialize(item:, document:, classes: nil, render_item_level_request_link: t
...
 
     def availability_text
-      item.status.status_text unless current_location_text
+      item.status.status_text unless temporary_location_text
     end
 
-    def current_location_text
-      return if item.effective_location&.details&.key?('availabilityClass') || item.effective_location&.details&.key?('searchworksTreatTemporaryLocationAsPermanentLocation')
+    def temporary_location_text
+      return if item.effective_location&.details&.key?('availabilityClass') || item.effective_location&.details&.key?('searchworksTreatTemporaryLocationAsPermanentLocation') || item.home_location == item.temporary_location_code
```
jcoyne, line 21: This line is a bit long. Would you mind adding a linebreak or two?

jcoyne (changes requested):

jcoyne (approved):

PR #4014, opened 2024-02-22
Encapsulate location details

jcoyne:
This ensures that only the Folio::Location class needs to understand the serialization for Location.  This keeps us from spreading the structure of the serialzied data throughout the codebase. Instead of 4 classes that depend on the structure of the data, now we have only one.

### DIFF
--- a/app/components/access_panels/location_item_component.rb
+++ b/app/components/access_panels/location_item_component.rb
@@ -18,7 +18,7 @@ module AccessPanels
     end
 
     def current_location_text
-      return if item.effective_location&.details&.key?('availabilityClass') || item.effective_location&.details&.key?('searchworksTreatTemporaryLocationAsPermanentLocation')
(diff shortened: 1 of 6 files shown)


cbeer:
I'm +1 for creating `Folio::Location#details`, but pretty -0 on creating methods for each particular detail (kind of agreeing with Marlo's comment on https://github.com/sul-dlss/SearchWorks/pull/3965#discussion_r1499763316). If we had additional logic, maybe I'd see the point but this seems like unnecessary indirection 🤷‍♂️

jcoyne:
If I want to know what is allowed in `details`, where would I start looking?  Currently I have to look all over the codebase.  While this does have some indirection, I think the centralization of knowledge about `details` is far more helpful to the code maintainer.

hudajkhan:
Based on discussion, we won't be going ahead with this particular set of changes.

PR #4015, opened 2024-02-22
We don't need to have different translations for different libraries

jcoyne:
We only ever use the default


cbeer:
I think this is another thing I wouldn't throw out yet. We've had various requests over the years and this is a useful seam to have 🤷‍♂️

PR #4018, opened 2024-02-22
override blacklight @current_bookmarks for browse view

### LINKED ISSUE #3966: Bookmarks do not work in browse related items
Go here:
https://searchworks.stanford.edu/browse?start=in00000039227&view=gallery

Then check some bookmarks.  Then refresh the page.  Items that you have bookmarked do not show as checked.

I believe this is because we're doing a fetch in BrowseController, but then using render_document_index, which expects a `@response` object to have all of the items on the page.    This is how the current bookmarks are loaded:
https://github.com/projectblacklight/blacklight/blob/release-7.x/app/helpers/blacklight/catalog_helper_behavior.rb#L292-L294

It looks like this will continue to be a problem in BL8 https://github.com/projectblacklight/blacklight/blob/d0505ed1da1d81e6b98350e6a6fae087cd27b9d5/app/helpers/blacklight/document_helper_behavior.rb#L45

dnoneill:
closes #3966

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -59,5 +59,6 @@ class BrowseController < ApplicationController
         search_service:
       )
     end
+    @current_bookmarks = current_or_guest_user.bookmarks
   end
 end
--- a/spec/features/callnumber_browse_spec.rb
+++ b/spec/features/callnumber_browse_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe 'Callnumber Browse', :js do
   end
 
   describe 'full browse view' do
-    it 'is successful' do
+    it 'is successful and checkboxes stay checked after refresh' do
       visit solr_document_path('1')
 
       within '.record-browse-nearby' do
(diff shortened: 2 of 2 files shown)


jcoyne (changes requested):
This doesn't seem to override any methods. Can you update the title of the PR and the commit?

I think the reason this broke in the first place is because we don't have a feature test for bookmarks on the browse page.  Can you write one so that we don't accidentally break it again?

dnoneill:
I set `@current_bookmarks = current_or_guest_user.bookmarks` does it need to be `@current_bookmarks = current_or_guest_user.bookmarks.where(record_type: 'catalog')` both work locally. Which is better?

It looks like there was a test in place and it bookmarking is working on production. If you check something on prod it does get added to selections and it does stay in selections after refresh it just doesn't understand that that item is in selections so the checkbox isn't checked. I have added a test to make sure that after a refresh those checkboxes stay sticky.

jcoyne (approved):
Looks great! Thank you.

PR #4023, opened 2024-02-23
Rename home_location (Symphony) to effective_permanent_location_code (FOLIO-esque)

cbeer:
I'm not attached to the name `effective_permanent_location_code`, and welcome better suggestions 🤷‍♂️  

Part of https://github.com/sul-dlss/searchworks_traject_indexer/issues/1348


jcoyne:
@cbeer this needs a rebase.

lib/holdings/item.rb
```diff
@@ -47,9 +47,11 @@ def library
...
     end
 
-    # @return [String]
-    def home_location
-      item_display[:home_location]
+    # This is generally the item's permanent location, but it can also be
+    # the temporary location if the location detail searchworksTreatTemporaryLocationAsPermanentLocation
+    # is set to true.
+    def effective_permanent_location_code
+      item_display[:effective_permanent_location_code] || item_display[:home_location]
```
jcoyne, line 54: Is this conditional temporary, or something we'll want to keep for a long time?

cbeer: Just until we do a full reindex with the new name.

jcoyne (approved):

PR #4028, opened 2024-02-23
Look at folio holdings for the library for proxy links

jcoyne:
This helps decouple from item_display_struct


cbeer:
I'm not sure why we want to decouple from `item_display_struct`. When we're rendering stuff on the search results page, I think we'd like to use it, because it's (presumably) a lighter-weight data structure than the FOLIO holdings.

PR #4043, opened 2024-02-27
Remove unused css class

jcoyne:


dnoneill:
@jcoyne do we really want to get rid of this? It is a blacklight class that "make tables more compact by cutting cell padding in half"

jcoyne:
Thanks, I wasn't seeing it applying any styles, but now I see its' just applying to child elements.

PR #4046, opened 2024-02-27
Don't draw empty locations on the search results page

### LINKED ISSUE #4042: Don't display the library on the search results if there's only MHLD statements
Some records have MHLD statements for locations without any physical holdings, e.g. https://searchworks.stanford.edu/?search_field=search&q=1463888

It is currently displaying on the search results page like:
Image

We should suppress SAL1+2 on the search results page because there are no items (and we don't show the MHLD statements there).  We should continue to show SAL1+2 on the item's page because we show the MHLD statements.

jcoyne:
This was ocurring because there were MHLDs with only 'library_has'

Fixes #4042

### DIFF
--- a/app/components/search_result/location_component.rb
+++ b/app/components/search_result/location_component.rb
@@ -12,6 +12,10 @@ module SearchResult
 
     delegate :code, to: :library, prefix: true
 
+    def render?
+      locations.any?
+    end
+
     def link_to_record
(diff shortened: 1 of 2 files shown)


marlo-longley (approved):
Screenshot 2024-02-29 at 12 46 48 PM

Looks good.

PR #4053, opened 2024-02-28
update data-controller live-lookup to fire once per instance record

### LINKED ISSUE #4050: Due date displaying twice on index view
https://searchworks.stanford.edu/?search_field=search&q=+Elizabethan+poetry%3B+modern+essays+in+criticism+2395828

Screenshot 2024-02-28 at 10 07 56 AM

dnoneill:
Due to the data-controller="live-lookup" being in each table for each location, it would fire the live_lookup js for each library location. Since rtac looks up an instance record we only need to fire the live lookup once record. This PR moves the data-controller="live-lookup" out of the loop into a wrapper.

closes #4050

### DIFF
--- a/app/views/catalog/_accordion_section_library.html.erb
+++ b/app/views/catalog/_accordion_section_library.html.erb
@@ -16,7 +16,7 @@
         <%= snippet %>
       </span>
     </div>
-    <div class="details <%= id %>-location" aria-expanded="false">
+    <div class="details <%= id %>-location" aria-expanded="false" data-controller="live-lookup" data-live-lookup-url-value="<%= availability_index_path %>">
       <%= render :partial => "catalog/index_location", :locals => { :document => document } %>
     </div>
   </div>
--- a/app/components/search_result/location_component.html.erb
+++ b/app/components/search_result/location_component.html.erb
@@ -1,4 +1,4 @@
-<table class="table table-sm availability" data-controller="live-lookup" data-live-lookup-url-value="<%= availability_index_path %>">
+<table class="table table-sm availability">
   <caption class='sr-only'>Status of items at <%= library.name %></caption>
   <thead>
     <tr class='active'>


app/views/catalog/_index_location.html.erb
```diff
@@ -1,5 +1,7 @@
 <% if document.holdings.present? %>
-  <% document.holdings.libraries.select(&:present?).each do |library| %>
-    <%= render SearchResult::LocationComponent.new(library:, document:) %>
-  <% end %>
+  <div data-controller="live-lookup" data-live-lookup-url-value="<%= availability_index_path %>">
```
jcoyne, line 2: Can we move this up to the existing div in `_accordion_section_library.html.erb`?

dnoneill: Are you talking about that line or the entire template? 
I wouldn't move that line outside that if statement. We could move the line plus the if statement to _acccordian_section_library. I don't think that would cause any issues.

In terms of the template I think that is a lot of refactoring and moving things around which I think should probably be a separate PR.

jcoyne: Moving it outside of the component means you've already circumvented the `render?` statement, so perhaps it needs to be in there.

jcoyne: After having another look it seems like the javascript is querying for `[data-live-lookup-id]`, so it should just submit an empty query if there are no holdings present.  I think this div (just this line) can move up to `_accordion_section_library.html.erb`

dnoneill: I am not sure why we would want to run JS if we don't have to.

jcoyne: I'm pretty sure that if `_accordion_section_library` is rendering anything, then it's going to check availability.

dnoneill: good point. I can't think of any record that would have not have document.holdings if it is being rendered by accordion_section_library and it doesn't look like index_location is being used by anything else.

app/views/catalog/_accordion_section_library.html.erb
```diff
@@ -17,7 +17,9 @@
       </span>
     </div>
     <div class="details <%= id %>-location" aria-expanded="false">
-      <%= render :partial => "catalog/index_location", :locals => { :document => document } %>
+      <div data-controller="live-lookup" data-live-lookup-url-value="<%= availability_index_path %>">
```
jcoyne, line 20: Can these `data-*` properties move to the parent div? Then we don't need a new div.

jcoyne (approved):

PR #4054, opened 2024-02-28
Only show the bound-with message in the appropriate library

### LINKED ISSUE #4049: Only show bound-with notice on locations that have a bound-with holding
I think we're going to revisit this as part of our upcoming work on bound-with. However, it looks like our check for bound-with items on the index view isn't appropriately scoped to the location so we incorrectly repeat the bound-with notice for each location:

Example record: https://searchworks.stanford.edu/?search_field=search&q=id%3A5684521

We should only be showing the bound-with notice/link for the SAL3 location and not Spec.
Screenshot 2024-02-28 at 9 36 30 AM

I think the fix is going to be to change the [SearchResult::LocationComponent#bound_with?](https://github.com/sul-dlss/SearchWorks/blob/02bccb8ea5bedf91eb4454ed957d7c60ea34689c/app/components/search_result/location_component.rb#L20) method to check only for bound-withs within the location and not all the items in the document.

jcoyne:
Fixes #4049

### DIFF
--- a/spec/components/search_result/location_component_spec.rb
+++ b/spec/components/search_result/location_component_spec.rb
@@ -103,7 +103,7 @@ RSpec.describe SearchResult::LocationComponent, type: :component do
         SolrDocument.new(
           id: '123',
           item_display_struct: [
-            { barcode: '123', library: 'GREEN', effective_permanent_location_code: 'STACKS', callnumber: 'ABC 123' }
+            { barcode: '123', library: 'GREEN', effective_permanent_location_code: 'GRE-STACKS', callnumber: 'ABC 123' }
           ],
           mhld_display: ['GREEN -|- GRE-CURRENTPER -|- -|- library has -|-']
         )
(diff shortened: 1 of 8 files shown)


spec/components/search_result/location_component_spec.rb
```diff
@@ -101,4 +101,61 @@
...
+              ]
+            }
+          ]
+        )
+      end
+
+      context 'when looking at SAL3' do
+        let(:library) { document.holdings.libraries.first }
+
+        it "doesn't show" do
```
corylown, line 147:
```suggestion
        it "does show" do
```

corylown (changes requested):
Confirmed that it fixes #4049. One suggested change for a test description.

spec/components/search_result/location_component_spec.rb
jcoyne: Good eye! I've fixed this.

corylown (approved):

PR #4061, opened 2024-03-01
Remove old courses from the cache

jcoyne:
Otherwise the cache will grow without bound


app/services/folio_client.rb
```diff
@@ -54,7 +54,13 @@ def circulation_rules
 
   # https://s3.amazonaws.com/foliodocs/api/mod-courses/r/courses.html#coursereserves_courses_get
   def courses
-    get_json('/coursereserves/courses', params: { limit: 2_147_483_647 }).fetch('courses', []).sort_by { |x| x['id'] }
+    # Querying with CQL on courseListingObject properties doesn't seem to work, e.g.
+    # get_json('/coursereserves/courses', params: { query: 'courseListingObject.termId=="d5d04559-a7b1-4e4b-a95d-995e0be98b61"'})
+    # So we'll post process instead.
+    get_json('/coursereserves/courses', params: { limit: 2_147_483_647 })
+      .fetch('courses', [])
+      .filter { |course| Time.parse(course.dig('courseListingObject', 'termObject', 'startDate')).future? } # rubocop:disable Rails/TimeZone
```
cbeer, line 62: Wouldn't this exclude the current term too?

jcoyne: Thanks, I will change it to endDate

PR #4065, opened 2024-03-01
Maintain 12 columns per row in the layout

jcoyne:
This fixes the problem where the subheader was not aligning with the toolbar.

Before:
Screenshot 2024-03-01 at 2 19 05 PM

After:
Screenshot 2024-03-01 at 2 19 16 PM


dnoneill (dismissed):
Around 952 px I think the margins look a little weird. I think we need some clean up for larger screens to match what is on prod.

On this branch
Screenshot 2024-03-11 at 10 48 04 AM

On prod
Screenshot 2024-03-11 at 10 48 11 AM

jcoyne:
@dnoneill I don't understand what you are asking me to do.

jcoyne:
>  margins for the record view on a screen 952px don't look even

I think that's the change that was necessary so that the mini-menu doesn't collide with the stuff in the middle. I'm not sure how to have it both ways.

dnoneill:
What if the margin is changed to margin: 0px calc(-30px * .5 + 35px); So that you have equal margins on either side?

I am including some phone screenshots to compare.  

This branch:


With margin set to  margin: 0px calc(-30px * .5 + 35px);
Screenshot 2024-03-11 at 3 10 02 PM

With margin set to  margin: 0px calc(-30px * .5 + 35px); and .content padding set to 0
Screenshot 2024-03-11 at 3 10 20 PM

jcoyne:
@dnoneill that feels like we're not taking advantage of all the available real-estate, making the design look more cramped than necessary.

dnoneill:
@dbranchini Do you have an opinion on this?

dbranchini:
Out of all of the screenshots, I like the third example best, but I also see Justin's point and wonder if the left and right margins could be smaller. I suppose they can't because you're providing room for the side toolbar. In an ideal world, I'd prefer that the side toolbar overlapped content and/or was collapsible into a single button (like Material's [floating action button](https://m2.material.io/components/buttons-floating-action-button#usage)). When I look at it on my phone in production, it does seem to overlap some content, but there also seems to be a horizontal scroll too? The up/down button on the search results page also seems to overlap content.

PR #4066, opened 2024-03-01
Pick the representative item for browse_nearby_struct for each base c…

cbeer:
…all number

The lopped call number and shelfkey are, currently, effectively the same thing. It's probably better to use the original value instead of the transformed one, though.

### DIFF
--- a/app/models/concerns/solr_holdings.rb
+++ b/app/models/concerns/solr_holdings.rb
@@ -29,7 +29,15 @@ module SolrHoldings
                              browseable_schemes = %w[LC DEWEY ALPHANUM]
                              browseable_items = items.select { |v| v.shelfkey.present? && browseable_schemes.include?(v.callnumber_type) }
 
-                             browseable_items.uniq(&:shelfkey).map do |v|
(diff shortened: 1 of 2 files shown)


app/models/concerns/solr_holdings.rb
```diff
@@ -29,7 +29,15 @@ def browseable_spines
                              browseable_schemes = %w[LC DEWEY ALPHANUM]
                              browseable_items = items.select { |v| v.shelfkey.present? && browseable_schemes.include?(v.callnumber_type) }
 
-                             browseable_items.uniq(&:shelfkey).map do |v|
+                             representative_items = browseable_items.group_by(&:truncated_callnumber).map do |_base_call_number, items|
+                               if items.one?
+                                 items.first
+                               else
+                                 items.min_by(&:full_shelfkey)
```
jcoyne, line 36: Do we need the conditional?  Can't we use `min_by` for all cases?

jcoyne (changes requested):

app/models/concerns/solr_holdings.rb
cbeer: Functionally, yes, although I thought being explicit was clearer 🤷‍♂️

PR #4070, opened 2024-03-04
Bound with redesign

jcoyne:
Ref #4074 

Screenshot 2024-03-08 at 12 30 11 PM

### DIFF
--- a/app/components/search_result/location_item_component.rb
+++ b/app/components/search_result/location_item_component.rb
@@ -1,39 +1,17 @@
 # frozen_string_literal: true
 
-module AccessPanels
+module SearchResult
   class LocationItemComponent < ViewComponent::Base
     with_collection_parameter :item
-    attr_reader :item, :document, :classes, :counter
+    attr_reader :item, :document, :classes
 
-    def initialize(item:, document:, item_counter:, classes: nil, render_item_level_request_link: true)
+    def initialize(item:, document:, classes: nil, render_item_level_request_link: true)
       super
 
       @item = item
       @classes = classes
       @document = document
       @render_item_level_request_link = render_item_level_request_link
-      @counter = item_counter
-    end
-
-    delegate :bound_with_parent, to: :item
-
-    def html_data
-      data = { barcode: item.barcode }
-      return data if counter < 5
-
-      data.merge(long_list_target: 'hideable')
-    end
-
-    def bound_with_parent?
-      bound_with_parent.present?
-    end
-
-    def bound_with_title
-      bound_with_parent['title']
-    end
-
-    def bound_with_callnumber
-      [bound_with_parent['call_number'], bound_with_parent['volume'], bound_with_parent['enumeration'], bound_with_parent['chronology']].compact.join(' ')
     end
 
     def availability_text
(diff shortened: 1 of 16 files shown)


app/components/access_panels/location_item_component.rb
```diff
@@ -3,16 +3,46 @@
...
+    end
+
+    def bound_with_parent?
+      item.bound_with_parent.present?
+    end
+
+    def bound_with_title
+      bound_with = item.bound_with_parent
+
+      # TODO: remove this after the index is rebuild with https://github.com/sul-dlss/searchworks_traject_indexer/pull/1379
```
dnoneill, line 33: Looks like this was merged 2 weeks ago so the index should have been reindexed by now. Please remove these lines before merge.

app/components/access_panels/location_item_component.rb
```diff
@@ -3,16 +3,46 @@
...
+
+      bound_with['title']
+    end
+
+    def bound_with_callnumber
+      bound_with = item.bound_with_parent
+
+      # TODO: remove this after the index is rebuild with https://github.com/sul-dlss/searchworks_traject_indexer/pull/1379
+      return if bound_with.respond_to?(:instance)
+
```
dnoneill, line 44: See above comment

app/components/search_result/location_item_component.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
```
dnoneill, line 2: This seems to be repeating a lot of what is in module AccessPanels.LocationItemComponent. Is there a reason we have to repeat all the same code. Could we use inheritance or something along those lines to not repeat code?

lib/holdings/item.rb
```diff
@@ -62,6 +62,11 @@ def type
       item_display[:type]
     end
 
+    # TODO: I believe we can look in `bound_with.hrid` for this after a reindex
+    def instance_hrid
+      item_display[:instance_hrid]
```
dnoneill, line 67: Please double check before merge.

lib/holdings/item.rb
```diff
@@ -107,6 +112,10 @@ def on_reserve?
     end
 
     def bound_with_parent
+      # TODO: This depends on https://github.com/sul-dlss/searchworks_traject_indexer/pull/1379
+      return item_display[:bound_with] if item_display[:bound_with]
+
+      # TODO: we're only keeping the rest of this until the index is rebuild with the PR linked above PR
```
dnoneill, line 118: Remove. Reindex should have happened.

dnoneill (changes requested):

app/components/search_result/location_item_component.rb
jcoyne: Yes, but I believe the way to do that is by extracting a new class rather than forcing inheritance.  I'm trying not to make this PR too challenging to review, thus I've avoided this at this time.

jcoyne: See https://github.com/sul-dlss/SearchWorks/pull/4151 as a way to preemptively reduce the duplication.

PR #4080, opened 2024-03-05
Use holdings_library_code_ssim to look up the eresources owning library

cbeer:
Goes with the indexer changes in https://github.com/sul-dlss/searchworks_traject_indexer/pull/1390 so this is decoupled from `item_display_struct` data entirely.

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -132,15 +132,17 @@ class SolrDocument
   # Checks holdings to see if there are any physical holdings
   # If there aren't and the library isn't sul, return the library name
   def eresources_library_display_name
-    library = holdings.libraries&.first
-    return unless library
+    return if has_physical_copies?
 
-    has_physical_copies = holdings.libraries.any?(&:present?)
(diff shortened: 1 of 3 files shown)


app/models/solr_document.rb
```diff
@@ -132,15 +132,17 @@ def file_ids
...
+    return if has_physical_copies?
 
-    has_physical_copies = holdings.libraries.any?(&:present?)
-    return if has_physical_copies
+    library_code = first(:holdings_library_code_ssim)
 
-    unless library.code == 'SUL'
-      library.name
-    end
+    return if library_code == 'SUL' || holdings&.libraries&.first&.code == 'SUL'
```
jcoyne, line 139: Presumably we're going to remember to remove this second clause after migrating the index?

jcoyne (approved):

app/models/solr_document.rb
cbeer: 🤞

PR #4083, opened 2024-03-05
Remove test about SolrDocument from a component test

jcoyne:
This isn't an appropriate place to test this method. This was added in 6837a2206ce0d9492334659c68668a9034bae39b


marlo-longley:
Rather than just deleting it, should we relocate it?

jcoyne:
@marlo-longley there are already tests for it in SolrDocument.

PR #4085, opened 2024-03-05
Add person entitlement attribute to check for authorization for articles

### LINKED ISSUE #4082: Route eduPersonEntitlement into authorization logic
Relates to authorization issue and work completed here: #3690 .

We have been using "eduPersonAffiliation" which is mapped to "unscoped-affiliation".  The code sends a Honeybadger notification if a person's access is enabled by the "suAffiliation" attribute value while the "unscoped-affiliation" value does not provide access.  Based on Honeybadger notifications we received after we implemented this change, there are still several cases where this particular discrepancy occurs, and many times, the "unscoped-affiliation" attribute is empty. 

We can eduPersonEntitlement as a value to check against, and maintain the Honeybadger notification to see if the discrepancy continues to occur.

hudajkhan:
Closes #4082 

We have been using "eduPersonAffiliation" which is mapped to "unscoped-affiliation". The code sends a Honeybadger notification if a person's access is enabled by the "suAffiliation" attribute value while the "unscoped-affiliation" value does not provide access. Based on Honeybadger notifications we received after we implemented this change, there are still several cases where this particular discrepancy occurs, and many times, the "unscoped-affiliation" attribute is empty.

This pull request adds "eduPersonEntitlement" as a value to check against, and maintain the Honeybadger notification to see if the discrepancy continues to occur.  The pull request also adds a few tests (application controller) to ensure that the code follows the logic we want i.e. if any of "eduPersonEntitlement", "unscoped-affiliation", or the check against the list of affiliations represented by "suAffiliation" return values that provide access, we will say the user has the necessary affiliations for access.  

Also, "eduPersonEntitlement" now has values coming in for production, staging, and uat.


app/views/articles/_search_results.html.erb
```diff
@@ -2,6 +2,11 @@
 <div class="breadcrumb row">
   <div class="col-md-12">
     <%= render 'constraints' %>
+    <% if ! current_user.blank? %>
+      <%= "SU affiliations: #{current_user.affiliations} "%>
```
jcoyne, line 6: do we want this to be visible to users?

hudajkhan: This is for testing. I will remove this before I undraft the PR.

corylown (approved):

PR #4101, opened 2024-03-07
Prevent the gallery from using duplicate identifiers for bookmarks

### LINKED ISSUE #4088: Duplicate identifiers on form elements
> IDs used in ARIA and labels must be unique: Document has multiple elements referenced with ARIA with the same id attribute: toggle-bookmark_6774934

```
document.querySelectorAll('#toggle-bookmark_6774934')
NodeList [ input#toggle-bookmark_6774934.toggle-bookmark, input#toggle-bookmark_6774934.toggle-bookmark ]
```

on https://searchworks.stanford.edu/view/6774934

jcoyne:
Fixes #4088

### DIFF
--- a/app/assets/stylesheets/modules/bookmarks.scss
+++ b/app/assets/stylesheets/modules/bookmarks.scss
@@ -17,6 +17,25 @@ div.toggle-bookmark {
   }
 }
 
+// Note this is same as above, but with a slightly modified className
+// This is for the bookmark checkboxes in the browse section.
+div.browse-toggle-bookmark {
+  label.browse-toggle-bookmark {
+    cursor: pointer;
+
+    input.browse-toggle-bookmark {
+      margin-right: 3px;
+    }
+
+    span {
+      font-size: 0.9em;
+      display: inline-block;
+      margin-bottom: 4px;
+      vertical-align: middle;
+    }
+  }
+}
+
 .index-document-functions {
   text-align: right;
 }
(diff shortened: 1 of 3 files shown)


hudajkhan:
As described in the original issue, in production the select box at the very top has the same id as one of the browse gallery select boxes.  Screenshots below:

Screenshot 2024-03-08 at 11 13 10 AM

Screenshot 2024-03-08 at 11 24 16 AM

hudajkhan:
There is some behavior around the interaction between the select box on top of the page and the gallery browse that is also in production, so I do not think is related to these changes. If one selects the checkbox on the top, and then also selects the same item from the list in the gallery, plus another item, and then deselects the first item from the gallery, the "selections" list on top now only shows one item.  Probably a different issue.

app/assets/stylesheets/modules/bookmarks.scss
```diff
@@ -17,6 +17,25 @@ div.toggle-bookmark {
...
+
+    span {
+      font-size: 0.9em;
+      display: inline-block;
+      margin-bottom: 4px;
+      vertical-align: middle;
+    }
+  }
+}
+
```
hudajkhan, line 38: Perhaps there's a way to define some of this once and use it twice? I won't hold up the PR for this though

hudajkhan (approved):
I have tried it out locally, and also don't see any conflicts with the changes within the code itself.

PR #4102, opened 2024-03-07
Rename method to showGallery as showPreview is confusing

jcoyne:
The preview is what shows when you click the 'preview' button.


hudajkhan (approved):
Gallery view still works after these changes.  Doing a search for "showPreview" doesn't show any stimulus style calls for this class.

PR #4105, opened 2024-03-07
Update honeybadger

jcoyne:


dnoneill:
@jcoyne echoing @marlo-longley's sentiments on slack is there a way to test to make sure that the new version of honeybadger works with the js in the code? https://github.com/sul-dlss/SearchWorks/blob/3b041f7f5712b9f1d6d468964124af821d96c5ee/app/views/layouts/searchworks.html.erb#L35-L47

jcoyne:
I don't think there is a good way to check this. I think this is very much just like a dependency update like we do every monday. It's just a minor release bump.

dnoneill (approved):

PR #4106, opened 2024-03-07
Upgrade to the css2 google font api

jcoyne:
This doesn't require the font loader script.


marlo-longley:
The fonts do look different to me with this checked out locally vs production/main branch
Digging into why...

new API / this branch
Screenshot 2024-03-08 at 11 40 03 AM

production / main
Screenshot 2024-03-08 at 11 39 55 AM

marlo-longley (changes requested):
Looking into why the font isn't loading now

jcoyne:
@marlo-longley pull the latest and see if that fixed it.

PR #4136, opened 2024-03-14
update link in alert info to be accessible

### LINKED ISSUE #4125: fix feedback form contrast error
Screenshot 2024-03-12 at 1 50 30 PM

dnoneill:
closes #4125

### DIFF
--- a/app/assets/stylesheets/bootstrap-variables.scss
+++ b/app/assets/stylesheets/bootstrap-variables.scss
@@ -43,6 +43,9 @@ $state-danger-text: #842029;
 // Colors
 $gray-light: #737373;
 
+// Alerts
+$alert-link-font-weight: 500;
+
 // Navbar
 $navbar-dark-color: $white;
 $navbar-dark-toggler-border-color: transparent;
--- a/app/views/shared/feedback_forms/_reporting_from.html.erb
+++ b/app/views/shared/feedback_forms/_reporting_from.html.erb
@@ -4,7 +4,7 @@
       Reporting from: <span class="reporting-from-field"><%= request.referer %></span>
     </div>
     <div class="col-sm-2">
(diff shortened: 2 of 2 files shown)


jcoyne:
@dnoneill can you explain what "in line with bootstrap" means?

dnoneill:
We override the color for alert-info with this line https://github.com/sul-dlss/SearchWorks/blob/7c5bc1f741cfa45ab231e5dfa4b82b5db26fd480/app/assets/stylesheets/base.scss#L20 so this is just overriding our override so that the color is being set from bootstrap.

jcoyne:
@dnoneill that's a `text-decoration-color`, not a `color`

jcoyne:
Have you considered using the `.alert-link`style provided by BS? https://getbootstrap.com/docs/4.6/components/alerts/#link-color

dnoneill:
Sorry I was thinking of something else. 
I think that might be a question for @dbranchini. We definitely could use digital_blue_dark. It does not cause a contrast error.

I did look at .alert-link. I did not think it looked very good. (see attached screenshot)
Screenshot 2024-03-14 at 1 41 02 PM

jcoyne:
@dnoneill if we use alert-link, we may need to override some styles for it, but it's a good descriptive class name anyway and it's purpose is documented.

dnoneill:
That seems like a lot of css changes so we can include a class that we are going to strip most of the css out of.

jcoyne:
@dnoneill the only css change would be to change the font weight which is a Bootstrap config variable.

dnoneill:
and I guess column wouldn't have to adjust column size because once the bold is removed it is fine. Okay. Let's go with that.

jcoyne (approved):

PR #4199, opened 2024-04-26
Using config/secrets.yml is deprecated. We can use the environment in…

jcoyne:
…stead


jcoyne:
See https://github.com/sul-dlss/SearchWorks/issues/4378#issuecomment-2291561872

PR #4203, opened 2024-05-01
Setup EDS session for all ArticlesController actions

corylown:


app/controllers/articles_controller.rb
```diff
@@ -17,7 +17,7 @@ class ArticlesController < ApplicationController
 
   before_action :set_search_query_modifier, only: :index
 
-  before_action :eds_init, only: %i[index show]
+  before_action :eds_init, only: %i[index show fulltext_link]
```
cbeer, line 20:
```suggestion
  before_action :eds_init
```

Are there actions that shouldn't go through this? It looks pretty safe?

corylown: Yeah, it's probably fine to run it for anything that goes through ArticlesController 🤷

PR #4210, opened 2024-05-06
Make sure item tables have a unique-to-the-page id

cbeer:
Without unique ids, the show more/less controls on the search results pages don't work. This seems like as a regression introduced in https://github.com/sul-dlss/SearchWorks/pull/4059/files (where previously we didn't give these elements id attributes at all?)

### DIFF
--- a/app/components/access_panels/library_location_component.html.erb
+++ b/app/components/access_panels/library_location_component.html.erb
@@ -26,7 +26,7 @@
       <th scope="col">Status</th>
     </tr>
   </thead>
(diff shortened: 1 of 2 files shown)


corylown (changes requested):
@cbeer isn't this the component that gets rendered on the full record page? I think this one also needs updating to fix the issue on the results page: https://github.com/sul-dlss/SearchWorks/blob/ce8823e23f4008a292c9636d2140b9f89c5f7fca/app/components/search_result/location_component.html.erb#L24

corylown (approved):

PR #4218, opened 2024-05-13
Use holdings_library_code_ssim to check library for ezproxy links

### LINKED ISSUE #4217: Proxy link not getting applied to LAW (and LANE) links
From SearchWorks Feedback:

Hello,

I noticed that Law’s proxy logic is not being correctly applied. I have included a few examples below. Please can you investigate?
 
Thanks,
Qi
 
Almanac of the federal judiciary
https://searchworks.stanford.edu/view/12864848
without law proxy:
https://almanacofthefederaljudiciary.com/
 
Immigration law & policy in the U.S.
https://searchworks.stanford.edu/view/11925296
with SUL proxy instead of Law’s:
https://stanford.idm.oclc.org/login?qurl=https://heinonline.org/HOL/Index?index=immigration&collection=immigration
 
LLMC digital
https://searchworks.stanford.edu/view/8336445
with SUL proxy instead of Law’s:
https://stanford.idm.oclc.org/login?qurl=https://www.llmc.com

corylown:
Fixes #4217

### DIFF
--- a/app/models/links/ezproxy.rb
+++ b/app/models/links/ezproxy.rb
@@ -10,7 +10,7 @@ class Links
     attr_reader :url, :link_title, :document
 
     def initialize(url:, link_title:, document:)
-      @url = url
+      @url = url&.strip
       @link_title = link_title
       @document = document
     end
--- a/app/models/links/ezproxy.rb
+++ b/app/models/links/ezproxy.rb
@@ -53,7 +53,7 @@ class Links
     end
 
     def libraries
-      @libraries ||= document.holdings.libraries.map(&:code)
+      @libraries ||= document.fetch(:holdings_library_code_ssim, [])
     end
 
     def ezproxied_hosts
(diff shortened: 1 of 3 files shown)


app/models/links/ezproxy.rb
```diff
@@ -53,7 +53,7 @@ def apply_sul_proxy_prefix?
     end
 
     def libraries
-      @libraries ||= document.holdings.libraries.map(&:code)
+      @libraries ||= document.folio_holdings.map { |h| h.effective_location.library.code }
```
cbeer, line 56: If these links are rendered on the search results page, I think we want to avoid digging into `folio_holdings` for performance reasons, right?

I think we use `holdings_library_code_ssim` and/or `library_code_facet_ssim` for similar things?

corylown: Makes sense. Fixed.

PR #4231, opened 2024-06-03
Update dependencies

sul-devops-team:


marlo-longley:
Waiting on a fix from newrelic for this -- should be coming soon.

PR #4237, opened 2024-06-06
Use the BM25 similarity class

cbeer:
BM25 is the "new" Solr default and causes some minor relevancy changes. The major difference seems to be a little boost to exact matching:

E.g. a query for "The John" no longer ranks the stemmed match "The Johns" higher than exact matches for "The John" 
Screenshot 2024-06-06 at 16 22 34

And putting a little more weight on more unique terms:

E.g. a query for "mcrae jazz" returns titles with Mcrae above ones with Jazz: 
Screenshot 2024-06-06 at 16 26 03


cbeer:
gryphon-search signed off on shipping this on 6/20.

PR #4249, opened 2024-06-17
Use database to store courses information, and update database hourly

### LINKED ISSUE #4175: Fetch and display course information more frequently
Reserves staff were accustomed to immediate updates in SearchWorks display when items were placed on Reserve in Symphony. 

With FOLIO, it's my understanding from Cory that the indexer knows pretty promptly that a title/item has been associated with a course but we don't update the display (https://searchworks.stanford.edu/reserves or in the record view) until SearchWorks is deployed. In the first two weeks of the quarter, when staff are creating many FOLIO Courses and communicating with Instructors about Course status, this is too slow.

In an ideal scenario, we'd be updating the SearchWorks display as soon as the title/item is associated with a current course in FOLIO. Cory has indicated that immediate updates aren't possible right now given the amount of info that needs to be updated.  

Since immediate isn't possible, updating the SearchWorks display every hour would be the next best thing.  If hourly isn't possible, then anything up to daily is workable. But at the very slowest, we should update the display once per day.

hudajkhan:
Closes #4175 .

What this pull request does:

- Removes courses.json from version control.
- Updates FOLIO types to remove courses from the list of classes to be synced to files.
- Updates the database to include a course reserves table that will capture name, course number, id, and instructors list (as an array).
- Since Course Reserve now uses "instructors" instead of "instructor", there are updates to the spelling in various places.
- A rake task now enables retrieving course information from FOLIO and updating the database. 
- Schedule.rb has an added entry to kick off the rake task every hour. 
- One server from stage and production now are designated with the role "courses".  This role is used to kick off the course update task.
- For testing, I've also removed the course reserves facet values listed in the fixture Solr documents. The reason for this is that many of these tests were failing because the course reserve info is no long read in directly from files but these tests were not testing course reserve functionality directly. I've instead updated the tests that do review course reserve functionality directly to use FactoryBot CourseReserve objects. 
- There is one feature test I merged in directly with a component test (since it was testing the rendering of content within that component anyway).  A final feature test I added "skip" too because I'm not sure how to handle this scenario.  If a better test approach is possible, we can work on that.


Gemfile
```diff
@@ -55,6 +55,8 @@ group :development, :test do
   gem 'selenium-webdriver'
   gem 'solr_wrapper'
   gem 'webmock'
+  # factory_bot_rails for creating fixtures in tests
+  gem 'factory_bot_rails', '~> 6.2.0' # Pinned until https://github.com/thoughtbot/factory_bot_rails/issues/433 is resolved
```
jcoyne, line 59: That issue was resolved already.  Is there a reason we can't use 6.4?

hudajkhan: I was copy-pasting from the requests gemfile, so I will update this one to remove the version restriction.

app/models/course_reserve.rb
```diff
@@ -1,35 +1,8 @@
...
-
-  def self.all
-    Repository.instance.all
-  end
-
-  def self.find(*id)
-    Repository.instance.find(*id)
-  end
+class CourseReserve < ApplicationRecord
+  serialize :instructors, coder: YAML, type: Array
```
jcoyne, line 5: Why YAML and not JSON?

hudajkhan: No reason except I needed a coder and that was one I saw existed. I can change to JSON.

app/services/folio/courses.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    attr_reader :folio_client
+
+    def initialize(folio_client: FolioClient.new)
+      @folio_client = folio_client
+    end
+
+    # Write courses from FOLIO to database
+    def sync_courses!
```
cbeer, line 20: Is there a mechanism for removing old course reserves data?

corylown, line 20: Do we also need to delete courses from the database that have been removed from FOLIO?

hudajkhan: You and @corylown below have the same question so will just answer here.  I had not put that in and should. I could empty out the whole database and rewrite to it (once I know for sure I do have any data coming back) instead of handling atomic deletes.  Or, if we would rather do atomic deletes, find which ones did not come through in the latest list and remove them.

hudajkhan: @corylown Chris had the same question above so I answered it there. In short, yes, and I should add that.

corylown: I'd lean toward atomic deletes to avoid moments throughout the day when the courses table would be empty.

hudajkhan: Ok. Will look into that. Thank you!

hudajkhan: Added deletion step.

corylown (approved):
I'll approve but not merge to give other reviewers a chance to have a look again.

hudajkhan:
Thanks @corylown !

app/models/course_reserve.rb
hudajkhan: I ran into some issues with JSON - requires a little more review so leaving it for now for this PR but will revisit separately.

hudajkhan:
Closing this pull request since we duplicated the work in a new branch and to prevent rebase confusion: https://github.com/sul-dlss/SearchWorks/pull/4284

PR #4254, opened 2024-06-18
Allow location-level request buttons for records that have a mix of b…

### LINKED ISSUE #4247: Request button not displaying when first item is boundwith
Example record: https://searchworks.stanford.edu/view/3454840

The first item in the SAL3 item list is a boundwith. The rest of the display looks ok (status of the boundwith item is inherited from the parent; truck icon & status for other volumes), but there is no request button at the location level. There is no way for a user to request the other 4 volumes. 

image

I'll add other examples as I find them.

cbeer:
…ound-with and regular holdings. Fixes #4247

### DIFF
--- a/app/components/access_panels/library_location_component.html.erb
+++ b/app/components/access_panels/library_location_component.html.erb
@@ -1,8 +1,7 @@
 <%= render partial: "catalog/stackmap_link", locals: { document: document, location: location, location_name_class: "location-name" } %>
+<span class="pull-right"><%= render location_request_link %></span>
 <% if location.bound_with? && document.bound_with_note? %>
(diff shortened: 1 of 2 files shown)


corylown (changes requested):
@cbeer there's some additional logic around boundwiths in https://github.com/sul-dlss/SearchWorks/blob/80d59b7bf6aeaf0d62214a45bc0b77d1bb3c2439/app/components/access_panels/library_location_component.html.erb#L2 preventing the location request button from displaying on the record view. It does show on the results view.

corylown:
And also...the location level request link already appears on the results page in production: https://searchworks.stanford.edu/?search_field=search&q=+Estudo+sobre+a+absentismo+e+a+instabilidade+da+m%C3%A3o-de-obra+africana+

corylown (approved):

PR #4260, opened 2024-06-19
fix due date when using back button

### LINKED ISSUE #4204: Due dates duplicating for boundwith records after using the browser "back" button
We got a report via SearchWorks feedback (from a library staff member) that praised when we show the due date from the parent record on the child record in "At the library". But the feedback was qualified that the due date "displays twice".  It appears that the due date is duplicated each time the browser back button is used. This seems to happen on both the parent on the child record. But it's particular obvious when you click on a child record, then the link to the parent record.  The only way to get back to the child record is to use the browser back button.

Here's a screen shot after using the back button twice on a child record:
image

Boundwith record used in the report: 
https://searchworks.stanford.edu/view/8003016

dnoneill:
closes #4204

### DIFF
--- a/app/javascript/controllers/live_lookup_controller.js
+++ b/app/javascript/controllers/live_lookup_controller.js
@@ -25,7 +25,8 @@ export default class extends Controller {
           status_text.html(live_data.status)
         }
 
-        if (live_data.due_date) {
+        // check for due date and if date in temporary location (if the user goes to the page by hitting back).
+        if (live_data.due_date && !temporary_location.text().includes('Due')) {
           temporary_location.append(`Due ${live_data.due_date}`);
         }
       }


app/javascript/controllers/live_lookup_controller.js
```diff
@@ -25,7 +25,7 @@ export default class extends Controller {
           status_text.html(live_data.status)
         }
 
-        if (live_data.due_date) {
+        if (live_data.due_date && !temporary_location.text().includes('Due')) {
```
jcoyne, line 28: Would you mind adding a code comment here explaining what this is doing?

jcoyne (changes requested):

app/javascript/controllers/live_lookup_controller.js
```diff
@@ -25,7 +25,8 @@ export default class extends Controller {
           status_text.html(live_data.status)
         }
 
-        if (live_data.due_date) {
+        // Check if there is a due date and if the due date is already in temporary location.
```
jcoyne, line 28: Why would there already be a due date?  Can you give a bit more context?

jcoyne (changes requested):

jcoyne (approved):

PR #4284, opened 2024-06-28
Move courses db

hudajkhan:
Closes https://github.com/sul-dlss/SearchWorks/issues/4175 .

What this pull request does:

- Removes courses.json from version control.
- Updates FOLIO types to remove courses from the list of classes to be synced to files.
- Updates the database to include a course reserves table that will capture name, course number, id, and instructors list (as an array).
- Since Course Reserve now uses "instructors" instead of "instructor", there are updates to the spelling in various places.
- A rake task now enables retrieving course information from FOLIO and updating the database.  Courses that are in the database but not in the latest list retrieved from FOLIO are deleted.  
- Schedule.rb has an added entry to kick off the rake task every hour.
- One server from stage and production now are designated with the role "courses". This role is used to kick off the course update task.
- For testing, I've also removed the course reserves facet values listed in the fixture Solr documents. The reason for this is that many of these tests were failing because the course reserve info is no long read in directly from files but these tests were not testing course reserve functionality directly. I've instead updated the tests that do review course reserve functionality directly to use FactoryBot CourseReserve objects.
- There is one feature test I merged in directly with a component test (since it was testing the rendering of content within that component anyway). A final feature test I added "skip" too because I'm not sure how to handle this scenario. If a better test approach is possible, we can work on that.


app/models/course_reserve.rb
```diff
@@ -1,35 +1,8 @@
 # frozen_string_literal: true
 
 # The store of course reserve data from courses.json
```
corylown, line 3: I think this comment is no longer accurate.

corylown (changes requested):

app/models/course_reserve.rb
hudajkhan: I changed it

corylown (approved):

PR #4311, opened 2024-07-09
Remove courses role; use db role for updating course data and pruning guest user data

corylown:
This PR is an alternative to https://github.com/sul-dlss/SearchWorks/pull/4310 with more cleanup of roles.


config/deploy.rb
```diff
@@ -39,6 +39,8 @@
 # We want Honeybadger to report deployments to our Capistrano stage names (e.g., dev, stage, prod)
 set :honeybadger_env, fetch(:stage)
 
+set :whenever_roles, %i[app db]
```
jcoyne, line 42: do we still need `app` here?  Aren't they all the tasks running on `db`?

jcoyne (approved):

PR #4318, opened 2024-07-10
Hook up rack-attack to redis, if configured.

cbeer:


jcoyne:
Do we want to change this cronjob so it doesn't have to run on all of the app nodes? https://github.com/sul-dlss/SearchWorks/blob/main/config/schedule.rb#L17-L19

jcoyne (approved):

PR #4336, opened 2024-07-17
Remove barcode search endpoint

cbeer:
This endpoint was replaced by a [Symphony lookup](https://github.com/sul-dlss/dor-services-app/pull/875) in 2020.


jcoyne (approved):
👏

PR #4357, opened 2024-08-01
add an access_panels component for linking to the same object in other user facing systems

### LINKED ISSUE #4312: Add link to view items listed in Earthworks
https://github.com/sul-dlss/earthworks/issues/310 will add the reverse of this (links from Earthworks to Searchworks), so it seemed prudent to have the link work both ways!

Release metadata seems to already be available in the page context (because it's used to render the schema.org metadata for records, which notes that they are available in Earthworks where appropriate):
```json
  "includedInDataCatalog": {
    "@type": "DataCatalog",
    "name": "https://earthworks.stanford.edu"
  }
```

Purl `bd285ct9109` is currently listed in both platforms.

We could add a simple link to the metadata (like we do for the purl), but it might be more visible/useful to render some sort of context box, like we do for collection links:

Screenshot 2024-07-09 at 13 57 07

As such, this might require some design.

jmartin-sul:
remove existing PURL link from MODS bibliographic info, since it's moving to this new panel

Closes #4312

**_PURL in new section, no longer appears in MODS bibliographic section, Earthworks link present also since it's released to EW:_**
Screenshot 2024-08-02 at 12 26 08 PM

**_no link to either EW or PURL, but still shows a non-PURL location in the bibliographic section:_**
Screenshot 2024-08-02 at 12 27 24 PM

**_link to PURL now appears in "Also listed in", didn't appear in bibliographic section before anyway, since that section is rendered from MARC for this object, and the MARC biblio partial doesn't show the 856 field:_**
Screenshot 2024-08-02 at 12 28 39 PM

### DIFF
--- a/app/components/access_panels/other_listings_component.rb
+++ b/app/components/access_panels/other_listings_component.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module AccessPanels
+  class OtherListingsComponent < AccessPanels::Base
+    def render?
+      earthworks_url.present? || purls.present?
+    end
+
+    def earthworks_url
+      @earthworks_url ||= construct_earthworks_url
+    end
+
+    # @return [Array<String>] a tags representing PURL links
+    def purl_links
+      @purl_links ||=
+        if purls.size > 1 # text of the link is the actual URL if there is more than one unique URL, so there aren't multiple links with the same title
+          purls.map { |purl| link_to(purl, purl) }
+        elsif purls.size == 1
+          Array(link_to('Stanford Digital Repository', purls.first))
+        else
+          []
+        end
+    end
+
+    private
+
+    # @return [Array<String>] URLs representing links to PURL, whether specified in MARC or MODS, and de-duped, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def purls
+      @purls ||= (mods_purls + marc_managed_purls).uniq
+    end
+
+    # @return [Array<String>] URLs representing links to other locations, extracted from MODS metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def mods_purls
+      Array(document['url_fulltext']).find_all { |str| str.start_with?(Settings.PURL_EMBED_RESOURCE) }
+    end
+
+    # @return [Array<String>] URLs representing links to other locations, extracted from MARC metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def marc_managed_purls
+      document.marc_links.managed_purls.filter_map(&:href) # filter_map because href can be nil, at least in fixture data
+    end
+
+    def construct_earthworks_url
+      return nil unless document['dor_content_type_ssi'] == 'geo'
+
+      # * Hardcoding this simple URL format because it should work for all things it'd apply to (anything released to both SW and EW should be a Stanford object)
+      # * Using Kernel.format because there's a clashing ViewComponent::Base#format, so the namespacing is needed
+      Kernel.format('https://%<hostname>s/catalog/stanford-%<bare_druid>s', hostname: Settings.earthworks.hostname, bare_druid: document.druid)
+    end
+  end
+end
--- a/app/views/catalog/record/_mods_bibliographic.html.erb
+++ b/app/views/catalog/record/_mods_bibliographic.html.erb
@@ -24,6 +24,6 @@
   <%= render ModsDisplay::FieldComponent.new(field: identifier, label_html_attributes: { class: 'col-md-3' }, value_html_attributes: { class: 'col-md-9' }) %>
 <% end %>
 
-<% document.mods.location.each do |location| %>
+<% document.mods.location.reject { |location| location.values.detect { |loc| /href=.#{Settings.PURL_EMBED_RESOURCE}/.match?(loc) } }.each do |location| %>
   <%= render ModsDisplay::FieldComponent.new(field: location, label_html_attributes: { class: 'col-md-3' }, value_html_attributes: { class: 'col-md-9' }) %>
 <% end %>
--- a/app/components/access_panels/other_listings_component.html.erb
+++ b/app/components/access_panels/other_listings_component.html.erb
@@ -1,0 +1,11 @@
+<%= render @layout.new do |component| %>
+  <% component.with_title { 'Also listed in' } %>
+  <% component.with_body do %>
+    <% if earthworks_url.present? %>
+      <p><%= link_to('Earthworks', earthworks_url) %></p>
+    <% end %>
+    <% purl_links.each do |purl_link| %>
+      <p><%= purl_link %></p>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/catalog/record/_metadata_panels.html.erb
+++ b/app/views/catalog/record/_metadata_panels.html.erb
@@ -11,6 +11,7 @@
 
   <% potential_context = capture do %>
     <%= render AccessPanels::ExhibitComponent.new(document: document) %>
+    <%= render AccessPanels::OtherListingsComponent.new(document: document) %>
   <% end %>
 
   <% if context.present? || potential_context.present? %>
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -275,3 +275,6 @@ folio_unrequestable_statuses:
   - Unavailable
   - Unknown
   - Withdrawn
+
+earthworks:
+  hostname: 'earthworks.stanford.edu'
--- a/spec/components/access_panels/other_listings_component_spec.rb
+++ b/spec/components/access_panels/other_listings_component_spec.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe AccessPanels::OtherListingsComponent, type: :component do
+  include MarcMetadataFixtures
+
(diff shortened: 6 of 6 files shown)


app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,55 @@
...
+      Array(document.fetch('marc_links_struct', {}).pluck('href'))
+    end
+
+    def construct_earthworks_url
+      return nil unless document.schema_dot_org?
+
+      # This value should always be a single hash.  It's in the required JSON-LD format for a value of type DataCatalog as
+      # specified by schema.org, even if there is a single value. It should be safe to assume the value won't be a list (despite
+      # the seemingly generalizable '@type' and 'name' fields in the hash).
+      data_catalog_name = document.as_schema_dot_org.dig('includedInDataCatalog', 'name')
```
cbeer, line 47: Pulling this out of `as_schema_dot_org` seems a little indirect (and surpising) to me. Can we just use the type `geo` like the indexer does?

https://github.com/sul-dlss/searchworks_traject_indexer/blob/main/lib/traject/config/sdr_config.rb#L351

Can we check that here, or

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,55 @@
...
+    # @return [Array<String>] URLs representing links to other locations, extracted from MODS metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def mods_other_locations
+      #   latter of those is used in app/views/catalog/record/_mods_bibliographic.html.erb (but has some clutter unlike the mockup, for this purpose)
+      Array(document['url_fulltext'])
+    end
+
+    # @return [Array<String>] URLs representing links to other locations, extracted from MARC metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def marc_other_locations
+      # the fetch with the default empty hash prevents errors for data without MARC representation
+      Array(document.fetch('marc_links_struct', {}).pluck('href'))
```
cbeer, line 38:
```suggestion
      document.marc_links.managed_purls.map(&:href)
```

probably gets you the right PURL links. There's probably records that link to supplemental resources or other things that you probably don't want to link here?

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,55 @@
...
+    def render?
+      earthworks_url.present? || purls.present?
+    end
+
+    def earthworks_url
+      @earthworks_url ||= construct_earthworks_url
+    end
+
+    # @return [Array<String>] URLs representing links to PURL, whether specified in MARC or MODS, and de-duped, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def purls
```
cbeer, line 14: Can this be a private method too?

app/views/catalog/record/_mods_bibliographic.html.erb
```diff
@@ -25,5 +25,7 @@
 <% end %>
 
 <% document.mods.location.each do |location| %>
-  <%= render ModsDisplay::FieldComponent.new(field: location, label_html_attributes: { class: 'col-md-3' }, value_html_attributes: { class: 'col-md-9' }) %>
+  <% unless location.values.detect { |loc| /href=.#{Settings.PURL_EMBED_RESOURCE}/.match?(loc) } %>
```
cbeer, line 28: Can this condition get pulled up to `document.mods.location.reject { ... }.each |location|`?

app/components/access_panels/other_listings_component.rb
jmartin-sul: thanks!

i only went this direction because it was the suggestion on the ticket, and because i'm pretty unfamiliar with searchworks and its indexing.  but your suggestion worked in my local testing, definitely feels less circuitous, and i think alleviates a spec difficulty i was running into.  (i was having trouble indexing a harvested fixture with a `schema_dot_org_struct` using `bin/rails searchworks:fixtures`, and i had to submit JSON directly through the solr web UI to get a document for local testing).

app/components/access_panels/other_listings_component.rb
jmartin-sul: ah yeah, i suspected the non-PURL links might be a problem, which is why i did the combining and filtering in the `#purls` method.  but i like this more direct approach, will try it.

app/components/access_panels/other_listings_component.rb
jmartin-sul: i think so, will do

app/views/catalog/record/_mods_bibliographic.html.erb
jmartin-sul: good call, will do

app/components/access_panels/other_listings_component.rb
cbeer: Sounds good. I think either using type `geo` or explicitly indexing release tags (if there's some reason type geo is problematic) is probably the way to go.

jmartin-sul: would the latter be a searchworks_traject_indexer change?  any reason to investigate that and block merging this before figuring more out?  i don't really have any sense one way or the other of whether or how using type geo could cause issues.

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def earthworks_url
+      @earthworks_url ||= construct_earthworks_url
+    end
+
+    # @return [Array<String>] a tags representing PURL links
+    def purl_links
+      @purl_links ||=
+        if purls.size > 1 # text of the link is the actual URL if there is more than one unique URL, so there aren't multiple links with the same title
```
jmartin-sul, line 16: no idea whether accounting for this possibility is overly complicated, i.e. i have no idea whether we'd ever see multiple PURLs in practice among the MARC and MODS metadata?

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,50 @@
...
+    private
+
+    # @return [Array<String>] URLs representing links to PURL, whether specified in MARC or MODS, and de-duped, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def purls
+      @purls ||= (mods_purls + marc_managed_purls).uniq
+    end
+
+    # @return [Array<String>] URLs representing links to other locations, extracted from MODS metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def mods_purls
+      Array(document['url_fulltext']).find_all { |str| str.start_with?(Settings.PURL_EMBED_RESOURCE) }
```
jmartin-sul, line 34: i wasn't sure if `document['url_fulltext']` was the right field to use.  other possible options i saw were:
* `document.as_schema_dot_org['identifier']` (which seems to have an array of PURLs, but i ended up moving away form using `as_schema_dot_org` at all, so guessing we don't want to reintroduce its usage here)
* `document.mods&.location&.map { |location| location.values }&.flatten` (in testing, this got me an array of `a` tags for the PURLs, but the text of the tag was the URL, which didn't match the mockup's language, and parsing the PURL out of the tag seemed unnecessarily indirect)

jmartin-sul (commented):
left a couple questions inline for reviewers with more SW knowledge than i have.

one other question i'd had at one point was whether i needed to do any filtering in `app/views/catalog/record/_marc_bibliographic.html.erb`.  but i was reminded when looking at the fixtures that the PURL is in the 856, and i didn't see the 856 in that partial's field list.  so i think nothing to do there?

jmartin-sul:
implemented all your feedback, @cbeer, thanks!

jmartin-sul:
i just realized i didn't add a test for the PURLs being filtered out of location display in the MODS bibliographic section, though i did test for it manually.  happy to add that test in either this PR, or a follow on, if that seems worth it.

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    # @return [Array<String>] URLs representing links to other locations, extracted from MARC metadata, e.g. ["https://purl.stanford.edu/hj948rn6493"]
+    def marc_managed_purls
+      document.marc_links.managed_purls.filter_map(&:href) # filter_map because href can be nil, at least in fixture data
```
jmartin-sul, line 39: this seemed safe to me based on looking at similar calls elsewhere in the app, but i can add more safeguards if `marc_links` or `managed_purls` might sometimes be `nil`.

jmartin-sul: mentioning this specifically because i would love for my first searchworks PR to not cause a bunch of unexpected 500 errors 😅

app/components/access_panels/other_listings_component.rb
thatbudakguy: Hmm. I think this actually may be problematic. There are many thousands of items with content type `map` (not `geo`) that are released to EarthWorks, and we allow quite a few other types too. There's also no guarantee that any given `geo` item is actually in EarthWorks. We might need to change the implementation (both here and for schema.org metadata) to look at release tags.

thatbudakguy: @cbeer is it a bad idea to ask purl-fetcher at render time about release info?

dnoneill:
closing in favor of https://github.com/sul-dlss/SearchWorks/pull/4676

PR #4405, opened 2024-08-27
Force boolean parsed queries to use edismax.

cbeer:

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -8,11 +8,18 @@ class SearchBuilder < Blacklight::SearchBuilder
   include BlacklightRangeLimit::RangeLimitBuilder
   include CJKQuery
 
-  self.default_processor_chain += [:add_advanced_parse_q_to_solr, :add_advanced_search_to_solr]
+  self.default_processor_chain += [:add_edismax_advanced_parse_q_to_solr, :add_advanced_search_to_solr]
   self.default_processor_chain += [:database_prefix_search]
   self.default_processor_chain += [:modify_params_for_cjk, :modify_params_for_cjk_advanced]
   self.default_processor_chain += [:consolidate_home_page_params]
   self.default_processor_chain += [:modify_single_term_qf]
+
+  # Tweak advanced search's boolean query output to use edismax instead of dismax
+  def add_edismax_advanced_parse_q_to_solr(solr_params)
+    add_advanced_parse_q_to_solr(solr_params)
+
+    solr_params[:q] = solr_params[:q].gsub('{!dismax', '{!edismax') if solr_params[:q].respond_to?(:to_str) && solr_params[:q].include?('{!dismax')
+  end
 
   # Override range limit to only add parameters on search pages, not the home page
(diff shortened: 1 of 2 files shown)


app/models/search_builder.rb
```diff
@@ -8,12 +8,19 @@ class SearchBuilder < Blacklight::SearchBuilder
...
   include CJKQuery
 
-  self.default_processor_chain += [:add_advanced_parse_q_to_solr, :add_advanced_search_to_solr]
+  self.default_processor_chain += [:add_edismax_advanced_parse_q_to_solr, :add_advanced_search_to_solr]
   self.default_processor_chain += [:database_prefix_search]
   self.default_processor_chain += [:modify_params_for_cjk, :modify_params_for_cjk_advanced]
   self.default_processor_chain += [:consolidate_home_page_params]
   self.default_processor_chain += [:modify_single_term_qf]
 
+  # Tweak advanced search's boolean query output to use edismax instead of dismax
```
jcoyne, line 17: I think this PR looks good.  Can you just add a bit more here why edismax is required or preferable?

cbeer: If it doesn't use edismax, it doesn't get to use the features edismax? 🤷‍♂️ 

I think you discovered this in https://github.com/projectblacklight/blacklight_advanced_search/issues/11.

jcoyne: I'd venture that most of our developers don't know the differences between the parsers without looking it up. It would be a kindness to them, and enable more efficient maintenance, if we can mention the User facing feature(s) that requires each parser in the code comments.

cbeer: This is just important for a consistent user experience.  We don't say anything about it when we configured it for true advanced searches: https://github.com/sul-dlss/SearchWorks/blob/main/app/controllers/catalog_controller.rb#L453

Feel free to suggest some verbiage, but I'm not particularly inclined to make a big deal about it.

cbeer: I'm not sure what to say about it. We've made a decision many years ago that edismax is our preferred query parser and done all of our relevancy testing + tuning with that expectation, so we want to use it for everything because that was the choice we made. I don't know why we'd want some other behavior if you happen to use "AND" or "OR" in a query string and trigger this particularly weird blacklight-advanced-search bug/feature.

corylown: Would a comment like this work?:

```
# Use the same (edismax) query parser for advanced search as we do for regular search
# so that search syntax and relevance ranking are consistent.
```

jcoyne: @corylown that's great.  It also sounds like Chris is aware of some default blacklight behavior that is undesirable and this fixes that.  It would be helpful to have that documented here too.

cbeer, line 17:
```suggestion
  # Tweak advanced search's boolean query output to use edismax instead of dismax.
  # By using the same (edismax) query parser for advanced search as we do for regular search,
  #  the search syntax and relevance ranking are consistent.
```

PR #4472, opened 2024-10-29
Ensure blocked deep pagination options are not shown

### LINKED ISSUE #4467: Suppress unusable pagination links
We limit how deep in the result set the user can page ("You have paginated too deep into the result set. Please contact us using the feedback form if you have a need to view results past page 250."). I believe we used to suppress the last pages (outer window) links:

https://github.com/sul-dlss/SearchWorks/blob/38824515e58a3759c68d9802e12ebcbfedac9a92/config/initializers/blacklight_initialize.rb#L18

But this is no longer working?

Screenshot 2024-10-23 at 12 03 46

thatbudakguy:
Fixes #4467

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -118,6 +118,9 @@ class CatalogController < ApplicationController
       search_series:  :search
     }
 
+    # restrict available pagination options because we block deep pagination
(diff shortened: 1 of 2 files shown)


thatbudakguy:
Before:

Screenshot 2024-10-29 at 11 02 49

After:

Screenshot 2024-10-29 at 11 02 29

cbeer (approved):

PR #4534, opened 2025-01-10
upgrade ruby ci to 3.4

dnoneill:
Needs https://github.com/bblimke/webmock/pull/1091/files to work

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ['3.2', '3.3']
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ['3.2', '3.3']
+        ruby: ['3.3', '3.4']
```
jcoyne, line 14: SW is currently deployed on 3.2. So we should keep that until we've update the deploy

jcoyne (changes requested):

PR #4610, opened 2025-03-14
Update Blacklight Range Limit

### LINKED ISSUE #4546: Upgrade blacklight-range-limit to version9

taylor-steve:
Closes #4546

Waiting on design approval/updated designs.


app/assets/stylesheets/blacklight.scss
```diff
@@ -73,8 +73,26 @@ $remove-icon-font-family:
 }
 
 .range_limit {
+  input {
+    width: 14ch !important;
```
jcoyne, line 77: Is there a reason we need `!important` here?

app/assets/stylesheets/blacklight.scss
```diff
@@ -73,8 +73,26 @@ $remove-icon-font-family:
...
+    display: none;
+  }
+
+  .input-group-append {
+    display: flex;
+  }
+
   .submit {
-    width: 14ch;
+    @extend .btn-sm;
```
jcoyne: Can we avoid using `@extend` so that we can decouple from SASS?

One alternative is to set:
```css
  --bs-btn-padding-y: 0.25rem;
  --bs-btn-padding-x: 0.5rem;
  --bs-btn-font-size: 0.875rem;
  --bs-btn-border-radius: var(--bs-border-radius-sm);
```

PR #4614, opened 2025-03-18
Use Noto Serif Tibetan for Tibetan Unicode range

taylor-steve:
A second attempt at adding Tibetan character support.

The first attempt added an overrides specific font family for the Tibetan unicode-range only. Safari had quirks with unicode-range, not falling back down the chain as expected which led to an odd style issue:
Screenshot 2025-03-17 at 3 11 09 PM

Our current "Source Sans Pro" stack doesn't have the Tibetan range. Adding the range to "Source Sans Pro" worked on all browsers except for Chrome/Edge, which ignored the addition.

Adding the Tibetan range to our current fallback font "Arial Unicode MS" is working as expected on Chrome, Edge, Firefox, and Safari.
Screenshot 2025-03-17 at 3 11 28 PM

### DIFF
--- a/app/assets/stylesheets/fonts.scss
+++ b/app/assets/stylesheets/fonts.scss
@@ -1,0 +1,9 @@
+/* tibetan */
+@font-face {
+  font-family: 'SUL Font Overrides';
+  font-style: normal;
+  font-weight: 100 900;
+  font-display: swap;
+  src: url(https://fonts.gstatic.com/s/notoseriftibetan/v22/gokzH7nwAEdtF9N45n0Vaz7O-pk0wsvbBMgmy2UTva2n.woff2) format('woff2');
+  unicode-range: U+0F00-0FFF;
+}
--- a/app/assets/stylesheets/fonts.scss
+++ b/app/assets/stylesheets/fonts.scss
@@ -1,6 +1,20 @@
+/*
+- Chrome/Edge did not show the Tibetan characters when added to Source Sans Pro
+- Safari had rendering issues when using an "Overrides" specific font-family with just noto sans tibetan
+- So we are using a font-family of "SUL Override" for the tibetan characters, and
+  using the "Arial Unicode MS" as a fallback for the rest of the characters.
+*/
+
+@font-face {
+  font-family: "SUL Override";
+  font-style: normal;
+  font-weight: 100 900;
+  src: local('Arial Unicode MS');
+}
+
 /* tibetan */
 @font-face {
-  font-family: 'SUL Font Overrides';
+  font-family: "SUL Override";
   font-style: normal;
   font-weight: 100 900;
   font-display: swap;
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -6,6 +6,7 @@
 
 // SUL Style Variables/Mixins
 @import 'deja-vu-font';
+@import 'fonts';
 @import 'sul-variables';
 @import 'searchworks-mixins';
 
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -31,7 +31,7 @@ a:not(.btn):not(.dropdown-item):not(.page-link):not(.remove).disabled {
 }
 
 body {
-  font-family: 'SUL Font Overrides', 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
+  font-family: 'Source Sans Pro', 'SUL Override', 'Arial Unicode MS', Helvetica, sans-serif;
 }
 
 .page-item  {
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/fonts.scss
```diff
@@ -0,0 +1,14 @@
+/*
+This is Arial Unicode MS for a reason.
+- Chrome/Edge did not show the Tibetan characters when added to Source Sans Pro
+- Safari had rendering issues when using an "Overrides" specific font-family
+*/
+/* tibetan */
+@font-face {
+  font-family: "Arial Unicode MS";
```
cbeer, line 8: This `font-family` surprises me, but maybe there's magic here I don't understand? Could it be explicit (but ordered after source sans pro?):

```
% git diff
diff --git a/app/assets/stylesheets/base.scss b/app/assets/stylesheets/base.scss
index 3041e927..ca3b678a 100644
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -31,7 +31,7 @@ a:not(.btn):not(.dropdown-item):not(.page-link):not(.remove).disabled {
 }
 
 body {
-  font-family: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
+  font-family: 'Source Sans Pro', 'Noto Serif Tibetan', 'Arial Unicode MS',  Helvetica, sans-serif;
 }
 
 .page-item  {
diff --git a/app/assets/stylesheets/fonts.scss b/app/assets/stylesheets/fonts.scss
index a0fa9bfa..422eb5de 100644
--- a/app/assets/stylesheets/fonts.scss
+++ b/app/assets/stylesheets/fonts.scss
@@ -5,7 +5,7 @@ This is Arial Unicode MS for a reason.
 */
 /* tibetan */
 @font-face {
-  font-family: "Arial Unicode MS";
+  font-family: "Noto Serif Tibetan";
   font-style: normal;
   font-weight: 100 900;
   font-display: swap;
```

taylor-steve: Yes, I'm being super conservative here. Adding it like that would work for all browsers. However, if for some reason Source Sans Pro was unavailble, we'd experience the same Safari issue as before. Adding it to an existing font-family that covers those ranges was the only way to make Safari happy 100% of the time.

However I suppose this case shouldn't ever happen? If Source Sans Pro isn't available/doesn't work, then neither would Noto Serif Tibetan because they are both Google sourced and woff2? Does that track with you? The only danger would be dev changes to the order and review should catch those... Given that I think it'd be fine to change to be more explicit with the font family.

cbeer: Hm. What about splitting the difference between this and your previous stab?

```diff
diff --git a/app/assets/stylesheets/base.scss b/app/assets/stylesheets/base.scss
index 3041e927..9586a1fc 100644
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -31,7 +31,7 @@ a:not(.btn):not(.dropdown-item):not(.page-link):not(.remove).disabled {
 }
 
 body {
-  font-family: 'Source Sans Pro', 'Arial Unicode MS', Helvetica, sans-serif;
+  font-family: 'Source Sans Pro', 'SUL Override', 'Arial Unicode MS', sans-serif;
 }
 
 .page-item  {
diff --git a/app/assets/stylesheets/fonts.scss b/app/assets/stylesheets/fonts.scss
index a0fa9bfa..a5c4619b 100644
--- a/app/assets/stylesheets/fonts.scss
+++ b/app/assets/stylesheets/fonts.scss
@@ -1,11 +1,20 @@
 /*
-This is Arial Unicode MS for a reason.
 - Chrome/Edge did not show the Tibetan characters when added to Source Sans Pro
-- Safari had rendering issues when using an "Overrides" specific font-family
+- Safari had rendering issues when using an "Overrides" specific font-family with just noto sans tibetan
+- So we are using a font-family of "SUL Override" for the tibetan characters, and
+  using the "Arial Unicode MS" as a fallback for the rest of the characters.
 */
+
+@font-face {
+  font-family: "SUL Override";
+  font-style: normal;
+  font-weight: 100 900;
+  src: local('Arial Unicode MS');
+}
+
 /* tibetan */
 @font-face {
-  font-family: "Arial Unicode MS";
+  font-family: "SUL Override";
   font-style: normal;
   font-weight: 100 900;
   font-display: swap;
```

taylor-steve: Yes, that works for all my browsers. I like it, thanks. Did you intend to remove Helvetica?

cbeer: Nope - I think I took it out of the list to test fallbacks a little easier 😀

taylor-steve: Great, I restored it.

PR #4638, opened 2025-03-31
Suppress the long table 'show all' button if we're not actually showing the items.

### LINKED ISSUE #4437: In search results, remove "Show all" button for Hoover Institution records
Reported by Jeanette in [SW-4305](https://jirasul.stanford.edu/jira/browse/SW-4305)

In search results, after clicking on "Check availability", we're displaying a "Show all/Show less" button for _some_ Hoover Institution records. It appears with more regularity on periodicals when there is more than one volume, but it's been hard to find a pattern for when it displays vs. doesn't display. 😢 

The issue with displaying the "Show all/Show less" button is that for Hoover, we instruct the user to "See full record for details" and don't display any item information. (Hoover staff requested this when they implemented Aeon years ago.) When the "Show all/Show less" appears for a Hoover title, it doesn't function (except to toggle between those two choices.)

image

Here is an example search where you can see the button on search results 2 and 3. https://searchworks.stanford.edu/?f%5Bbuilding_facet%5D%5B%5D=Hoover+Institution+Library+%26+Archives&f%5Bformat_main_ssim%5D%5B%5D=Journal%2FPeriodical&f%5Bformat_main_ssim%5D%5B%5D=Database

Notice that the first result is also a periodical with multiple Hoover volumes but "Show all" doesn't appear.  (It displays for the Green results - as it should.) 

In this search: https://searchworks.stanford.edu/?f%5Bbuilding_facet%5D%5B%5D=Hoover+Institution+Library+%26+Archives&f%5Bgenre_ssim%5D%5B%5D=Periodicals results numbers [5](https://searchworks.stanford.edu/view/13782909), [6](https://searchworks.stanford.edu/view/436442), and [12](https://searchworks.stanford.edu/view/in00000088552) also display "Show all" for Hoover holdings. 
image

We should remove the "Show all/Show less" button for all Hoover holdings in search results.

cbeer:
Fixes #4437


app/components/search_result/location_component.html.erb
```diff
@@ -21,7 +21,7 @@
     </tbody>
   <% end %>
   <% locations.each do |location| %>
-    <%= render LongTableComponent.new(node_id: "doc#{document.id}_location_#{location.code}", size: location.items.size) do %>
+    <%= render LongTableComponent.new(node_id: "doc#{document.id}_location_#{location.code}", size: show_items? ? location.items.size : 1) do %>
```
dnoneill, line 24: It seems like the LongTableComponent is used in the library_location_component. It doesn't seem like the longtablecomponent is causing a problem in library_location_component but wouldn't it make more sense for the component to have a render? function?

cbeer: We use `LibraryLocationComponent` on the catalog#show page, which doesn't suppress the items. We only do that on the catalog page (for some reason.. something about building Aeon or finding aid links for the request button?)


I'm not sure I'm following your question about the `render?` function. The `LongTableComponent` (for whatever reason) either renders the contents or the contents plus the show/hide buttons. If we did a `render?` function, wouldn't we completely suppress the contents? It seems way easier just to tell the component how many rows we're actually using 🤷‍♂️ 

I'm also really not sure why `LongTableComponent` does what it does. For some reason, we rewrote (#4059) some pretty trivial javascript decoration with a bunch of new ruby code(and still some javascript), and introduced this regression in the process.

dnoneill: Yeah my eyes are starting to swim from looking at all this 🤷

cbeer:
Replaced by #4639

PR #4645, opened 2025-04-01
Add Honeybadger notification for missing end date

taylor-steve:


app/services/folio/courses.rb
```diff
@@ -22,6 +22,10 @@ def sync_courses!
       # Add or update all courses retrieved from FOLIO
       folio_courses.each { |v|
         end_date = v.dig('courseListingObject', 'termObject', 'endDate')
+
+        # Debugging https://github.com/sul-dlss/SearchWorks/issues/4631
+        Honeybadger.notify('Course has no end date', context: { folio_course: v }) if end_date.blank?
```
cbeer, line 27: We might want to send just a subset of the data (like the stuff before)? I think  we discovered (#4119) the original object has a bunch of data, including identifiers + contact info for the instructors, etc.

taylor-steve: Thanks Chris. I'm hoping the name/number/start date will give me enough to go on.

PR #4654, opened 2025-04-02
add ris download to view page

dnoneill:


app/views/article_selections/_tool_dropdown.html.erb
```diff
@@ -11,12 +11,16 @@
...
 
     <% if @response.documents.any? {|d| d.export_formats.key?(:ris) } %>
-      <%= link_to t('blacklight.tools.refworks_html'), '', onclick: "event.preventDefault(); $(this).next('form').submit();", role: 'menuitem', tabindex: '-1', class: 'refworks dropdown-item' %>
-      <%= form_tag(refworks_export_url(filter: 'RIS Format')) do %>
-        <%= content_tag(:input, nil, type: :hidden, name: :ImportData, value: @response.documents.map(&:export_as_ris).join) %>
-      <% end %>
+      <% if @response.documents.any?(&:eds_ris_export?) %>
+        <%= link_to t('blacklight.tools.refworks_html'), '', onclick: "event.preventDefault(); $(this).next('form').submit();", role: 'menuitem', tabindex: '-1', class: 'refworks dropdown-item' %>
+        <%= form_tag(refworks_export_url(filter: 'RIS Format')) do %>
+          <%= content_tag(:input, nil, type: :hidden, name: :ImportData, value: @response.documents.map(&:eds_export_as_ris).join) %>
```
dnoneill, line 17: I am not actually sure this is how we want to do this. I am waffling.

dnoneill: found a much better solution.

dnoneill:
@dbranchini Is the RIS download text and icon okay or do we want to update that to Zotero? Also I am assuming we want the same text for all RIS downloads? We don't want to differentiate what we build vs what we get from eds.

jcoyne:
It looks like the option order changes depending on the page you are on.  Should that be consistent across the views?

dnoneill:
Good catch. Fixed

app/models/concerns/ris_mapping.rb
```diff
@@ -0,0 +1,42 @@
...
+      LA: 'language',
+      Y1: 'pub_date',
+      CY: 'pub_country',
+      AB: 'summary_display',
+      C3: 'physical',
+      CN: 'lc_assigned_callnum_ssim',
+      H1: 'building_facet',
+      SN: 'isbn_display',
+      DP: proc { searchworks_url },
+      UR: proc { searchworks_url },
```
taylor-steve, line 39: I think we should consider changing these to plain URLs, rather than markup. I'm assuming you're copying what EDS was doing, but I think it could be more compatible with other tools that use RIS. For reference, Zotero drops the markup on RIS export. There's no difference during import either.

dnoneill: Yeah that is what is done in https://searchworks.stanford.edu/view/6024828. I can remove the mark up

taylor-steve (approved):

PR #4660, opened 2025-04-04
Remove unused .fa-bookmark-o css rule

jcoyne:


cbeer:
It looks like we stopped using that icon when we implemented articles+: https://github.com/sul-dlss/SearchWorks/commit/a5868fccae069fa77eceeeef81c46d94ef40e45a

PR #4667, opened 2025-04-04
move finding_aid logic from indexer to Searchworks

### LINKED ISSUE #4606: Refactor the Searchworks index links logic for finding aids to no longer use the indexed `finding_aid` attribute
The `finding_aid` attribute of the `marc_links_struct` object should be a display property (based on the href matching OAC (and soon archives.stanford.edu)) and not determined at index time.

dnoneill:
closes #4606

### DIFF
--- a/app/models/concerns/index_links.rb
+++ b/app/models/concerns/index_links.rb
@@ -25,13 +25,13 @@ module IndexLinks
     end
 
     def to_index_link
-      Links::Link.new(link_struct.merge({ link_text:, href: }))
+      Links::Link.new(link_struct.merge({ link_text:, href:, finding_aid: finding_aid? }))
     end
 
     private
 
     def link_text
-      if link_struct[:finding_aid]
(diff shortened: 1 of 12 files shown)


app/models/links/link.rb
```diff
@@ -3,15 +3,15 @@
 class Links
   class Link
     include ActionView::Helpers::TagHelper
+    include IndexLinks::Structures
```
jcoyne, line 6: These behaviors only apply to IndexLinks, and not to MarcLinks or EdsLinks, so it seems to me this should say in IndexLinks.

jcoyne (changes requested):

app/models/links/link.rb
dnoneill: I don't understand what you are saying. Are you saying finding_aid? should go into a different concern?

jcoyne (approved):

PR #4668, opened 2025-04-04
Add per-environment folio configs

### LINKED ISSUE #4621: Update `Folio::Types` to support reading environment-specific configurations

jcoyne:
By settings Settings.folio.config_set to test or prod (default), you can select the set of configs you want to use

Fixes #4621

If accepted, we should write a follow on task that includes syncing the test configs as we do for the prod configs.


dnoneill:
@jcoyne This doesn't account for stage which I think we want to support? I am wondering if the best way of doing this is to have a rake task bin/rake folio:update_types_cache set it's config based on the server?

jcoyne:
The searchworks "stage" environment is pointing at the Folio "test" environment. This PR handles that.

PR #4672, opened 2025-04-07
add popover to stanford only icon

### LINKED ISSUE #4593: Add popover to Stanford restricted S icon for online items with restricted access
See screenshot. With the component library, we are moving to dark or white popovers (triggered on click) instead of turquoise tooltips (triggered on hover). Popovers are better for screenreader users. However, if it's easier to implement with a turquoise popover prior to SW4.0, then that's okay, too. 

aria-label on the "S" icon should be "Stanford-only" and the popver language should be "Available to Stanford-affiliated users only. Log in to access."

Image

---

[**Direction for SW4.0**](https://www.figma.com/design/fsVWCX66bpvFZGKEHNRBS5/SearchWorks-4.0?node-id=538-11078&t=lYsEO9ZdifpqMyFn-4) is the same.

dnoneill:
closes #4593 

Screenshot 2025-04-10 at 2 24 20 PM


app/components/stanford_only_icon_component.rb
```diff
@@ -0,0 +1,29 @@
...
+      '8.50097418 C8.36718334,7.79003736 7.77050422,6.71093682 6.79296609,6.71093682 C6.29784937,6.71093682 5.76464675,6.97753813 5.76464675,7.82812326 C5.76464675,' \
+      '8.61523188 6.76757549,9.1865204 7.60546531,9.71972302 C8.45605044,10.2656209 9.47167447,10.5956987 9.47167447,12.2841737 C9.47167447,13.6806567 8.10058203,' \
+      '14.6328043 6.70409898,14.6328043 C5.95507626,14.6328043 5.05370993,14.4804607 4.29199191,14.3027265 C4.1396483,13.7314379 4,12.550775 4,12.1953066 C4.0253906,' \
+      '12.1191348 4.16503891,12.0683536 4.22851541,12.0683536 C4.35546841,12.0683536 4.62206972,12.042963 4.63476502,12.1445254 C4.74902272,12.8300716 5.51074075,' \
+      '13.9091721 6.45019298,13.9091721 C7.1611298,13.9091721 7.70702771,13.5410084 7.70702771,12.5253844 C7.70702771,11.7636664 6.56445068,11.1923779 5.92968566,' \
+      '10.7607376 C5.00292873,10.1386679 4.0761718,9.61816061 4.0761718,8.29784937 Z'
+  end
+
+  def call
+    tag.svg(class: 'stanford-only', style: 'margin-left: 5px;',
```
cbeer, line 16: Shouldt the popover toggle be a `<button>` for accessibility?

dnoneill: Or tabindex="0" I think is also okay. Do you have a preference?

cbeer: I think a button would be much clearer. It looks like it'd probably fine everywhere except:

```
<a href="https://library.stanford.edu/chat" class="dropdown-item" role="menuitem" title="limited to Stanford community">Chat with us  <%= render StanfordOnlyIconComponent.new %><span class="sr-only"> (limited to Stanford community)</span></a>
```

dnoneill: Yeah I can move it out of the link so it isn't an issue.

dnoneill: Wait no I can't. I don't think it's that much of an issue. The current version doesn't have a popover

app/components/stanford_only_icon_component.rb
```diff
@@ -0,0 +1,31 @@
...
+      '12.1191348 4.16503891,12.0683536 4.22851541,12.0683536 C4.35546841,12.0683536 4.62206972,12.042963 4.63476502,12.1445254 C4.74902272,12.8300716 5.51074075,' \
+      '13.9091721 6.45019298,13.9091721 C7.1611298,13.9091721 7.70702771,13.5410084 7.70702771,12.5253844 C7.70702771,11.7636664 6.56445068,11.1923779 5.92968566,' \
+      '10.7607376 C5.00292873,10.1386679 4.0761718,9.61816061 4.0761718,8.29784937 Z'
+  end
+
+  def call
+    tag.button(class: 'stanford-only btn btn-link p-0 align-baseline', style: 'margin-left: 5px;',
+               data: { turbo: false, toggle: 'popover', placement: 'right', content: 'Available to Stanford-affiliated users only. Log in to access.' },
+               aria: { label: 'stanford only' }) do
+      tag.svg(width: '13px', height: '13px', viewBox: '0 0 13 13') do
```
cbeer, line 19:
```suggestion
      tag.svg(width: '13px', height: '13px', viewBox: '0 0 13 13', aria: { hidden: true }) do
```

Now that there's a button (with the aria-label), I think the svg tag needs to be hidden?

app/components/stanford_only_icon_component.rb
cbeer: I thought it isn't valid HTML5 to nest a button inside a link.

dnoneill: Okay. I think I have a solution.

app/components/stanford_only_popover_component.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class StanfordOnlyPopoverComponent < StanfordOnlyIconComponent
```
jcoyne, line 3: I don't think these should inherit as this is not a natural inheritance relationship.  The popover component can contain a StanfordOnlyIconComponent, instead right?

jcoyne (changes requested):

app/components/stanford_only_popover_component.rb
dnoneill: Then the StanfordOnlyIconComponent would need to take a variable because the aria variables are different if it is an svg vs a button.

jcoyne: How about always making the SVG `aria-hidden="true"` and wrap it with a `span aria-label="stanford only"` when not used in a button.

jcoyne (approved):

jcoyne:
Does Sarah or Darcy need to review this?

dnoneill:
Yes

dnoneill:
done

app/assets/stylesheets/bootstrap-variables.scss
```diff
@@ -88,6 +88,12 @@ $padding-xs-horizontal: 0.25rem;
 $nav-link-padding: 0.5rem 1rem;
 $grid-float-breakpoint: 768px;
 
+// popovers
+$popover-bg: $stanford-black;
+.popover-body {
+  color: #fff!important;
```
jcoyne, line 94: can we use `$popover-body-color` here?  Why do we need `!important`?

dnoneill: Yeah this happened because the stackoverflow I found did not include that variable. I was looking for it everywhere except the source code which is where I should have gone first. Thank you that was driving me crazy.

jcoyne (approved):

dbranchini:
Thank you @dnoneill !

PR #4676, opened 2025-04-08
add earthworks, sdr sidebar

### LINKED ISSUE #4312: Add link to view items listed in Earthworks

dnoneill:
closes #4312


app/views/catalog/record/_mods_bibliographic.html.erb
```diff
@@ -24,6 +24,6 @@
   <%= render ModsDisplay::FieldComponent.new(field: identifier, label_html_attributes: { class: 'col-md-3' }, value_html_attributes: { class: 'col-md-9' }) %>
 <% end %>
 
-<% document.mods.location.each do |location| %>
+<% document.mods.location.reject { |location| location.values.detect { |loc| /#{Settings.PURL_EMBED_RESOURCE}#{document.druid}/.match?(loc) } }.each do |location| %>
```
jcoyne, line 27: Can you move this into a component or a helper?  We should try to keep logic like this out of the ERBs.  Also please add a comment about what it's doing. It's not immediately clear.

app/components/access_panels/other_listings_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= render @layout.new do |component| %>
+  <% component.with_title { 'Also listed in' } %>
+  <% component.with_body do %>
+    <% if earthworks_url.present? %>
+      <p><%= link_to('Earthworks', earthworks_url) %></p>
+    <% end %>
+    <% if purl_url.present? %>
```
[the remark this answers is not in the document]
dnoneill: good point. thanks.

app/components/access_panels/other_listings_component.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def purl_url
+      "#{Settings.PURL_EMBED_RESOURCE}#{document.druid}"
+    end
+
+    private
+
+    def construct_earthworks_url
+      response = Net::HTTP.get_response(URI.parse("#{purl_url}.meta_json"))
```
jcoyne, line 20: Can we do this on the front end so that it doesn't block page drawing?

dnoneill: Done

app/javascript/controllers/meta_json_check_controller.js
```diff
@@ -0,0 +1,25 @@
+// controllers/meta_json_check_controller.js
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = [ "earthworks" ]
```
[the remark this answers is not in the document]
dnoneill: Nope. Cruft from a refactor. removed.

spec/components/record/mods_bibliographic_component_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe "Object bibliographic" do
+    let(:document) { SolrDocument.new(modsxml: mods_everything) }
+    let(:component) { described_class.new(document:) }
+
+    before do
+      render_inline(component)
+    end
+
+    it "should display titles" do
```
jcoyne, line 17: Using "Should" is typically a violation of `RSpec/ExampleWording` I know it's not enabled on this project, but I think it's best to follow that as we typically use it everywhere.

app/components/record/mods_bibliographic_component.html.erb
```diff
@@ -1,5 +1,3 @@
-<% document ||= @document %>
```
jcoyne, line 1: Should this file be in this PR?  I don't understand how this relates.

jcoyne:
This seems good. Does Sarah need to approve it too?

dnoneill:
@jcoyne I would assume no because it wasn't opened by her or Darcy but I can double check.

dnoneill:
Should be okay to merge: https://github.com/sul-dlss/SearchWorks/issues/4312#issuecomment-2794522378

spec/components/record/mods_bibliographic_component_spec.rb
```diff
@@ -0,0 +1,40 @@
+# encoding: UTF-8
```
jcoyne, line 1: Please remove this line

app/javascript/controllers/meta_json_check_controller.js
```diff
@@ -0,0 +1,25 @@
...
+    field: String,
+    linkUrl: String,
+    linkLabel: String
+  }
+
+  connect() {
+    fetch(this.urlValue)
+      .then((response) => response.json())
+      .then((data) => this.updateElement(data))
+      .catch(console.error)
```
jcoyne, line 16: Do we need to call the console.error, or is just passing the method handle sufficient?

jcoyne: I was thinking something like:
```js
  console.error('meta_json fetch failed',error);
```

dnoneill: sounds good to me. updated.

jcoyne (approved):

PR #4696, opened 2025-04-09
New design for no-results page

### LINKED ISSUE #4350: Rethink "no results" page, especially emphsis on "CHECK SPELLING, USE FEWER WORDS TO START, OR REMOVE LIMITS"
We received SearchWorks feedback from a graduate student about search results. ([SW-4346](https://jirasul.stanford.edu/jira/browse/SW-4346)) It turned out that there were spelling errors and additional words in their search. 

It encouraged a discussion in Gryphon Search about search recall and emphasizing precision over broader returns that would account for spelling errors. Gryphon Search decided to keep precision over broad recall. 

It was also discussed that perhaps the "[0 results](https://searchworks.stanford.edu/?search_field=search&q=%3Blakjdlkl%3Bda)" page could use a rethink and redesign, particularly the on-screen encouragement to "CHECK SPELLING, USE FEWER WORDS TO START, OR REMOVE LIMITS"

Relatedly, it was discovered that Bento doesn't have the equivalent of a "no results" page if all apps return no results.

jcoyne:
Fixes #4350

Screenshot 2025-04-14 at 10 45 42 AM


dnoneill:
@jcoyne needs a rebase

dnoneill:
I believe mini bento's heading should be in line with the other two boxes (I think the bento just needs to be moved up)
Design
Screenshot 2025-04-10 at 4 28 37 PM
Current
Screenshot 2025-04-10 at 4 28 37 PM

dnoneill:
mini bento should have a border-radius of 4px

dnoneill:
Guessing you already know that there needs to be a stanford icon next to Stanford affiliates, which is why this might be blocked by the popover.

dnoneill:
Remove filters should be in line with selected facets
 Design
Screenshot 2025-04-10 at 4 34 39 PM
Branch
Screenshot 2025-04-10 at 4 34 39 PM

dnoneill:
I also have a question of what the expected behavior is when the mini bento is closed. Right now it just closes and doesn't resize the view. I am wondering if it is supposed to look like that.

dnoneill:
Think there might need to be more padding but maybe not? The design has 40px padding between the body and the footer.
Design
Screenshot 2025-04-10 at 4 39 30 PM
Current
Screenshot 2025-04-10 at 4 39 30 PM

dbranchini:
First note -- we should be using the new footer (with the new logo) (and same goes for header). When I started to design this, I captured the page before they were replaced. 

From the screenshots above, I don't see a problem with the padding or margins, but did the wrong screenshot get added for the "current'?  cc: @dnoneill

dnoneill:
Yeah wrong screenshot was added. I added the correct one.

dnoneill:
@jcoyne @dbranchini  I don't know if this is the expected behavior but right now the chat with a librarian is greyed out even though it is currently open. Shouldn't that be available? The Figma have it as a link if the hours are open even if the user isn't logged in? 
Screenshot 2025-04-14 at 3 58 54 PM

dnoneill:
Screenshot 2025-04-14 at 4 03 23 PM

dnoneill:
Also it would probably be good if the link `All reference hours` points to https://library.stanford.edu/libraries/cecil-h-green-library#additional-hours (add #additional-hours) otherwise the hero image takes up most of the screen when the user opens the page.

jcoyne:
@dnoneill was it disable there because you are not signed in?

dbranchini:
> @jcoyne @dbranchini I don't know if this is the expected behavior but right now the chat with a librarian is greyed out even though it is currently open. Shouldn't that be available? The Figma have it as a link if the hours are open even if the user isn't logged in? Screenshot 2025-04-14 at 3 58 54 PM

"Chat with a librarian" should be a blue link when the hours are open and a **_disabled_** gray link when it is closed.

dbranchini:
> Also it would probably be good if the link `All reference hours` points to https://library.stanford.edu/libraries/cecil-h-green-library#additional-hours (add #additional-hours) otherwise the hero image takes up most of the screen when the user opens the page.

This is a good idea. I forgot we added the jump links to this page.

dbranchini:
> @jcoyne Yes. But the figma seems to indicate it shouldn't require you to be logged in for it to be clickable. The requiring log in might be due to a shared component? I think this is really a @dbranchini on what is the correct format. Screenshot 2025-04-14 at 4 20 42 PM

This is true. You shouldn't have to be logged in for it to be clickable -- mainly because the login to searchworks and the login for chat are two separate logins anyway.

app/assets/stylesheets/modules/zero-results.css
```diff
@@ -0,0 +1,61 @@
+.zero-results {
+  .border-before {
+    border-left: 1px solid gray;
```
dnoneill, line 3: This should be Primary/Black 20

app/components/search_result/chat_with_librarian_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SearchResult
+  class ChatWithLibrarianComponent < ViewComponent::Base
```
dnoneill, line 4: Why is the exact same code defined twice? Could we use inheritance so we aren't replicating the same code?

dnoneill: Also the app/components/search_result/chat_with_librarian_component.rb and app/components/feedback/chat_with_librarian_component.html.erb seem incredibly similar except for the popover icon, link text, and the header class. Is there a reason they are seperate?

app/views/articles/_search_results.html.erb
```diff
@@ -22,7 +22,7 @@
         <%= render 'search_header' %>
         <%= render 'research_starters' %>
         <%= render_document_index @response.documents %>
-        <%= render 'alternate_catalog' if show_alternate_catalog? %>
+        <%= render 'alternate_catalog', close: true if show_alternate_catalog? %>
```
dnoneill, line 25: This might be a future problem and not applicable to this PR, but I think this is a class called alert-dismissible in bootstrap5 and we are going to get rid of the close button in bootstrap 5 so I am wondering if we should rename close to dismissible?

app/components/search_result/chat_with_librarian_component.rb
jcoyne: These have very different layouts.  I don't think this is a direct inheritance relationship.

dnoneill: But the RB files are exactly the same. Couldn't you keep the different layouts and have the RB files inherit the same functions?

app/assets/stylesheets/modules/zero-results.css
jcoyne: updated this.

app/components/search_result/chat_with_librarian_component.rb
jcoyne: We could, but we shouldn't because this isn't a natural inheritance relationship. https://medium.com/@b.stoilov/why-you-should-think-twice-before-using-inheritance-573545dcd8e1.

saseestone:
re: The "Clear all" next to "Remove filters" - I assumed that this would only clear filters I had selected, and that it would NOT clear the search term I typed in. @dbranchini is the intention to clear everything and function more like a "Catalog start"? 

Or should the "Clear all" simply remove selected filters and keep the search term?

saseestone:
Except for my question above about the function of the "Clear all" link, I think this look great and is ready from my POV.

jcoyne:
@saseestone the search term typed in is one of the filters that Blacklight uses.  Here's a view of just a search term:
Screenshot 2025-04-16 at 6 49 29 AM

dbranchini:
Good catch, @saseestone , it should just remove the filters shown below the "Remove filters" with clear all link (so not the search term).

jcoyne:
@dbranchini what if they want to keep the other filters, but change the search term?

dbranchini:
Just the typical replace the words in the search input. I think that's fine.

spec/components/search_result/no_result/filters_component_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+RSpec.describe SearchResult::NoResult::FiltersComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
+  #   expect(
+  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
+  #   ).to include(
+  #     "Hello, components!"
+  #   )
+  # end
```
taylor-steve, line 14: Is this a useful placeholder or can we remove it?

taylor-steve (approved):

PR #4711, opened 2025-04-11
have bound-with children get permanent location library for display

### LINKED ISSUE #4349: Boundwith child should inherit Library of principle record (in addition to effective location)
Reported by Jeanette via SW feedback ([#4347](https://jirasul.stanford.edu/jira/browse/SW-4347))

On child records, we're displaying the effective location of the item record (from the principle record) but the Library of the child is defaulting to the library of the effective location (often SUL if the item is in a processing location).  We should be using the Library of the principle as well.

Example records can be hard to find because we appear to be handling this correctly for items in SAL3 Shelved Elsewhere. The example record provided by Jeanette is https://searchworks.stanford.edu/view/in00000126045

dnoneill:
closes #4349

### DIFF
--- a/app/services/folio/locations.rb
+++ b/app/services/folio/locations.rb
@@ -21,6 +21,11 @@ module Folio
       location && location['details']
     end
 
+    def self.library_id(code:)
+      location = find_by(code:)
+      location && location['libraryId']
+    end
+
     def self.find_by(code:)
       instance.data[code]
     end
--- a/lib/holdings/item.rb
+++ b/lib/holdings/item.rb
@@ -170,9 +170,17 @@ class Holdings
       availability_class.present? && availability_class == 'In_process'
     end
 
+    def bound_with_location
+      library_id = Folio::Locations.library_id(code: item_display[:effective_permanent_location_code])
+      Folio::Types.libraries[library_id]
+    end
+
     private
 
     def standard_or_zombie_library
+      # we want the parent library, not the child record's library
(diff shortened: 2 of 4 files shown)


cbeer:
Do we need to do all these work-arounds, or is it enough to initialize our bound-with items with the same data we use for regular items?

```
diff --git a/app/models/concerns/solr_holdings.rb b/app/models/concerns/solr_holdings.rb
index c535a78d..04c6b42c 100644
--- a/app/models/concerns/solr_holdings.rb
+++ b/app/models/concerns/solr_holdings.rb
@@ -57,10 +57,10 @@ module SolrHoldings
   end
 
   def bound_with_folio_items
-    @bound_with_folio_items ||= Array(holdings_json['holdings']).filter_map do |holdings_record|
-      next if holdings_record['boundWith'].blank?
+    @bound_with_folio_items ||= folio_holdings.filter_map do |holdings_record|
+      next unless holdings_record.bound_with_parent
 
-      Folio::Item.from_dynamic(holdings_record.dig('boundWith', 'item'))
+      Folio::Item.from_dynamic(holdings_record.bound_with_parent.item, holdings_record: holdings_record)
     end
   end
```

dnoneill:
@cbeer That works and is cleaner. I am closing this in favor of https://github.com/sul-dlss/SearchWorks/pull/4724. I implemented your suggestion and added tests to make sure it works correctly.

PR #4724, opened 2025-04-15
initialize bound-withs items to include holdings data

### LINKED ISSUE #4349: Boundwith child should inherit Library of principle record (in addition to effective location)

dnoneill:
closes #4349

### DIFF
--- a/spec/models/concerns/solr_holdings_spec.rb
+++ b/spec/models/concerns/solr_holdings_spec.rb
@@ -10,6 +10,53 @@ RSpec.describe SolrHoldings do
     expect(document.holdings).to be_a Holdings
   end
 
+  describe '#bound_with_folio_items' do
+    let(:document) {
+      SolrDocument.new(holdings_json_struct: [{ 'holdings' => [
+        {
+          'id' => '123456', 'holdingsType' => { 'code' => 'Bound-with' },
+          'location' => {
+            'effectiveLocation' => {
+              'id' => "158168a3-ede4-4cc1-8c98-61f4feeb22ea",
+              'code' => "SAL3-SEE-OTHER",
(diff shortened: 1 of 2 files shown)


spec/models/concerns/solr_holdings_spec.rb
```diff
@@ -10,6 +10,52 @@
...
+      expect(document.bound_with_folio_items.first.id).to eq '1234580'
+    end
+
+    it 'provides populates the correct locations' do
+      # The bound with item effectiveLocation should populate the effective_location
+      expect(document.bound_with_folio_items.first.effective_location.code).to eq 'GREEN'
+      # The holding's record's effectiveLocation should populate the permanent location
+      expect(document.bound_with_folio_items.first.permanent_location.code).to eq 'SAL3-SEE-OTHER'
+    end
+  end
```
taylor-steve, line 57: Maybe pick one of `provides populates`.

I think there's an argument to be made here to rework the comments as `it` tests. I feel they are more informative than the current statement.

dnoneill: separated into separate it statements.

taylor-steve (approved):

Issue #4701, opened 2025-04-10
500 page


dbranchini:
### Designs:
[Figma with specifications](https://www.figma.com/design/fsVWCX66bpvFZGKEHNRBS5/SearchWorks-4.0?node-id=467-33081&t=coiiTTWNVGkeMORe-4)

### Prototype:
[Figma in browser view](https://www.figma.com/proto/fsVWCX66bpvFZGKEHNRBS5/SearchWorks-4.0?page-id=447%3A3400&node-id=467-33081&viewport=-22%2C-34%2C0.28&t=DvlbjgTNSX7iOt9T-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=467%3A33081&show-proto-sidebar=1)

### Design notes
- Use site's current header/footer 
- See related ticket for 404 errors
- Heading: 

> Internal server error (500)

- Message: 

> Something went wrong on our end. Please check the [system status](url) for updates or try again later. 
> 
> If you believe this is an error, please report it via the [feedback form](url) or [contact a librarian](url) for help.

- system status goes to https://library-status.stanford.edu/
- feedback form opens the form
- contact a librarian goes to https://library.stanford.edu/contact-us

### Images needed

![Image](https://github.com/user-attachments/assets/39ebc524-d759-4f0c-8100-09d95170a4bb)

jcoyne:
I don't believe searchworks shows a 500 page. The loadbalancer redirects to the system status page, right?

dbranchini:
We discussed during standup on 4/10 that it would be a better user experience to show a 500 error and give them the option to check the system status page so the user understands what happened. It's a bit jarring to be auto-redirected to a different app.

jcoyne:
I'm not sure it's possible to use the sites existing header for this.  The header depends on the server working (to process login for example), but if the server isn't working we can't do that.  We'll need a custom header for this with most of the features paired back.

taylor-steve:
We may need to generate a static page to send to Julian

PR #4738, opened 2025-04-16
add 500 page

dnoneill:
part of #4701 

Screenshot 2025-04-30 at 12 51 35 PM


saseestone:
Where does the link for the "feedback form" go to? I thought we could only display the feedback form if SearchWorks was up.

dbranchini:
Yes, that's true, so we could only provide the link to contact a librarian. In that case, we'd modify the second sentence to read:

> "If you believe this is an error, please contact a librarian for help." 

After our conversation yesterday, it sounds like sometimes these errors occur and the app is still running. Maybe we keep the original sentence with the link to the feedback form in those cases? However, I'm not sure if it's a good presumption that the feedback form will be working if parts of the app are not and that's why the user got this error.

dnoneill:
This is just for when the app is up in running. I need to double check with Julian if we can detect when the 500 is based on the app being down because I think that is going to affect how this is approached

config/environments/production.rb
```diff
@@ -11,7 +11,7 @@
   # Eager load code on boot for better performance and memory savings (ignored by Rake tasks).
   config.eager_load = true
 
-  # handle 404 exception in routes, not public/404.html
+  # handle specificied exception in routes, not public/code.html
```
cbeer, line 14:
```suggestion
  # handle specified exception in routes, not public/code.html
```

app/components/status/error_page_component.html.erb
```diff
@@ -5,5 +5,8 @@
   <div class="ml-4 fw-600">
     <h2 class="h3"><%= header %> (<%= code %>)</h2>
     <%= message %>
+    <p>If you believe this is an error, please report it via the
```
cbeer, line 8: @dbranchini : is the language here a little weird? It's definitely an error, so I'm not sure what we're asking the user to do

dbranchini:
Good point, @cbeer ! Ugh... I copied that from the 404. I changed it to:

> Something went wrong on our end. Please check the [system status](url) for updates or try again later. 

@dnoneill , can we center the message and image vertically? So the heading, "Internal server error (500)" doesn't sit that far below the 500 image?

dnoneill:
PR updated.

cbeer (approved):

PR #4739, opened 2025-04-16
Prioritize display of Archives finding aid links

### LINKED ISSUE #4605: Show OAC finding aids links under "more options" on the record show page
If there's an OAC finding aid link that's not the only finding aid link it should display under "More options". Example: https://searchworks.stanford.edu/view/12441511

An OAC finding aid is a link that contains `oac.cdlib.org`.

If it's the only link, leave it as is.

If there is no "More options", e.g., https://searchworks.stanford.edu/view/zy483zc0975, then don't display the OAC finding aid link if there is another finding aid link.

taylor-steve:
Closes #4605.

This PR adds a sort to the access panel finding_aid links to prioritize Archives links. This will naturally change what is being displayed in the all the places we use `access_panel_links.finding_aid.first`, but I believe that is in line with our desires.

I thought about finding aids in terms of "a single preferred link" versus "the rest" which I like. I do think drawing a harder line is an option as well. That we have 0 or 1 Archives finding aid and 0 or 1 OAC finding aids and assume anything leftover is garbage data? I don't think we have other sources of finding aids at the moment.


app/models/concerns/marc_links.rb
```diff
@@ -52,6 +52,20 @@ def finding_aid?
       link_struct[:material_type]&.downcase&.include?('finding aid') || link_struct[:note]&.downcase&.include?('finding aid')
     end
 
+    def archives_finding_aid?
+      finding_aid? && link_struct[:href]&.include?('//archives.stanford.edu')
+    end
+
+    def finding_aid_link_text
+      archives_finding_aid? ? 'Archival Collections at Stanford' : 'Online Archive of California'
+    end
```
cbeer, line 61: If we're now distinguishing between finding aid hosts, maybe it's time to extract something like `oac_finding_aid?` too?

```suggestion
    def archives_finding_aid?
      finding_aid? && link_struct[:href]&.include?('//archives.stanford.edu')
    end

    def oac_finding_aid?
      finding_aid? && link_struct[:href]&.include?('oac.cdlib.org')
    end

    def finding_aid_link_text
      if archives_finding_aid?
        'Archival Collections at Stanford'
      elsif oac_finding_aid?
        'Online Archive of California'
      else
        link_struct[:link_text] || link_host
      end   
    end
```

app/models/links.rb
```diff
@@ -24,7 +24,7 @@ def supplemental
   end
 
   def finding_aid
-    all.select(&:finding_aid?)
+    all.select(&:finding_aid?).sort_by { |x| [x.href.present?, x.sort.present? ? 0 : 1, x.sort] }
```
cbeer, line 27: Could `Links::Link` provide a the default sort?

taylor-steve: Okay the `a the` is throwing me off more than it should 😄 

Are you taking issue with me somewhat abusing the existing 856 x sort mechanism that is in place? 

Or are you asking if we can literally provide a default sort value in `Links::Link#initialize`? In which case wouldn't that impact the current behavior of `managed_purls`?

PR #4745, opened 2025-04-17
rename bound_with? to bound_with_child?

dnoneill:


lib/holdings/item.rb
```diff
@@ -114,8 +114,13 @@ def bound_with_parent
       item_display[:bound_with]
     end
 
-    # @return [Bool] true if this is a bound-with child
+    # @return [Bool] true if this is a bound-with
     def bound_with?
+      bound_with_child? || bound_with_principal?
```
jcoyne, line 119: Do you also need to change the caller in `SearchResult::LocationComponent`?

jcoyne (changes requested):

lib/holdings/item.rb
jcoyne: My understanding was that this PR is just renaming and not affecting functionality.  Is that incorrect?  There is no linked issue or explanation in the commit message to indicate otherwise.

dnoneill: okay I can do that.

jcoyne (approved):

PR #4748, opened 2025-04-18
Consolidate finding aid call numbers in access panel

### LINKED ISSUE #4727: Show base call number only in availability panel for records with finding aids
Split from #4605 

Show the base call number only in the availability panel for records with finding aids. See the ticket referenced above for more context and pink circled area on the designs below. 

Image
Image

taylor-steve:
Closes #4727

### DIFF
--- a/app/components/access_panels/location_item_component.html.erb
+++ b/app/components/access_panels/location_item_component.html.erb
@@ -1,6 +1,6 @@
 <%= tag.tr class: classes, data: { barcode: item.barcode } do %>
   <td class="callnumber">
-    <%= item.callnumber %>
+    <%= callnumber %>
   </td>
   <% if render_item_note? %>
     <td class="public-note">
--- a/app/components/access_panels/library_location_component.html.erb
+++ b/app/components/access_panels/library_location_component.html.erb
@@ -24,6 +24,6 @@
     </tr>
   </thead>
   <tbody data-long-table-target="body">
-    <%= render AccessPanels::LocationItemComponent.with_collection(location.items, document:, classes: 'availability-item', render_item_level_request_link: !location_request_link.render?) %>
+    <%= render AccessPanels::LocationItemComponent.with_collection(location.items, document:, classes: 'availability-item', render_item_level_request_link: !location_request_link.render?, consolidate_for_finding_aid: consolidate_items?) %> 
   </tbody>
 <% end %>
--- a/app/components/access_panels/library_location_component.rb
+++ b/app/components/access_panels/library_location_component.rb
@@ -18,5 +18,15 @@ module AccessPanels
     def location_request_link
       @location_request_link ||= LocationRequestLinkComponent.for(document:, library_code: library.code, location:)
     end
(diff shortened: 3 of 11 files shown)


app/components/access_panels/location_item_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= tag.tr class: classes, data: { barcode: item.barcode } do %>
   <td class="callnumber">
-    <%= item.callnumber %>
+    <%= callnumber %>
```
dnoneill, line 3: Is there a reason this can't be just changed to item.truncated_callnumber and remove the 
```
def callnumber item.truncated_callnumber end
```
in app/components/access_panels/location_consolidated_item_component.rb

app/components/access_panels/location_consolidated_item_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module AccessPanels
+  class LocationConsolidatedItemComponent < LocationItemComponent
+    def callnumber
+      item.truncated_callnumber
```
dnoneill, line 6: Is there a reason this needs to be a separate component? Can't we just put the call number function in LocationItemComponent. Also it could be easily done by moving consolidated_items and consolidate_items? into the LocationItemComponent

taylor-steve: I'm not sure I'm understanding. `LocationItemComponent` is responsible for displaying a single row representing a single item. If we moved the item consolidation behavior out of `LibraryLocationComponent` wouldn't the `with_collection` call render `n` rows instead of our desired 1 (or whatever they get reduced to)?

dnoneill: Right sorry. I wasn't thinking. Is there a reason you can't add a function to LocationItemComponent so you don't have to have this component?

    def callnumber
      return item.truncated_callnumber if document.has_finding_aid? && policy.aeon_pageable?
      item.callnumber
  end

taylor-steve: `LocationItemComponent` does already have that method, via the delagate. You are not wrong, your approach would work. It would mean we would no longer be able to use `LocationItemComponent` to show an aeon pageable item with its complete call number.

My thinking is this: `LocationItemComponent` and `LocationConsolidatedItemComponent` represent different concepts.
* `LocationItemComponent` represents one item's display for the given location
* `LocationConsolidatedItemComponent` represents one item being used as a stand-in for potentially many items for the given location

I don't know that the logic for making that choice is the responsibility of `LocationItemComponent`

cbeer: I'm fine with `LocationConsolidatedItemComponent`; if nothing else, this approach opens things for other item display patterns.  

I think the thing I don't love about this approach is `items` sometimes returning all the items and somethings returning a subset, but maybe it's fine to wait until we have another example of alternate display?

PR #4749, opened 2025-04-18
add show more modal for bound with children

### LINKED ISSUE #4077: See all bound-with children button and modal
See #3200 for context and background and #4074 for related information - mainly the UI on the record view.

This ticket allows the user to click a button to see all bound with child items for a parent item. The UI for each parent item on the instance record view shows one of the child records, but not all of them. See screenshots below for display:

If there's existing styles for modals and/or table views, we should be able to use those, including the X icon to close at the top right as well as the close button.

dnoneill:
closes #4077 

Screenshot 2025-04-22 at 6 16 53 PM
Screenshot 2025-04-22 at 6 16 33 PM

### DIFF
--- a/app/assets/stylesheets/modules/bound-withs.css
+++ b/app/assets/stylesheets/modules/bound-withs.css
@@ -1,0 +1,21 @@
+.bound-with {
+  .btn {
+    padding: 2px 6px;
+  }
+}
+
+.bound-with-modal {
+  table {
+    .bound-with-item {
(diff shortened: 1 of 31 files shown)


jcoyne (commented):
I'd like to see a system/feature spec with this that shows the modal can open.

dnoneill:
@jcoyne added a feature test for bound_with_children

PR #4777, opened 2025-04-24
Stop overriding small and use the small font size variable

jcoyne:


dnoneill:
@jcoyne it seems like the .small class is used for the icons on preview-opacity https://github.com/sul-dlss/SearchWorks/blob/main/app/javascript/controllers/preview_filmstrip_controller.js#L11-L12. Maybe instead of overriding the small default font we should change .small to .preview-opacity .bi

jcoyne:
@dnoneill `.small` is defined by bootstrap: https://github.com/twbs/bootstrap/blob/7bbbc492ef8bebf38ea965c3aba46b742ead1762/scss/_type.scss#L48-L50 So it uses the variable setting.

Issue #2090, opened 2018-10-30
Trigger login from linked title


ahafele:
"This title is not available for guests" wording

https://jirasul.stanford.edu/jira/browse/SW-1876

dbranchini:
"strongly correlated systems" was Sarah's search term in Articles+, but this happens in Articles+ and Bento. See screenshot. 

<img width="906" alt="Image" src="https://github.com/user-attachments/assets/7a8e87ee-d300-4190-83d0-c8c4fd217bab" />

dbranchini:
My initial ideas:
- If a user is NOT logged in, don't show these titles 
- Add a prominent alert at the top of the page to encourage the user to login for a better experience. Will provide design.

cbeer:
> If a user is NOT logged in, don't show these titles

Unless EDS has a way to suppress them when we're making the query (🤞) , we might end up with different numbers of results on each page.

dbranchini:
### Designs
[Figma with specifications](https://www.figma.com/design/fsVWCX66bpvFZGKEHNRBS5/SearchWorks-4.0?node-id=665-27850&t=Fdqyvr0NK3S5aVKQ-4)

### Prototype
[Figma in browser view](https://www.figma.com/proto/fsVWCX66bpvFZGKEHNRBS5/SearchWorks-4.0?page-id=664%3A27849&node-id=665-27850&viewport=-310%2C589%2C0.55&t=RXcr4aHorkUuQiuh-9&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=665%3A27850&show-proto-sidebar=1)

### Context
See above

### Design Notes
- Add the login alert at the top of the results (see screenshot)
- Change restricted titles item display (see screenshot)
- S icon (button or link) 
  - Focusable (can keyboard nav to and blue ring to indicate focus state)
  - On click: popover appears
  - aria-label: Stanford-only
  - Popover reads: Available to Stanford-affiliated users only. Log in to access.

### Screenshots

![Image](https://github.com/user-attachments/assets/a8b0ad09-5790-48f9-82d6-7919d464b0e8)
![Image](https://github.com/user-attachments/assets/dead1f85-e300-4b5a-8575-4848807575cb)

dnoneill:
@dbranchini should that prompt display on all the pages or just pages with results that are restricted?

dnoneill:
@dbranchini I mean the results on the page do not include the "title can not be displayed for guests" result. Do we still want the header?

PR #4783, opened 2025-04-25
convert index_brief.html.erb to ViewComponent

dnoneill:
This is to make things easier for a bug fix I found for https://github.com/sul-dlss/SearchWorks/issues/2090. This displays correctly on the default view but if you switch to the brief view we are displaying the old content. This partial is where the change needs to get made to fix this issue. I have split the fix into two PRs to make things easier to review.

### DIFF
--- a/app/components/articles/index_brief_component.html.erb
+++ b/app/components/articles/index_brief_component.html.erb
@@ -1,4 +1,3 @@
-<% doc_presenter = document_presenter(document) %>
 <% # header bar for doc items in index view %>
 <div class="brief-document container-fluid">
   <div class="row row-eq-height brief-container">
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

dnoneill:
@jcoyne do you mind doing a re-review? I found an issue because we use different partials for catalog vs articles.

jcoyne (approved):

cbeer:
I think this change has also broken the display of articles selections?

It used to look like this:
Screenshot 2025-04-28 at 15 28 59

and now it is missing the link and has a broken preview button (presumably coming from the catalog behavior): 
Screenshot 2025-04-28 at 15 29 08

PR #4788, opened 2025-04-28
Fix for icon facet alignment

jcoyne:
Screenshot 2025-04-28 at 2 17 21 PM

### DIFF
--- a/app/assets/stylesheets/modules/sul-icons.scss
+++ b/app/assets/stylesheets/modules/sul-icons.scss
@@ -1,14 +1,19 @@
-$sul-icon-prefix: 'sul-icon';
+$sul-icon-prefix: "sul-icon";
 //Sizing
-.#{$sul-icon-prefix}-1-5x { font-size: 1.5em; }
-.#{$sul-icon-prefix}-2x { font-size: 2em; }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/sul-icons.scss
```diff
@@ -18,31 +23,31 @@ $sul-icon-prefix: 'sul-icon';
...
-  align-items: center;
-  color: $black;
-  display: table-cell;
-  width: 30px;
-  &::before, svg {
-    height: 21px;
-    width: 21px;
-    vertical-align: top;
+.facet-values {
+  .blacklight-icons {
```
dnoneill, line 31: You need to use the .sul-icon selector of there is a problem with the `x` on selected facets.

Screenshot 2025-04-28 at 4 15 31 PM

dnoneill (changes requested):
if you use the

app/assets/stylesheets/modules/sul-icons.scss
jcoyne: Thank you. I'll look at that.

jcoyne: This is fixed  up.

PR #4790, opened 2025-04-28
Use bootstrap icon for rss rather than font-awesome

jcoyne:

### DIFF
--- a/app/assets/stylesheets/modules/sul-icons.scss
+++ b/app/assets/stylesheets/modules/sul-icons.scss
@@ -1,10 +1,6 @@
 .rss-icon {
-  // replace this extend when we have SUL Styles available to us
-  @extend .fa;
-  @extend .fa-lg;
-  @extend .fa-rss-square;
-
-  color: $sul-rss-icon-color;
+  font-size: 1.1rem;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/modules/sul-icons.scss
```diff
@@ -1,10 +1,6 @@
 .rss-icon {
-  // replace this extend when we have SUL Styles available to us
-  @extend .fa;
-  @extend .fa-lg;
-  @extend .fa-rss-square;
-
-  color: $sul-rss-icon-color;
+  font-size: 1.1rem;
```
dnoneill, line 2: Probably a bit nitpicky but right now the height on the rss icon is 16px. If you change this to 1rem it is exactly what we have right now.

jcoyne: It looks to me like it's overflowing the 16px box, which is why I made it bigger.
Screenshot 2025-05-12 at 5 21 05 PM

dnoneill: Isn't the 16 the width? and 14 the height?

PR #4796, opened 2025-04-28
Revert "convert index_brief.html.erb to ViewComponent"

cbeer:
This reverts commit f22c07d713bddfbfa2daa9773bec1af1a7f5db03.


cbeer:
There's a weird difference in behavior between articles + article selections, but I think #4783 + #4795 made it consistent with the index page.

PR #4825, opened 2025-04-30
Remove max-width from li

jcoyne:
This interferes with other styles and it's unclear what if any effect it was meant to have

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -49,7 +49,6 @@ dd {
 }
 
 p,
-li,
 dd {
   max-width: 45em;
 }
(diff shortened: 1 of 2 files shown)


cbeer:
Isn't that the max-width I just added in #4824?

cbeer:
I think that max-width is to stop things like tables of contents from getting uncomfortably long. I don't know if there are other things it targets, but completely removing it seems dangerous to me.

jcoyne:
@cbeer table of contents is already within a dd, so it can't exceed the 45ch max-width: See https://searchworks.stanford.edu/view/8595148

cbeer:
🤷

PR #4827, opened 2025-05-01
Update to bootstrap 5 [HOLD until after 5/5?]

cbeer:

### DIFF
--- a/app/assets/stylesheets/bootstrap-overrides.scss
+++ b/app/assets/stylesheets/bootstrap-overrides.scss
@@ -1,3 +1,7 @@
+body {
+  background-color: $sul-body-bg;
+}
+
 // I believe the text-success color override can be removed after upgrading to BS5.
 .text-success {
   color: $state-success-text !important;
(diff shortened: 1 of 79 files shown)


app/assets/stylesheets/bootstrap-overrides.scss
```diff
@@ -1,3 +1,7 @@
+body {
+  background-color: $sul-body-bg;
```
jcoyne, line 2: Doesn't `$body-bg` handle this? https://github.com/twbs/bootstrap/blob/ba5b5a630081fae0e793218ad49364ba46d15605/scss/_reboot.scss#L57 Or is this preventing dark-mode?

cbeer: `body-bg` also gets used for cards and all sorts of things, and I guess we want those backgrounds white but the true body bg to be that beige (for now.. I think that's changing in 4.0) 🤷‍♂️

cbeer:
- [ ] course reserves table striping is bad:
Screenshot 2025-05-01 at 13 03 20

jcoyne (approved):
Looks good, but needs a rebase.

PR #4837, opened 2025-05-01
[refactor] Extract mini-bento to components 

jcoyne:


app/components/search_result/mini_bento/layout_component.html.erb
```diff
@@ -9,15 +9,15 @@
...
         <div class='alternate-catalog-desc'>
-          Books, media, physical &amp; digital resources
+          <%= description %>
         </div>
       </div>
       <div class='col-md-6'>
-        <%= link_to(search_catalog_path(q: params.fetch(:q)), class: 'btn btn-primary') do %>
-          See <span class="alternate-catalog-count"%></span> catalog results
+        <%= link_to(url, class: 'btn btn-primary') do %>
+          See <span class="alternate-catalog-count"%></span> <%= result_singular %> results
```
cbeer, line 20:
```suggestion
          See <span class="alternate-catalog-count"></span> <%= result_singular %> results
```

config/locales/searchworks.en.yml
```diff
@@ -42,6 +41,15 @@ en:
         mla: MLA
         turabian: Turabian
         preferred: Preferred citation
+    mini_bento:
```
cbeer, line 44: This might just be a me-problem, but it hurts my brain that the ArticleComponent (and these i18n keys) are the thing that renders the catalog, and the catalog renders articles. Is there any hope to make that less confusing?

My only thought is trying to unify things into a single component that's more explicit about what is (or is not) rendering?

```
 render SearchResult::MiniBentoComponent.new(close: true, except: ['articles'])
```

🤷

jcoyne: I'm not sure there's a way to make this less confusing.  The mini bento on the articles page, shows catalog and the mini bento on the catalog page shows articles.  It's a confusing pattern.    I'm willing to flip the i18n_keys, but I think the `except: ['articles']` would be even weirder as it's not a set of things, it's a binary choice.

taylor-steve (approved):

PR #4846, opened 2025-05-02
Not every css property needs to be a variable

jcoyne:
Removing variables that are only used one time. This helps reduce indirection.


cbeer:
Unless this is fixing a problem we have, I'd be interested in seeing where the SW 4.0 designs go. I'm sure it won't work out, but updating the variables in one place seems a little nicer than searching through a huge pile of CSS 🤷‍♂️

PR #4850, opened 2025-05-02
Remove unnecessary partials

jcoyne:


cbeer:
Wasn't `search_header` a Blacklight thing? Would we be better off migrating to `SearchHeaderComponent` to stay somewhat synced with upstream?

jcoyne:
@cbeer yes, that parial was removed in BL9.  I've update this to use the components as blacklight does.

taylor-steve (approved):

PR #4851, opened 2025-05-02
style log in link

### LINKED ISSUE #4812: Styles on "Log in" prompt at the top of Articles+
### Summary
We might want to wait until SW4.0 for this, but if it's not too difficult, the background-color should be white `#FFFFFF`. The Log in text should be in semi-bold (600) in Digital Red `#B1040E` with a dotted underline in Black 60 `#767674` with the hover changing it to Cardinal Red - Dark `#820000` with a solid underline in the same color. See screenshots below.

### As Designed
Image

### As Implemented

Image

dnoneill:
closes #4812 

Screenshot 2025-05-02 at 1 42 27 PM
Screenshot 2025-05-02 at 1 42 23 PM

### DIFF
--- a/app/assets/stylesheets/variables.css
+++ b/app/assets/stylesheets/variables.css
@@ -1,3 +1,4 @@
 :root {
   --stanford-illuminating-dark: #fec51d1a;
+  --stanford-cardinal-dark: #820000;
 }
--- a/app/assets/stylesheets/component-library-overrides.css
+++ b/app/assets/stylesheets/component-library-overrides.css
@@ -69,3 +69,13 @@
     height: 164px;
   }
 }
+
+.stanford-digital-red-link {
+  font-weight: 600;
+  color: var(--stanford-digital-red);
+}
+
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/variables.css
```diff
@@ -1,3 +1,14 @@
 :root {
   --stanford-illuminating-dark: #fec51d1a;
+  --stanford-cardinal-dark: #820000;
 }
+
+.stanford-digital-red-link {
```
jcoyne, line 6: I don't think we want these styles in `variables.css`, do we? How about `component-library-overrides.css` instead?

dnoneill: Is it really a component override though? Maybe because it is overriding a link? I originally thought it was kind of a variable because if we use the stanford-digital-red-link elsewhere isn't it kind of a variable. Like how we do bg-blue in bootstrap. Honestly I had a lot of trouble figuring out where this went.

jcoyne (approved):

PR #4852, opened 2025-05-02
change border color on articles login alert

dnoneill:
The figma has a border of stanford red, we currently have a gray border. This fixes that.

**Figma**

Screenshot 2025-05-02 at 1 49 48 PM

**Current**

Screenshot 2025-05-02 at 1 49 52 PM

**This PR**

Screenshot 2025-05-02 at 1 49 56 PM

### DIFF
--- a/app/assets/stylesheets/modules/results-documents.scss
+++ b/app/assets/stylesheets/modules/results-documents.scss
@@ -46,8 +46,12 @@
   }
 }
 
-.eds-restricted {
-  border-left: 4px solid var(--stanford-cardinal)!important;
+.alert.eds-restricted {
+  border: 1px solid var(--stanford-cardinal);
+}
+
+#documents .document.eds-restricted, .alert.eds-restricted {
+  border-left: 4px solid var(--stanford-cardinal);
   .stanford-only {
     margin-left: 0;
   }
--- a/app/components/search_result/login_banner_component.rb
+++ b/app/components/search_result/login_banner_component.rb
@@ -5,7 +5,7 @@ module SearchResult
     delegate :new_user_session_path, to: :helpers
 
     def classes
-      'alert px-3 py-3 border-end border-top border-bottom alert-dismissible rounded-0 fade show eds-restricted'
+      'alert px-3 py-3 alert-dismissible rounded-0 fade show eds-restricted'
     end
   end
 end


app/assets/stylesheets/modules/results-documents.scss
```diff
@@ -46,8 +46,12 @@
   }
 }
 
+.alert.eds-restricted {
+  border: 1px solid var(--stanford-cardinal);
+}
+
 .eds-restricted {
-  border-left: 4px solid var(--stanford-cardinal)!important;
+  border-left: 4px solid var(--stanford-cardinal) !important;
```
jcoyne, line 54: Do we need `!important` here?

dnoneill: Yes. It is needed to override the border in .alert.eds-restricted  and also the border coming from blacklight.

jcoyne: Can we just put `4px solid var(--stanford-cardinal)` in the `.alert.eds-restricted`?

dnoneill: Even if we did, it would break the search view, so the important has to stay.

dnoneill: Screenshot 2025-05-05 at 3 24 00 PM
Screenshot 2025-05-05 at 3 23 44 PM

jcoyne: Would this be the case even if there was a `.document.eds-restricted` selector?  It seems like having these as two separate styles makes sense, because one is an alert and the other is a document.

dnoneill: I am not quite sure why we would want to seperate the same style into two different selectors. If we did .document.eds-restricted I would assume we would do a comma which means we would still have to do important. i.e. 
.document.eds-restricted, .alert.eds-restricted {
border-left: ....
}

I am not quite sure what you are suggesting? The only thing this pr does is add a space. The important is already there. Shouldn't this be a separate issue?

jcoyne: > "seperate the same style into two different selectors"

They aren't really the same component. They just happen to look similar right now.  I can see a future where we make the alert look like something in the component library but keep the `.document` looking like it does in this PR.   In that case it's more difficult to understand that we have this one rule that affects two very different components.

dnoneill: But isn't that css? The thing they share is the same style. I don't have a problem being more specific, i.e. .document.eds-restricted, .alert.eds-restricted {} but that doesn't get rid of the important. Or do we not care anymore?

jcoyne: Why wouldn't .document.eds-restricted get rid of the important?  Isn't that a higher precedence selector?

dnoneill: It does not, #documents .document.eds-restricted however does. Fixed.

jcoyne (approved):

PR #4854, opened 2025-05-02
Add duplicates removed text in articles search results header

### LINKED ISSUE #1713: Add "duplicates removed" text to articles results header
Add right-aligned text to the articles+ results count header row:

> Exact duplicates have been removed

screen shot 2017-09-07 at 12 56 11 pm

(To address a common user question/confusion.)

jcoyne:
Fixes #1713
Screenshot 2025-05-02 at 12 59 47 PM

### DIFF
--- a/app/views/articles/_search_results.html.erb
+++ b/app/views/articles/_search_results.html.erb
@@ -17,7 +17,13 @@
         <div class="search_num_of_results">
           <div class='results-heading'>
             <h1 class="visually-hidden"><%= t('blacklight.search.page_heading', search_type: search_type_name) %></h1>
-            <h2><%= pluralize(number_with_delimiter(@response.response[:numFound]), "#{search_type_name} result") %></h2>
+            <h2>
+              About <%= pluralize(number_with_delimiter(@response.response[:numFound]), "#{search_type_name} result") %>
+              <span class="duplicates">Duplicates removed</span>
+              <button class="btn duplicate-info" aria-label="More information" data-bs-container="body" data-bs-toggle="popover" data-bs-placement="right" data-bs-content="<%= t('.duplicates') %>">
+                <span class="bi-info-circle"></span>
+              </button>
+            </h2>
           </div>
         </div>
         <%= render 'search_header' %>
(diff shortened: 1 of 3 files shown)


dbranchini:
Screenshot looks good to me.

dnoneill:
@jcoyne can you have the conversion of app/assets/stylesheets/modules/article.scss to css into a different PR? I am having trouble viewing the differences.

app/views/articles/_search_results.html.erb
```diff
@@ -17,7 +17,13 @@
         <div class="search_num_of_results">
           <div class='results-heading'>
             <h1 class="visually-hidden"><%= t('blacklight.search.page_heading', search_type: search_type_name) %></h1>
-            <h2><%= pluralize(number_with_delimiter(@response.response[:numFound]), "#{search_type_name} result") %></h2>
+            <h2>
+              About <%= pluralize(number_with_delimiter(@response.response[:numFound]), "#{search_type_name} result") %>
+              <span class="duplicates">Duplicates removed</span>
+              <button class="btn duplicate-info" aria-label="explanation" data-bs-container="body" data-bs-toggle="popover" data-bs-placement="right" data-bs-content="<%= t('.duplicates') %>">
```
dnoneill, line 23: I am not sure explanation is a good aria-label

jcoyne: What is your suggestion?

dnoneill: That I am not sure. Maybe something like `results explanation`?

jcoyne: It's more of an explanation of "Duplicates removed", not of the results.

dnoneill: Isn't sort of both? the explanation has to do with the results `This may cause the number of displayed results to differ from the total found.` But it could also be Duplicates removed info? Honestly I don't really have a good handle on what a good aria label is but explanation seems a little too 🤷. The icon is an info icon so at the very least I think it should be `info`

dbranchini:
I'd suggest, "More information" or "Show details."

taylor-steve (approved):

Issue #4791, opened 2025-04-28
Prepare to rename SAL1+2, SAL3, and SAL-NEWARK


cbeer:
The `building_facet` values currently index the display label for the FOLIO library (*mostly...). It's possible/likely the SAL library names will change to provide greater clarity for patrons, and it'd be really good if:
- we didn't end up with multiple labels for the same place e.g. "SAL3 (off-campus storage)" and "Off-campus storage (SAL3)"
- existing links for e.g. SAL3 (off-campus storage) continued to work in the short to medium term (perhaps redirecting to the corrected value?)

PR #4858, opened 2025-05-02
Use library_code_facet_ssim for the library

jcoyne:
This allows us to change the label with config, rather than have to reindex everything.  Before we activate this facet in the new UI, we need to populate it in the indexer.

Ref #4791

See https://github.com/sul-dlss/searchworks_traject_indexer/pull/1638

### DIFF
--- a/app/components/searchworks4/library_facet_component.rb
+++ b/app/components/searchworks4/library_facet_component.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Searchworks4
+  class LibraryFacetComponent < Blacklight::Facets::ListComponent
+    def facet_item_presenters
+      super.sort_by(&:label).reject do |presenter|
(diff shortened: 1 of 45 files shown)


cbeer:
I think we're ready to review this in light of #5210. Can you rebase this PR?

Two comments to start:
- I wonder if there's a way to keep existing urls with `building_facet` working (in case it's been bookmarked or linked to externally), perhaps by translating it to the `library_code_facet_ssim` behind the scenes?

- #5210 calls for having multiple library codes show the same label. Do you think this approach is compatible with that easily (say, but  non-library code like SDR and having things in SAL1-2/SAL-NEWARK/SAL3 use that for indexing?), or is there something better?

jcoyne:
@cbeer I rebased and added the redirect from the old facet values. I think the "Off-campus collections" first requires us to index all the SALs together as one code.

taylor-steve (approved):

PR #4861, opened 2025-05-02
update eds restricted item view to show message and popover

### LINKED ISSUE #1452: Behavior for guest-restricted content
Currently, in search results the link goes to the record view which shows all available metadata, which may include ISSN or a DOI (it varies by source).

### Example of current behavior

screen shot 2017-07-06 at 3 28 25 pm

### New behavior

For search results and record view:
- Display the title as a link
- Clicking the link requires the user to login
- Do not show any metadata for guest users
- The title is always `"This title is unavailable for guests, please login to see more information."`

dnoneill:
closes #1452. Would use `hide_whitespace` because app/views/articles/_show_main_content.html.erb is a small change but looks large due to tabbing fixes.
Screenshot 2025-05-02 at 3 34 25 PM

### DIFF
--- a/app/components/articles/document_component.html.erb
+++ b/app/components/articles/document_component.html.erb
@@ -9,8 +9,13 @@
   class: classes.flatten.join(' ') do %>
   <%= header %>
   <div class="document-main-section">
-    <%= title %>
-    <%= document_metadata %>
+    <% if @document.eds_restricted? %>
+      <h1><%= I18n.t('searchworks.guest_message') %>
+      <%= render StanfordOnlyPopoverComponent.new %></h1>
+    <% else %>
+      <%= title %>
+      <%= document_metadata %>
+    <% end %>
   </div>
   <%= footer %>
 <% end %>
--- a/app/views/articles/_show_main_content.html.erb
+++ b/app/views/articles/_show_main_content.html.erb
@@ -1,12 +1,18 @@
-    <%= render RecordToolbarComponent.new(presenter: document_presenter(@document), search_context: @search_context, search_session: search_session) %>
-    <% @page_title = t 'blacklight.search.show.title_html', :document_title => document_presenter(@document).html_title, :application_name => "#{I18n.t('blacklight.application_name')} catalog" -%>
-    <%# content_for(:head) { render_link_rel_alternates } -%>
- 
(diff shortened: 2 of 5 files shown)


app/components/articles/document_component.html.erb
```diff
@@ -9,8 +9,13 @@
   class: classes.flatten.join(' ') do %>
   <%= header %>
   <div class="document-main-section">
-    <%= title %>
-    <%= document_metadata %>
+    <% if @document.eds_restricted? %>
```
jcoyne, line 12: Shouldn't there be a test about who the user is too?

app/views/articles/_show_main_content.html.erb
```diff
@@ -1,12 +1,16 @@
...
-    <%# content_for(:head) { render_link_rel_alternates } -%>
- 
-    <%= render (blacklight_config.view_config(:show).document_component || Blacklight::DocumentComponent).new(document: document_presenter(@document), component: :div, title_component: :h1, show: true) do |component| %>
-      <% component.with_footer do %>
-        <% if @document.respond_to?(:export_as_openurl_ctx_kev) %>
-          <!-- COinS, for Zotero among others. -->
-          <span class="Z3988" title="<%= @document.export_as_openurl_ctx_kev(@document[@document.format_key]) %>"></span>
-        <% end %>
-      <%  end %>
+<div class="mt-4">
```
jcoyne, line 1: Should this div only show conditionally?

app/components/articles/document_component.html.erb
jcoyne: Ideally using https://github.com/sul-dlss/SearchWorks/pull/4744 😉

dnoneill: I would assume no? The eds_restricted tests the title and if the title is the eds restricted pattern (eds title is restricted for guests) https://github.com/sul-dlss/SearchWorks/blob/54f901e3734ff2b43ead7648b2844b5cc6af3ce5/app/models/eds_document.rb#L85. I would assume that if the user is logged in, eds would not return that title?

app/views/articles/_show_main_content.html.erb
dnoneill: Oops. yes. Thanks

app/components/articles/document_component.html.erb
jcoyne: Thanks. I missed that.

PR #4868, opened 2025-05-05
Simplify accordion-section css

jcoyne:
Some of these classes are no longer applied in the markup.  Switch to using Bootstrap 5 idioms for clarity


taylor-steve:
Works well but looks like we may have lost some space after the button.

Before:
Screenshot 2025-05-05 at 5 05 42 PM

After:
Screenshot 2025-05-05 at 5 05 50 PM

jcoyne:
Oh. Interesting, chromium browsers show the padding but Safari and Firefox do not.

jcoyne:
@taylor-steve we believe the intent is to have zero space after the button.  There was a preexisting discrepancy on Chromium vs Firefox and Safari.  If that space is no longer present, we should consider it a fix for consistency.

taylor-steve (approved):

PR #4872, opened 2025-05-05
Remove duplicate rule

jcoyne:
border-bottom is included in border, which is defined just above this removed line


cbeer:
It looks like the other borders used to be a slightly different color until #957 and the bottom border variable was garbage collected in [#3898](https://github.com/sul-dlss/SearchWorks/pull/3898).

PR #4886, opened 2025-05-06
Don't use aria-expanded on non-control elements

### LINKED ISSUE #4877: aria-expanded on a div is not legal
See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded

Found in callnumber browse:
```
<div id="callnumber-1" class="embed-callnumber-browse-container" style="overflow: hidden; height: 64.5127px; padding-top: 2.44257px; margin-top: 2.15521px; padding-bottom: 0px; margin-bottom: 0px;" aria-expanded="true">
```

Maybe also in course reserves?
```
app/views/catalog/_accordion_section_course_reserves.html.erb:    <div class="details <%= id %>-course-reserves" aria-expanded="false">
```

jcoyne:
Also converted the accordion section javascript to stimulus

Fixes #4877


taylor-steve:
Will this expand what is being tracked by GA? https://github.com/sul-dlss/SearchWorks/blob/eb74361eeca5bb9dc4eac54bc52d21abfd32c25a/app/javascript/analytics.js#L50

jcoyne:
@taylor-steve Good catch. I've updated it.

taylor-steve:
Looks good but could you speak to the full removal of 'Accordion selection'? I'm not sure I'm following.

jcoyne:
@taylor-steve As far as I can tell there is a duplicate section in analytics.js for `[data-accordion-section-target]`

PR #4891, opened 2025-05-07
Restore facet collapse functionality

jcoyne:


app/components/searchworks/response/facet_group_component.html.erb
```diff
@@ -1,23 +1,23 @@
-<% # This is a copy of the Blacklight 8 version of this component. 
+<% # This is a copy of the Blacklight 8 version of this component.
```
jcoyne, line 1: @cbeer should we remove this now that it's a customized version?

PR #4892, opened 2025-05-07
Move conditional for mini-bento rendering to the component

jcoyne:


app/components/search_result/mini_bento/article_component.rb
```diff
@@ -10,6 +10,10 @@ def initialize(close:)
 
       attr_reader :close
 
+      def render?
+        params[:q].present?
+      end
+
```
cbeer, line 16:
```suggestion
```

I think you took away the `:gallery` condition because it isn't used for articles, but I also think articles+ requires a params[:q] (if I'm reading https://github.com/sul-dlss/SearchWorks/blob/main/app/controllers/articles_controller.rb#L23 right)

jcoyne: Even though it requires the params[:q], it still draws the page without it.  When that happens, we still want to prevent mini-bento from drawing.

jcoyne: It's for the no-results page like: https://searchworks.stanford.edu/articles?f%5Beds_content_provider_facet%5D%5B%5D=Journal+provider&f%5Beds_search_limiters_facet%5D%5B%5D=Stanford+has+it

cbeer: Huh. I don't know why we let people remove the query string (because it'll always break things) but ok 🤷‍♂️

PR #4893, opened 2025-05-07
Move imprint from a helper to a model attribute

jcoyne:

### DIFF
--- a/app/views/catalog/_marc_search_results_document_fields.html.erb
+++ b/app/views/catalog/_marc_search_results_document_fields.html.erb
@@ -20,7 +20,7 @@
     <%= render 'marc_fields/linked_author_index', linked_author: document.linked_author(target) %>
   <% end %>
 
-  <% if (imprint = results_imprint_string(document)).present? %>
-    <li><%= imprint %></li>
+  <% if document.imprint_string %>
+    <li><%= document.imprint_string %></li>
   <% end %>
 </ul>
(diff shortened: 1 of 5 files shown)


app/views/catalog/_marc_search_results_document_fields.html.erb
```diff
@@ -20,7 +20,7 @@
     <%= render 'marc_fields/linked_author_index', linked_author: document.linked_author(target) %>
   <% end %>
 
-  <% if (imprint = results_imprint_string(document)).present? %>
-    <li><%= imprint %></li>
+  <% if document.imprint_string %>
```
cbeer, line 23:
```suggestion
  <% if document.imprint_string.present? %>
```

... or does something about attribute handle that magically? I assume we're guarding against empty strings and the like 🤷‍♂️

jcoyne: Empties are discarded https://github.com/sul-dlss/stanford-mods/blob/eb4392be6817078737daaf3cb0b6362145cd921f/lib/stanford-mods/concerns/origin_info.rb#L72 and https://github.com/sul-dlss/searchworks_traject_indexer/blob/df5aa3ea292e927f5ea15531698350e564feb02d/lib/traject/config/folio_config.rb#L810

cbeer: 🤞 , I guess?

PR #4897, opened 2025-05-07
Remove unnecessary selector

jcoyne:


dnoneill (changes requested):
This is a necessary selector. Without it the alert login looks like this:
Screenshot 2025-05-12 at 5 52 46 PM
When it should look like this:
Screenshot 2025-05-12 at 5 53 35 PM

PR #4908, opened 2025-05-09
don't display filter bullet when no filters

### LINKED ISSUE #4906: Don't display "Clear all" when there are no filters.
After Modify your search > Remove filters, we show the user's filters and a link "Clear all" to remove all filters. If there aren't any filters, don't display this link.  See screenshot.

Image

dnoneill:
closes #4906 

Before:

After:
Screenshot 2025-05-09 at 11 51 08 AM

### DIFF
--- a/app/components/search_result/no_result/filters_component.rb
+++ b/app/components/search_result/no_result/filters_component.rb
@@ -3,7 +3,9 @@
 module SearchResult
   module NoResult
     class FiltersComponent < Blacklight::ConstraintsComponent
-      def query_constraints; end
+      def render?
+        !(@search_state.filters.blank? && @search_state.clause_params.blank?)
+      end
     end
   end
 end
--- a/app/views/shared/_zero_results.html.erb
+++ b/app/views/shared/_zero_results.html.erb
@@ -4,9 +4,12 @@
     <div class="border-before">
       <h2 class="ps-3"><%= t 'blacklight.search.zero_results.modify_your_search', search_type: search_type_name %></h2>
       <ul>
-        <li>Remove filters <%= link_to 'Clear all', search_action_path(q: params[:q]), class: "text-primary ms-2" %>
-          <%= render SearchResult::NoResult::FiltersComponent.new(search_state:) %>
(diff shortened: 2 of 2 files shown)


app/components/search_result/no_result/filters_component.rb
```diff
@@ -3,7 +3,9 @@
 module SearchResult
   module NoResult
     class FiltersComponent < Blacklight::ConstraintsComponent
-      def query_constraints; end
```
cbeer, line 6: Do we want to display the original query now? I don't see that in the ticket, but maybe there was a conversation somewhere about it.

app/components/search_result/no_result/filters_component.rb
```diff
@@ -3,7 +3,9 @@
...
+      def render?
+        !(@search_state.filters.blank? && @search_state.clause_params.blank?)
```
cbeer, line 7:
```suggestion
        @search_state.filters.present? || @search_state.clause_params.present?
```

Maybe that's clearer, or at least takes less reasoning?

app/components/search_result/no_result/filters_component.rb
cbeer: ... or was this to include adv search parameters, maybe? I don't know if that needs to be something upstream or if we should just own more of the logic here..

dnoneill: my understanding is this was on purpose https://github.com/sul-dlss/SearchWorks/pull/4696#issuecomment-2808549092

app/components/search_result/no_result/filters_component.rb
taylor-steve: FWIW I have a slight preference for inverting the logic. I wouldn't immediately recognize that as Blacklight code.

app/components/search_result/no_result/filters_component.rb
dnoneill: Okay I misunderstood what this comment was saying. This is fixed.

PR #4910, opened 2025-05-09
add timeout to folio client requests

### LINKED ISSUE #4876: Set a reasonable timeout for FOLIO mod-rtac requests for availability
FOLIO doesn't have a built-in timeout, but we can/should abort mod-rtac requests after some unreasonable length of time (15s? 30s?) to save FOLIO from itself.

dnoneill:
closes #4876

### DIFF
--- a/app/services/folio_client.rb
+++ b/app/services/folio_client.rb
@@ -54,7 +54,7 @@ class FolioClient # rubocop:disable Metrics/ClassLength
 
   # https://s3.amazonaws.com/foliodocs/api/mod-courses/r/courses.html#coursereserves_courses_get
   def courses
-    get_json('/coursereserves/courses', params: { limit: 2_147_483_647 }).fetch('courses', []).sort_by { |x| x['id'] }.map do |course|
+    get_json('/coursereserves/courses', params: { limit: 2_147_483_647 }, timeout: 60).fetch('courses', []).sort_by { |x| x['id'] }.map do |course|
       instructors = course.dig('courseListingObject', 'instructorObjects').map { |x| x.slice('name') }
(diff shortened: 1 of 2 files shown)


taylor-steve:
It could be nice to add those timeout value(s) in the Folio settings since we're using them already for this client?:
https://github.com/sul-dlss/SearchWorks/blob/88fa18cf7157ff2408280bf05f6f72d5b3ee745f/config/settings.yml#L75

cbeer:
I think 30s is enough for most things, but I worry about the course reserves API call which usually takes a little while to run (20s+?). Do we want to bump the global timeout, or is there a way to do it by request type?

app/services/folio_client.rb
```diff
@@ -155,8 +155,11 @@ def authenticated_request(path, headers: {}, **other)
   end
 
   def request(path, headers: {}, method: :get, **other)
+    timeout = other[:timeout] || Settings.folio.timeout
+    other.delete(:timeout)
```
cbeer, line 159:
```suggestion
  def request(path, headers: {}, method: :get, timeout: Settings.folio.timeout, **other)
```

Is there a reason we want to pass timeout to the `HTTP#request` method?

cbeer: Sorry, I'm not following. Isn't the timeout kwarg already flowing through those methods?

cbeer: Maybe I'm missing something, but I still don't think it needs to move? Here's what I'm seeing:

```
def authenticated_request(**other)
  request(**other)
end

def request(timeout: -1)
  puts timeout
end
```

```
authenticated_request(timeout: 5)
# => prints 5

authenticated_request
# => prints -1
```

Is there something else going on here?

PR #4919, opened 2025-05-12
switch to using blacklight skip links

### LINKED ISSUE #4871: Update skip links to use upstream components
Blacklight provides some reasonable default skip links and a way to augment them in our views. We should convert our custom skip link use to use the upstream components.

dnoneill:
closes #4871


cbeer:
Are these working for you? I think I'm able to tab through to them, but they aren't visible when selected?

PR #4935, opened 2025-05-15
Correctly pass data-long-table-button-context-value attribute

jcoyne:


taylor-steve:
Does https://github.com/sul-dlss/SearchWorks/blob/eb495f844400f8f3fb645632c1108652194aa838/app/components/access_panels/library_location_component.html.erb#L17 need to change as well?

jcoyne:
@taylor-steve thanks. Got it.

taylor-steve (approved):

PR #4946, opened 2025-05-16
Restore scroll-behavior

### LINKED ISSUE #4941: Odd page focus (?) when user moves from search results page to record view
When a user clicks from the search results into a record view, they aren't always taken to the top of the record view page. 

To see this, search for something, scroll down the search results to about the 4th or 5th result and click into a record view. Notice that the page initially loads somewhere lower than the top of the page, and then moves up. Sometimes the focus is down by the call number browse, sometimes in the middle of the page, and sometimes at the top. The focus seems to be related to how far down the search results you were when you clicked into the record, but that's just a quick theory. (Based on clicking on the first result and being taken to the top of the record view page. Then back to the results and scrolling down to around the 10th result, clicking on that title and noticing that the page load seems to be down around the call number browse before moving up.) 

https://github.com/user-attachments/assets/f9903999-16c5-4f28-876c-1f3e74ce2486

I would have expected that anytime a user selects a search result, they would be taken to the top of the record view page.

taylor-steve:
Closes #4941.

Bootstrap 5 changed the default:
```css
@media (prefers-reduced-motion: no-preference) {
  :root {
    scroll-behavior: smooth;
  }
}
```

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -5,6 +5,10 @@
   --link-hover-decoration-line: underline;
   --link-hover-decoration-style: solid;
   --link-text-decoration-color: gray;
+
+  /* Bootstrap 5 sets scroll-behavior to `smooth` when `prefers-reduced-motion` is `no-preference`
+    Some found this undesirable for turbo page transitions. See https://github.com/sul-dlss/SearchWorks/issues/4941 */
+  scroll-behavior: auto;
 }
 
 a {


app/assets/stylesheets/base.scss
```diff
@@ -5,6 +5,7 @@
   --link-hover-decoration-line: underline;
   --link-hover-decoration-style: solid;
   --link-text-decoration-color: gray;
+  scroll-behavior: auto;
```
jcoyne, line 8: Can you add a comment here, or improve the commit message explaining the issues this is solving?

PR #4953, opened 2025-05-19
Combine landmark assertions into a single test

jcoyne:
Feature specs are slow, so multiple assertions avoids extra overhead

### DIFF
--- a/spec/features/blacklight_customizations/accessibility_spec.rb
+++ b/spec/features/blacklight_customizations/accessibility_spec.rb
@@ -2,42 +2,17 @@
 
 require 'rails_helper'
 
-RSpec.feature "Aria Landmarks", :js do
+RSpec.feature "Aria Landmarks" do
   before do
     visit root_path
     fill_in "q", with: ''
     click_button 'search'
   end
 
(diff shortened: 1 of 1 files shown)


spec/features/blacklight_customizations/accessibility_spec.rb
```diff
@@ -2,42 +2,17 @@
...
-RSpec.feature "Aria Landmarks", :js do
+RSpec.feature "Aria Landmarks" do
   before do
     visit root_path
     fill_in "q", with: ''
     click_button 'search'
   end
 
-  scenario "should have header landmark" do
+  it "has landmarks" do
```
cbeer, line 12: Maybe `aggregate_failures` would be helpful here? 

https://rspec.info/features/3-13/rspec-expectations/aggregating-failures/

jcoyne: I think that's just more complexity than necessary. If any one of these assertions fails, it's not drawing the appropriate landmarks.   Maybe I'm not understanding why we'd want to run the rest of the assertions even if one fails.

PR #4702, opened 2025-04-10
Create new citation service

jcoyne:
```
search-works(dev)> doc = SolrDocument.find('2472159')
Solr query: get get {qt: nil, ids: "2472159"}
Solr fetch (43.7ms)
=>
#<SolrDocument:20040 _source: {"id" => "2472159", "hashed_id_ssi" => "ebb4709083b1412420969973c0d82e4f", "marc_json_struct" => ["{\"leader\":\"51206cem a2203913 a 450...
search-works(dev)> CitationService.new(doc).export('apa')
=> ["Great Britain, Great Britain Ordnance Survey, &amp; Great Britain Army Royal Engineers. (1958). <i>World 1:500,000</i> [Map]."]
```

PR #4962, opened 2025-05-20
Guard against missing 008 or 245 fields.

cbeer:


jcoyne (changes requested):
Would you please add a test for these cases?

cbeer:
This is fixing a bunch of tests on #4702?

Issue #1019, opened 2015-04-08
Remove shelving range from the StackMap display.


jkeck:
SW-1470

jvine:
https://jirasul.stanford.edu/jira/browse/SW-1396

jvine:
also https://jirasul.stanford.edu/jira/browse/SW-1530

jvine:
Call number PS1303 .R37 2001 is located in Green Library, 3rd floor of the East Wing.

PR #4966, opened 2025-05-20
Remove unused variables

jcoyne:


cbeer:
It looks like we dropped shelving range in https://github.com/sul-dlss/SearchWorks/issues/1019

PR #4973, opened 2025-05-20
add eds_pub_date facet

### LINKED ISSUE #1832: Facet based on recent publication date
Faculty would like to be able to refine the search so that only the papers published in the last week would appear.

https://jirasul.stanford.edu/jira/browse/SW-1914

dnoneill:
closes #1832 
needs https://github.com/ebsco/edsapi-ruby/pull/131 to work

Screenshot 2025-05-20 at 3 23 23 PM
Screenshot 2025-05-20 at 2 53 51 PM


dbranchini:
Screenshots look good. Thank you!

cbeer:
@dbranchini : are radio buttons the right affordance for this, or would it be better to treat them as regular facet links? With the radio buttons, it seems a little harder to escape from that context and get back to using the date slider (and.. are optional radio buttons even ok?)

dbranchini:
I'm not sure what you mean by "optional radio buttons"? (I assume they work as a group and you can only select one.) I noticed the radio buttons when Niqui implemented it. I think they're appropriate, but probably better if all of our facets were either radio buttons or checkboxes or none at all. Since I didn't give her much guidance to begin with, initially I thought I'd let it go. BUT this is probably the right time to fix it. Let's go with a standard facet link since that's what we use everywhere else.

PR #4984, opened 2025-05-21
Convert preview scripts to use Stimulus

jcoyne:
This allows it to get automatically initialized when these elements show on the page, rather than having to manually initialize them after they are dynamically added to the page.


app/components/search_result/document_gallery_component.rb
```diff
@@ -9,5 +9,27 @@ def preview_container_dom_class
     def classes
       super - ['document'] + ['gallery-document']
     end
+
+    def data
+      {
+        doc_id: @document.id,
+        document_counter: @counter,
+        preview_in_gallery: 'true'
```
cbeer, line 17:
```suggestion
        document_counter: @counter
```

I think you determined this property is ignored in #4986?

jcoyne: Yes. This falls out after I rebased.

PR #4986, opened 2025-05-21
Remove unused data property

jcoyne:


cbeer:
I don't see any sign this was used in the initial implementation (but there was a little discussion on the PR about calculating widths.. maybe `data-preview-in-gallery` was an abandoned approach?):
https://github.com/sul-dlss/SearchWorks/commit/9245b41223e4d755197f48d44c2394ae7ad694af

PR #4988, opened 2025-05-21
[Refactor] mini bento to use bento to get counts

### LINKED ISSUE #4975: mini bento guides badge should be 100+
Image

dnoneill:
closes #4975


app/components/search_result/mini_bento/layout_component.html.erb
```diff
@@ -34,30 +34,30 @@
       </div>
 
       <% if false %>
-      <div data-controller="blacklight-result-count" data-blacklight-result-count-url-value="<%= archive_search_url %>">
-        <%= link_to archive_search_url do %>
+      <div data-controller="blacklight-result-count" data-blacklight-result-count-url-value="<%= bento_search_url('archives') %>">
```
jcoyne, line 37: Doesn't this just proxy a request through bento to archives?  That means it's putting load on an extra server and introducing extra places to fail.  I'm not sure why we'd want to do this.

jcoyne (changes requested):

app/components/search_result/mini_bento/layout_component.html.erb
dnoneill: I guess we could just do this only for libguides which was already doing what you are talking about. I was figuring it would be better to have a common pattern.

dnoneill: @jcoyne here is a version that just updates the code to use bento for libguides. https://github.com/sul-dlss/SearchWorks/pull/5000/files

PR #4997, opened 2025-05-21
Update to Justin's edsapi fork so that we can use latest csl gems

jcoyne:


jcoyne:
The upstream 1.1.7 was released which fixed this

PR #5008, opened 2025-05-22
Exclude screen-reader style from font-awesome

jcoyne:


cbeer:
This fixes the skip links visibility problem (#4871 )

PR #5010, opened 2025-05-22
Use turbo frame for preview content

jcoyne:

### DIFF
--- a/app/javascript/controllers/preview_filmstrip_controller.js
+++ b/app/javascript/controllers/preview_filmstrip_controller.js
@@ -1,10 +1,10 @@
 import { Controller } from "@hotwired/stimulus"
-import PreviewContent from '../preview-content'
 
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/preview_filmstrip_controller.js
```diff
@@ -36,8 +35,8 @@ export default class extends Controller {
     this.preview.addClass('preview').empty();
 
     this.appendPointer()
-
-    PreviewContent.append(this.urlValue, divContent);
+    const id = this.urlValue.split('/').pop()
```
cbeer, line 38: I think we probably hate this, right? Is it possible to either pass the id in somewhere or not have to dynamically generate the HTML?

jcoyne: Yes. I've passed it in.

PR #5024, opened 2025-05-23
Test Chrome settings for flappy tests

taylor-steve:
https://issues.chromium.org/issues/402796660

It's a slightly different situation but maybe it's related


taylor-steve:
Not it!

Issue #5035, opened 2025-05-28
Exhibits panel doesn't look quite right


saseestone:
<img width="321" alt="Image" src="https://github.com/user-attachments/assets/e3eee7c2-879c-4111-b037-4fbc448e9ba7" />

Issues I see:

- I don't think we displayed these in a long list before.  They used to be side-by-side.
- We're displaying all 5 and providing a link to all 5. 

Example record: https://searchworks.stanford.edu/view/10357851

cbeer:
Same thing with the media objects on the home page:
<img width="638" alt="Image" src="https://github.com/user-attachments/assets/09c800a9-0328-4a6d-bea2-8bec2fc62629" />

PR #5044, opened 2025-05-30
Fix homepage layout

jcoyne:
Ref #5305

After:
Screenshot 2025-05-30 at 10 32 19 AM

Before:
Screenshot 2025-05-30 at 10 17 37 AM


taylor-steve:
Ref https://github.com/sul-dlss/SearchWorks/issues/5035

taylor-steve (approved):

PR #5053, opened 2025-06-03
Add JavaScript linting with ESLint and GitHub Actions workflow

jcoyne:


jcoyne:
I'd prefer we keep it in a separate GH workflow.  Currently the ruby build does `ci`.  You can run this locally with `yarn lint`

PR #5064, opened 2025-06-04
Remove the extra halo around the Stanford-only button

jcoyne:
before
Screenshot 2025-06-04 at 1 18 53 PM

after
Screenshot 2025-06-04 at 1 18 37 PM


jcoyne:
@cbeer I added `--bs-btn-focus-box-shadow` to restore the focus ring.

PR #5071, opened 2025-06-06
Remove brief and gallery views

jcoyne:
These are no longer used in the searchworks 4 designs


cbeer:
That's _brief_ and gallery view, right?

jcoyne:
@cbeer Yes. Thanks for the catch. I've updated the commit message.

cbeer:
(Also noting that brief view might be staying after all)

Issue #5055, opened 2025-06-03
SW home


dbranchini:
### Figma designs with specifications

- [Catalog XXL](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=5-3591&t=y1epxKf3V98cseBW-4) 
- [Articles XXL](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=5-3670&t=y1epxKf3V98cseBW-4)
- [Catalog responsive designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=185-43448&t=y1epxKf3V98cseBW-4)
- [Articles responsive designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=185-47752&t=y1epxKf3V98cseBW-4)

Design notes:

- The specific wording/language is not finalized. 

In browser mode:

- [Catalog XXL](https://www.figma.com/proto/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?page-id=2%3A3883&node-id=5-3591&p=f&viewport=1040%2C1052%2C0.45&t=7WPSSwZgSghVYSr3-1&scaling=scale-down-width&content-scaling=fixed&starting-point-node-id=5%3A3591&show-proto-sidebar=1)
- Responsive designs for both Catalog and Articles+ are available from sidebar

taylor-steve:
Now that we've got a first pass merge we need a second pass to refine. Here are some examples (not an exhaustive list). We may want to pair/swarm to dial these in details.

* Spacing/sizing
  * Header links
  * Bottom of header
  * "What are you searching for" column seems less cramped in the design
* Search Box
  * Drop shadow
  * Rounded corners
  * Place holder should be italicized
* Typography
  * The search tips are bolder than the design
* Responsive Design. 
  * We might need Darcy's help for some of this but I think another pass would be good first. E.g., at ~1400px it drops the left margin:

<img width="221" alt="Image" src="https://github.com/user-attachments/assets/a3491708-2d58-4a9a-91eb-21ccc1e92088" />
  
   * This line on the left isn't present in design at smaller sizes:

[...]

PR #5098, opened 2025-06-09
Adjust search tips section fonts

marlo-longley:
Closes ##5078

In looking at this, I also noticed that the "Featured resources" links differ from Figma -- they are blue because they are links, but the Figma has them dark gray/the same as what I am pushing up in this PR for "Search tips".

If desired I can also change that.

This PR:
Screenshot 2025-06-09 at 3 02 37 PM


taylor-steve:
Regarding the question, Darcy confirmed in https://github.com/sul-dlss/SearchWorks/issues/5055 that the links are what we want for now.

jcoyne:
@marlo-longley this requires a rebase.

app/views/catalog/_home_page.html.erb
```diff
@@ -59,26 +59,26 @@
             </header>
             <ul class="list-unstyled">
               <li class="mb-3">
-                <div class="fw-bold">Start with a simple keyword search</div>
-                Examples: “the new york times” or “stephan graham thesis”
+                <span class="">Start with a simple keyword search</span>
```
jcoyne, line 62: Can we use the bootstrap class `fw-medium` instead?

app/assets/stylesheets/searchworks4/homepage.css
```diff
@@ -75,6 +75,10 @@
     font-weight: 700;
     color: var(--stanford-cardinal);
   }
+  .tips span {
+    color: var(--bs-header-color);
```
jcoyne, line 79: Where is `--bs-header-color` defined?  I don't see it being set anywhere.

jcoyne (changes requested):

app/assets/stylesheets/searchworks4/homepage.css
marlo-longley: I can just set it to black rather than the variable. Without setting the color it defaults to `--bs-body-color` which to me looks lighter than the provided design. If you have a particular way you would like me to set black, let me know.

app/views/catalog/_home_page.html.erb
```diff
@@ -58,26 +58,26 @@
         </header>
         <ul class="list-unstyled">
           <li class="mb-3">
-            <div class="fw-bold">Start with a simple keyword search</div>
-            Examples: “the new york times” or “stephan graham thesis”
+            <span class="fw-medium">Start with a simple keyword search</span>
```
jcoyne, line 61: Why did these change from div to span?  Don't we want a block element here?  Doesn't the `<p>` added below add extra margin at the bottom?

PR #5110, opened 2025-06-11
Add `blacklight-search-storage` meta tag from upstream

cbeer:
I'm not sure what this does, but it's present in the upstream partial and maybe we want/need it?


jcoyne:
I've asked Ben what this is about.  If it's not useful, we may want to remove it from blacklight.

jcoyne:
Ben indicated this is a hook from a not yet implemented feature.  So we aren't depending on it.

PR #5130, opened 2025-06-13
update stanford_only_icon to be a popover

dnoneill:
Right now the stanford only icon next to an article is static. It should be a popover

Before:
Screenshot 2025-06-13 at 2 58 55 PM

After:
Screenshot 2025-06-13 at 2 58 45 PM


jcoyne:
@dnoneill yes, but we can check the response (we will cease using the current html field) and write our own markup and scripts.

dnoneill:
I am confused. Why would we want to check the response on the SearchWorks side?
Screenshot 2025-06-16 at 1 55 24 PM

PR #5165, opened 2025-06-19
Reorder facets per designs

jcoyne:


app/javascript/controllers/facet_list_controller.js
```diff
@@ -0,0 +1,55 @@
+import { Controller } from "@hotwired/stimulus"
+
+(function() {
+  const BREAKPOINT = 1400; // xxl Bootstrap breakpoint
+  let isNarrow = window.innerWidth < BREAKPOINT;
+
+  window.addEventListener('resize', function() {
```
cbeer, line 7: Do we need to do this with javascript? What's javascript do for us that CSS can't?

app/views/catalog/_search_results.html.erb
```diff
@@ -48,7 +48,7 @@
       <div class="sidebar col-12 col-md-4">
         <!-- TODO: search mode toggle -->
 
-        <%= render Blacklight::Response::FacetGroupComponent.new(id: nil, title: 'Filters') do |component| %>
+        <%= render Facets::GroupComponent.new(id: nil, title: 'Filters') do |component| %>
```
cbeer, line 51: Would the implementation be easier if we split the facets into separate groups (collapsible + non-collapsible?) Maybe that's the explicit identification I'm suggesting in https://github.com/sul-dlss/SearchWorks/pull/5165/files#r2157474808 🤷‍♂️

app/components/facets/group_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+    <div class="text-center mt-2 d-xxl-none">
+      <button class="btn btn-primary" data-facet-list-target="showButton" data-action="facet-list#show" hidden>
+        <i class="bi bi-sliders" aria-hidden="true"></i> Show all filters
+      </button>
+
+      <button class="btn btn-primary" data-facet-list-target="hideButton" data-action="facet-list#hide">
+        <i class="bi bi-sliders" aria-hidden="true"></i> Hide filters
+      </button>
+    </div>
```
cbeer, line 21: Is this accessible? Toggling between totally different buttons might be annoying, and maybe we need `aria-expanded`, `aria-controls` or something?

jcoyne: I'm not sure about that.  We aren't expanding a container here, so I don't think aria-expanded would be exactly right. Then the expanded would have to be reset if we resized to a bigger page?  This is beyond my understanding of how that is all suppose to work.

cbeer: > We aren't expanding a container here, so I don't think aria-expanded would be exactly right.

Maybe I'm missing something about this implementation. I thought these buttons were controlling the visibility of`<div data-facet-list-target="hideFacets">`?  That seems like an entirely appropriate use of `aria-expanded`:

> The aria-expanded attribute is set on an element to indicate if a control is expanded or collapsed, and whether or not the controlled elements are displayed or hidden.
>There are several widgets that can be expanded and collapsed, including menus, dialogs, and accordion panels. Each of these objects, in turn, has an interactive element that controls their opening and closing. The aria-expanded attribute is applied to this focusable, interactive control that toggles the visibility of the object.

---

> Then the expanded would have to be reset if we resized to a bigger page

Yes, I think that's what ARIA would have you do, but I thought these controls totally disappear at the larger breakpoint.

jcoyne: Okay, I was interpreting that as "fully collapsed/hidden", where ours is only partially collapsed.

jcoyne:
I've redone it with the single button + aria attributes

taylor-steve:
This is working well for to me. I think we need to remove the ambiguity from `Any facet fields that are currently selected should likely appear by default` from the original issue.

After testing this, I'm not sure I love the idea of pulling facets out of the given facet order (which is how I read that statement). I like that this PR keeps them in order, I find a consistent order important. I'm wondering if the question is "should the facet list be expanded by default if a facet from the collapsable group has an existing selection". We might need @dbranchini or did you all already talk about this when I was gone?

cbeer:
> should the facet list be expanded by default if a facet from the collapsable group has an existing selection

We haven't talked about it, but that proposal sounds like a good starting point.

app/components/facets/group_component.html.erb
```diff
@@ -0,0 +1,18 @@
+<% # main container for facets/limits menu -%>
+<%= content_tag :div, id: @id, class: 'facets sidenav facets-toggleable-md',
+  data: { controller: 'facet-list', action: 'resize-above@window->facet-list#deactivate resize-below@window->facet-list#activate' } do %>
```
cbeer, line 3:
```suggestion
  data: { controller: 'facet-list' } do %>
```

Were these left-over, or doing some kind of magic?

app/components/facets/group_component.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <div class="facets-header">
+    <%= content_tag :h2, @title, class: 'facets-heading h4' if @title %>
+
+    <%= collapse_toggle_button(@panel_id) %>
+  </div>
+
+  <div id="<%= @panel_id %>" class="<%= body_classes %>" data-facet-list-target="facet">
```
cbeer, line 10:
```suggestion
  <div id="<%= @panel_id %>" class="<%= body_classes %>">
```

Same here. I don't see this target registered on the controller?

PR #5185, opened 2025-06-20
Sanitize bad records to prevent upstream citeproc + mla issues

### LINKED ISSUE #5181: [SearchWorks/prod] NoMethodError: undefined method 'end_with?' for nil
## Backtrace

    line 32 of [PROJECT_ROOT]/app/models/citations/marc_citation.rb: Citations::MarcCitation#citation
    line 23 of [PROJECT_ROOT]/app/models/citations/marc_citation.rb: block in Citations::MarcCitation#all_citations
    line 20 of [PROJECT_ROOT]/app/models/citations/marc_citation.rb: Array#each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/120844865)

jcoyne:
See https://github.com/inukshuk/citeproc-ruby/issues/88

Fixes #5181


cbeer:
Should this have a test? 

and, what do you think about adding some kind of `rescue` to make sure we're not blowing up the whole page if we can't render a citation?

PR #5245, opened 2025-06-26
fix flashing button bug

dnoneill:
Before:

https://github.com/user-attachments/assets/755e3557-12f9-4f32-a283-6fb4cc70a51c

After:

https://github.com/user-attachments/assets/21f28ffe-2d01-48fb-b874-f0dd80b85e1e


jcoyne (commented):
Can you port this up into blacklight?

jcoyne (approved):

PR #5251, opened 2025-06-27
Fix up the constraints component styling.

cbeer:
Before (little to no padding between items, `x` floating off in space):
Screenshot 2025-06-26 at 17 05 46

After:
Screenshot 2025-06-26 at 17 05 55


cbeer:
I'm not sure what you're suggesting I should do. I put the CSS from 148 into my branch, and the spacing is maybe slightly tweaked but still not good, and there are significant changes the hover/focus states (is that even desired?)

PR #5259, opened 2025-06-27
Flash messages need a margin

jcoyne:
before:
Screenshot 2025-06-26 at 10 20 43 PM

After

Screenshot 2025-06-26 at 10 24 30 PM


cbeer:
Won't this create the margin whether there are flash messages or not? It looks like our other pages have the search bar inject the margin:
Screenshot 2025-06-27 at 06 52 00
Screenshot 2025-06-27 at 06 52 21

(I'm not sure that's good practice, but I think we could be consistent with what element is supposed to apply the spacing.)

PR #5289, opened 2025-06-27
update modal background

dnoneill:
Screenshot 2025-06-27 at 3 13 10 PM


jcoyne:
Is there a ticket that should be linked to this work?

dnoneill:
Quite possibly. We can check. The figma does show a background but I can't get the values. Figured at the very least whoever does get this will know how to set the variable and not have to go hunting.

PR #5292, opened 2025-06-27
Facet toggle button style fixes

jcoyne:
Display as blue. Don't display the header and toggle button at the same time

### DIFF
--- a/app/components/facets/group_component.html.erb
+++ b/app/components/facets/group_component.html.erb
@@ -3,7 +3,7 @@
   <div class="facets-header">
     <%= content_tag :h2, @title, class: header_classes if @title %>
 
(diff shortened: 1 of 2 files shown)


dnoneill:
Pretty sure this is unnecessary https://github.com/sul-dlss/SearchWorks/pull/5281 should fix the issue.

dnoneill:
Also this doesn't fix the issue, I think you need to update this line and it will work

```Facets::GroupComponent.new(id: nil, title: 'Filters', header_classes: 'facets-heading')```
to
```Facets::GroupComponent.new(id: nil, title: 'Filters', header_classes: 'facets-heading d-none d-md-block')```

Issue #5058, opened 2025-06-03
Update FOLIO cached data in the release branch


cbeer:
After #5057, we should make sure the FOLIO data updates for folio-prod are also applied to the release branch.

PR #5309, opened 2025-06-30
Update FOLIO types

sul-devops-team:


cbeer:
@jcoyne : there's something still not right about this FOLIO data sync'ing to the release branch (from #5058). It's made exactly the same commit + PR every day:

https://github.com/sul-dlss/SearchWorks/pull/5301
https://github.com/sul-dlss/SearchWorks/pull/5262

PR #5311, opened 2025-06-30
Add google cover image controller to the record summary in the modals

### LINKED ISSUE #5303: Add bookcover image to email and cite modals
The image is missing from the email and cite modals. In testing this on stage, I tried emailing [this record](https://searchworks-stage.stanford.edu/view/13407275), which has a book cover image. See screenshot of modal.

Image

**As designed -- see below.**

jcoyne:
Fixes #5303


cbeer:
I'm going to merge this PR to get the book cover, but it seems like we might benefit from some refactoring around the google-cover-image controller so it's less confusing when we have to add it or when the top-level version is all we need, right (I'm thinking either to put the data-controller statement closer to each thumbnail and figure out how to batch independent controllers, or hook up `imageTargetConnected` to handle added content?)

PR #5344, opened 2025-07-01
Add hover styles and tooltips to icons

### LINKED ISSUE #5318: Hover states for icon-only links (examples - save aka bookmark and ellipses)
### Context

- We need to add hover states for our icon only links. 

### Design previews

<br>
<br>
<br>

### Designs with specifications

- [Save (aka bookmark)](https://www.figma.com/design/K1syTsalaB1nlRftgOi0Pw/DLSS-Components?node-id=3124-5410&t=GL63gQ4Cp5XQWoWi-4)
  - Default and hover states for an unsaved record and a saved record

- [More actions](https://www.figma.com/design/K1syTsalaB1nlRftgOi0Pw/DLSS-Components?node-id=3142-11762&t=GL63gQ4Cp5XQWoWi-4)
  - Default and hover states

### Design notes

- Hover states with tooltips were already implemented for the viewers -- see screenshot below. Accessibility was taken into consideration per Alan.
- Save hover states should display a tooltip that reads "Save record" or "Remove from saved records" (depending on state)
- Ellipses hover state should display a tooltip that reads "More actions"

Image

jcoyne:
I converted the bookmarks to use a turbo-stream rather than the blacklight javascript.  This allows us to switch icons, launch a toast, deal with hover all in a single spot. This switches to a simple `button`, which I believe makes keyboard navigation better. 

Fixes #5318

Screenshot 2025-07-01 at 1 31 43 PM

Screenshot 2025-07-01 at 1 31 40 PM
Screenshot 2025-07-01 at 1 32 36 PM


taylor-steve:
In #5318 I'm seeing:
* The add text should be `Save record`
* The saved icon state is a checkmark

For keyboard nav, if I bookmark using using the keyboard, the button loses focus. It's next in the tab order after that, so I'm not sure how big an issue this is.

jcoyne:
@taylor-steve sorry. I was getting confused between all the states. Here's what we've got now:
Screenshot 2025-07-03 at 12 38 57 PM
Screenshot 2025-07-03 at 12 38 48 PM
Screenshot 2025-07-03 at 12 38 44 PM

taylor-steve:
@jcoyne Thanks, that all looks correct to me now. One thing @dnoneill helped spot when the group was looking over this, it seems we're not completely handling failure. If saving/unsaving the record fails on the backend, the success toast is still displayed. The bookmark icon remains in the correct state. How difficult would it be to address the toast?

jcoyne:
Do we have designs for the failing state yet?

taylor-steve:
We do not. I'd like to merge this sooner rather than later. I think simply only showing a success toast when we have success would be a fine start. If that's not straightforward or would complicate the future work, we can merge now and I'll capture that as a todo for WC3.

taylor-steve (approved):

PR #5347, opened 2025-07-01
Update the record page styling

cbeer:
Part of #5332


app/assets/stylesheets/searchworks4/typography.css
```diff
@@ -77,3 +77,8 @@ aside,
 .text-poppy {
   color: var(--stanford-poppy-dark);
 }
+
+.btn-link {
```
jcoyne, line 81: Should this be in the component library?

cbeer: I don't know. I assume the component library (intentionally?) broke bootstrap's defaults and have no interest in going down that rabbit hole.

jcoyne (approved):
Feel free to merge after considering suggestions around "subnavbar".

app/assets/stylesheets/searchworks4/typography.css
jcoyne: It set the values based on the specifications given to us by DLUX.  I think it makes sense to have a single correct style for `.btn-link`.  If it's not correct in the component library, we ought to get it fixed.

taylor-steve:
I will create an issue for the subnavbar spec

PR #5382, opened 2025-07-02
Eds error

### LINKED ISSUE #5182: [SearchWorks/prod] ArgumentError: Attributes per element limit exceeded
## Backtrace

    line 16 of [PROJECT_ROOT]/app/models/eds/session.rb: Eds::Session#search
    line 39 of [PROJECT_ROOT]/app/models/eds/repository.rb: Eds::Repository#eds_search
    line 7 of [PROJECT_ROOT]/app/models/eds/repository.rb: block in Eds::Repository#search

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50022/faults/119440764)

hudajkhan:
Closes #5182 

Context: The "attributes per element limit exceeded" error occurs in certain conditions where the HTML has errant "<" characters and very long text.  The parser identifies these as HTML elements with the text that follows (before any other opening tags) being identified as attributes. 

What this PR does:

- Increases the number of maximum attributes allowed for Sanitize so errors are not thrown for the original case linked in the Honeybadger report.
- Also adds a rescue condition so that if an argument error is thrown since the number of attributes has been exceeded even beyond the new limit, the code will send a honeybadger notification sending the HTML that triggered the error.  The code will also replace the html with an empty string so the page can still function. 
- Sets up related tests

### DIFF
--- a/app/models/eds_document.rb
+++ b/app/models/eds_document.rb
@@ -360,9 +360,9 @@ class EdsDocument # rubocop:disable Metrics/ClassLength
     rescue ArgumentError => e
       # We still want to notify honeybadger but can provide more information
       # about the html
-      Honeybadger.notify(e, error_message: "Error with arguments passed to Sanitize/Nokogiri", context: { html: })
-      # Replacing the html with an empty string to allow the page to still load
-      sanitized_html = ""
+      Honeybadger.notify(e, error_message: 'Error with arguments passed to Sanitize/Nokogiri', context: { html: })
+      # Return an empty string to allow the page to still load
+      return ''
     end
     # sanitize 5.0 fails to restore element case after doing lowercase
     sanitized_html = sanitized_html.gsub('<searchlink', '<searchLink')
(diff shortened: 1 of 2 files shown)


app/models/eds_document.rb
```diff
@@ -351,7 +355,15 @@ def html_decode_and_sanitize(data, config = nil)
...
 
-    sanitized_html = Sanitize.fragment(html, sanitize_config)
+    begin
+      sanitized_html = Sanitize.fragment(html, sanitize_config)
+    rescue ArgumentError => e
+      # We still want to notify honeybadger but can provide more information
+      # about the html
+      Honeybadger.notify(e, error_message: "Error with arguments passed to Sanitize/Nokogiri", context: { html: })
+      # Replacing the html with an empty string to allow the page to still load
+      sanitized_html = ""
```
cbeer, line 365: We can just return the fallback, right? There's no need to do any of the other processing with the empty string or anything.

```suggestion
     return ''
```

cbeer:
Good question. I don't know if HB truncates any of that for us. I guess we can try it and find out 🤷‍♂️

PR #5385, opened 2025-07-03
Restore default bootstrap padding to the header menus.

cbeer:
Darcy thinks the current menu is crunched. I'm not sure why the [component library makes the padding so small](https://github.com/sul-dlss/component-library/blame/main/styles/header.css#L168-L198), but the regular bootstrap padding looks better:

Before:
Screenshot 2025-07-02 at 18 53 07

After:
Screenshot 2025-07-02 at 18 52 35


cbeer:
What's the status of that? Can we merge this in the meantime so it looks better now?

jcoyne:
At standup today @dbranchini said we can fix this in the component library

cbeer:
Do you want to do that right now, or can we merge this PR and clean it up later? I'm concerned we're going to spend a lot of time trying to figure out where the existing style came from and remediate other applications.

PR #5396, opened 2025-07-03
Fix a11y errors in the help menu

### LINKED ISSUE #5330: i18n errors: Help menu
- [ ] `aria-label` on span 
Image

- [ ] Role not inside the required context (role="listitem"Interlibrary loan, role="menuitem"Interlibrary loan)
> If the menuitem is not a descendant of a menu in the DOM, include the [aria-owns](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-owns) attribute on menu to indicate the relationship.

Image
(etc)

cbeer:
Fixes #5330

### DIFF
--- a/app/views/shared/_search_navbar_services_menu.html.erb
+++ b/app/views/shared/_search_navbar_services_menu.html.erb
@@ -1,11 +1,11 @@
 <ul class="dropdown-menu sul-services-menu" role="menu" aria-labelledby="searchNavbarTopMenu" id="libraryServicesMenu">
-  <li><span class="dropdown-item-text text-uppercase text-secondary fw-semibold"><small>Need assistance?</small></span></li>
-  <li><a href="https://library.stanford.edu/contact-us" class="dropdown-item" role="menuitem">Ask a librarian</a></li>
-  <li><a href="https://guides.library.stanford.edu/searchworks" class="dropdown-item" role="menuitem">SearchWorks user guide</a></li>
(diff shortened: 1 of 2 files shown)


app/views/shared/_search_navbar_services_menu.html.erb
```diff
@@ -1,11 +1,11 @@
 <ul class="dropdown-menu sul-services-menu" role="menu" aria-labelledby="searchNavbarTopMenu" id="libraryServicesMenu">
-  <li><span class="dropdown-item-text text-uppercase text-secondary fw-semibold"><small>Need assistance?</small></span></li>
-  <li><a href="https://library.stanford.edu/contact-us" class="dropdown-item" role="menuitem">Ask a librarian</a></li>
-  <li><a href="https://guides.library.stanford.edu/searchworks" class="dropdown-item" role="menuitem">SearchWorks user guide</a></li>
-  <li><a href="https://guides.library.stanford.edu/c.php?g=1081009&p=7878021" class="dropdown-item" role="menuitem">Access to online articles</a></li>
-  <li><a href="https://library-status.stanford.edu/" class="dropdown-item" role="menuitem">Check system status</a></li>
+  <li role="presentation"><span class="dropdown-item-text text-uppercase text-secondary fw-semibold"><small>Need assistance?</small></span></li>
```
jcoyne, line 2: If a `role="presentation"` wraps a `role="menuitem"`, does that remove the menuitem from the accessibility tree?

jcoyne: Good enough.

PR #5412, opened 2025-07-03
Force the library header link to be semibold.

cbeer:
The component library forces links to be regular weight 😦


app/components/access_panels/library_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="d-flex flex-column mt-2 mb-3 panel-library-location" data-hours-route="<%= hour_path(library.code) if library.holding_library? %>">
   <div class="library-location-heading bg-light ms-n3 ps-3 py-1 position-relative">
-    <h3 class="h4 p-0 m-0 fw-semibold"><%= link_to_library_about_page class: 'text-body stretched-link' %></h3>
+    <h3 class="h4 p-0 m-0 fw-semibold"><%= link_to_library_about_page class: 'text-body fw-semibold stretched-link' %></h3>
```
jcoyne, line 3:
```suggestion
    <h3 class="h4 p-0 m-0"><%= link_to_library_about_page class: 'text-body fw-semibold stretched-link' %></h3>
```

cbeer: I think we need both. If it's a link, we want to be semibold, and if it's not a link, we want it to be semibold too.

Issue #5334, opened 2025-07-01
Build out facet in-place searching


cbeer:
<img width="452" alt="Image" src="https://github.com/user-attachments/assets/6ce542f9-136e-4b15-a6df-9a55597eab77" />

PR #5439, opened 2025-07-08
Stub out facet searching

cbeer:
Screenshot 2025-07-07 at 21 05 54

See #5334 

TODO:
- [ ] hide the dropdown on blur (and restore on focus)
- [ ] make the dropdown scrollable (especially on small screens)
- [ ] get a hover background color
- [ ] keyboard accessibility (key up + down should move through the search box + menu) + aria attributes


cbeer:
Huh. I don't know where I found that.

PR #5457, opened 2025-07-08
Move Access Conditions to an access panel and display in the sidebar

### LINKED ISSUE #5398: Update catalog#show aside for MODS records
Image

I think the only change from the current behavior is moving the access conditions from the main content into the sidebar.

jcoyne:
Fixes #5398

Screenshot 2025-07-08 at 3 07 59 PM


cbeer:
This seems like a legitimate test failure, right?

jcoyne:
@cbeer yes, I've removed that assertion now.

Issue #5449, opened 2025-07-08
Theses and dissertations - featured resource


dbranchini:
### Context

- We already have this featured resource in production. Users will get to this page from the home page or from the top right navigation labeled, "Featured resources."

### Design with specifications

- [Figma design](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=1331-133320&t=wOQbHg46ijl8j6if-4)

### Design notes

- Top filters include Access, Stanford school or department, and Type of work.
- Checkbox for "Show Stanford work only" prompts a reload of the page and narrows the results to Stanford only works.
  - Note this is a change from the current behavior. Currently, you have to either select a type of work or a Stanford school or department in order to see Stanford only work. 

### Related tickets

- #5283 (note that the featured resources design patterns are outlined in this ticket)
- #4881 
- #5443 
- #5450

PR #5494, opened 2025-07-10
Add Stanford only checkbox to Theses/Dissertation page

marlo-longley:
This should close [#5449](https://github.com/sul-dlss/SearchWorks/issues/5449)

Adding a `:js` makes the spec file run a lot slower -- up for debate if the test is worth the slowdown.

Screenshot 2025-07-09 at 10 00 25 PM

### DIFF
--- a/app/components/searchworks4/theses_dissertation_facet_header_component.html.erb
+++ b/app/components/searchworks4/theses_dissertation_facet_header_component.html.erb
@@ -1,0 +1,34 @@
+<div class="p-2 mb-2 d-flex justify-content-center" data-controller='facet-checkboxes'>
+  <button class="btn btn-outline-secondary p-3 position-relative">
+    <%= label_tag("facet_option_stanford_only") do %>
+      <%= check_box(
+            'f',
+            'stanford_only',
+            {
+              multiple: true,
+              id: "facet_option_stanford_only",
+              checked: stanford_only_selected?,
+              data: { action: 'facet-checkboxes#toggleCheckbox' }
+            },
+            "Show Stanford work only",
+            nil
+          )
+      %>
+      <%= link_to(
+            stanford_only_selected? ? theses_all_path : theses_stanford_only_path,
+            data: {
+              newState: !stanford_only_selected?,
+              action: 'facet-checkboxes#toggleLink',
+              turbo: false
+            },
+            class: 'ms-1 btn',
+          ) do %>
+        Show Stanford work only
+      <% end %>
+    <% end %>
+  </button>
+
+  <div data-facet-checkboxes-target="loadingIndicator" class="d-none position-absolute opacity-75 top-50 start-50 translate-middle">
+    <div class='loading-spinner'></div>
+  </div>
+</div>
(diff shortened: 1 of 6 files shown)


spec/features/access_points/dissertation_theses_spec.rb
```diff
@@ -30,4 +30,23 @@
...
+  it 'Toggles the Stanford work facet' do
+    expect(page).to have_css('#facet_option_stanford_only', visible: :visible)
+
+    checkbox = find_by_id('facet_option_stanford_only')
+
+    checkbox.click
+
+    checked = find_by_id('facet_option_stanford_only')
+    expect(page).to have_css('.filter-stanford_work_facet_hsim')
+    expect(checked.checked?).to be(true)
```
jcoyne, line 43: I'd prefer a check that shows this actually filters documents in/out from the result set.  Checking that "checked" is set can be tested in a component spec with much less overhead than a integration test.

jcoyne (changes requested):

app/components/searchworks4/theses_dissertation_facet_header_component.html.erb
```diff
@@ -0,0 +1,35 @@
...
+              id: "facet_option_stanford_only",
+              checked: selected,
+              data: { action: 'facet-checkboxes#toggleCheckbox' }
+            },
+            "Show Stanford work only",
+            nil
+          )
+      %>
+      <%= link_to(
+            selected ? unchecked_url : checked_url,
```
jcoyne, line 18: Doesn't this just overwrite any other facets that are also selected?

jcoyne: Can this confitional be moved into the class (e.g. create a `url` method).

app/components/searchworks4/theses_dissertation_facet_header_component.html.erb
```diff
@@ -0,0 +1,35 @@
...
+              id: "facet_option_stanford_only",
+              checked: stanford_only_selected?,
+              data: { action: 'facet-checkboxes#toggleCheckbox' }
+            },
+            "Show Stanford work only",
+            nil
+          )
+      %>
+      <%= link_to(
+            stanford_only_selected? ? all_theses_dissertations_url : stanford_only_url ,
```
jcoyne, line 18: I think this ought to be using FacetItemPresenter#href. https://github.com/projectblacklight/blacklight/blob/508f67a95ea1a53884f199ef594b8aa6e8fb9483/app/presenters/blacklight/facet_item_presenter.rb#L81-L87

jcoyne:
@marlo-longley this needs a rebase.

app/components/searchworks4/theses_dissertation_facet_header_component.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      %>
+      <%= link_to(
+            stanford_only_selected? ? theses_all_path : theses_stanford_only_path,
+            data: {
+              newState: !stanford_only_selected?,
+              action: 'facet-checkboxes#toggleLink',
+              turbo: false
+            },
+            class: 'ms-1 btn',
+            id: 'stanford-only-facet-option'
```
cbeer, line 25:
```suggestion
            class: 'ms-1 btn text-secondary'
```

Does this need an id and custom styling, or could we just use a class?

cbeer (approved):

PR #5515, opened 2025-07-10
add note-highlight css

### LINKED ISSUE #5502: Item and location notes
Let's use the same treatment for both location-level notes (such as latest) and item notes. As an example, I've added both to this [availability panel](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=950-59213&t=oTubIcWyZd3wMlb4-4). 

It's Stanford accent color Illuminating Base at 20%.

Image

dnoneill:
closes #5502 

Screenshot 2025-07-10 at 12 53 31 PM


cbeer:
Got it - I was confused looking at the screenshots.

PR #5551, opened 2025-07-11
Remove jquery dependency

jcoyne:


cbeer:
Is the click handler on https://github.com/sul-dlss/SearchWorks/blob/main/app/components/searchworks4/citations/citation_component.html.erb#L37 a little bit of left-over jquery?

jcoyne:
@cbeer Thank you. I've updated that code too.

PR #5552, opened 2025-07-11
Remove unused select-all.js and sort_and_per_page partial

jcoyne:


cbeer:
I think see all might be sticking around -- see #5632

taylor-steve (approved):
After looking at #5632 more closely, I agree, I think we can merge this now

PR #5578, opened 2025-07-11
fix escaping on marc fields

### LINKED ISSUE #5543: Encoded ampersands show in the physical description
See e.g.  https://searchworks-uat.stanford.edu/view/4817330

Image

dnoneill:
closes #5543 

I don't know how to do this without html safe. I could use help
Screenshot 2025-07-11 at 4 55 15 PM

### DIFF
--- a/app/components/record/marc_document_component.html.erb
+++ b/app/components/record/marc_document_component.html.erb
@@ -10,7 +10,7 @@
     <% end %>
 
     <div class='upper-record-metadata mb-3'>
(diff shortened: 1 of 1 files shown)


cbeer:
Isn't the problem here that we're unsanitizing it with the join:

https://github.com/sul-dlss/SearchWorks/blob/main/app/components/record/marc_document_component.html.erb#L13
```
<%= document.marc_field('3003abcefg').values.join(' ') %>
```

and should have something like this instead:

```
<%= safe_join(document.marc_field('3003abcefg').values, ' ') %></div>
```

dnoneill:
That fixes it. Thank you.

PR #5593, opened 2025-07-14
setting up proxy links for popular database links

### LINKED ISSUE #5539: Popular databases links need to go through ezproxy.
The popular databases links were hard-coded in the template:
https://github.com/sul-dlss/SearchWorks/blob/main/app/views/databases/_popular_databases.html.erb

They need to have the ezproxy prefix applied (e.g.  https://github.com/sul-dlss/SearchWorks/blob/main/app/models/links/ezproxy.rb is how we do this for regular 856 links).

hudajkhan:
Closes #5539 

What this PR does:

- Adds two URLs from the database links that were not already in the sul proxy list (research.ebsco.com and scholar.google.com)
- Allow the EZProxy class to handle a situation where there is no Solr document (i.e. for the database links)
- Sets up a helper to get the proxied url. The helper also appends text to the title which triggers the proxy URL to be formed.

### DIFF
--- a/app/models/links/ezproxy.rb
+++ b/app/models/links/ezproxy.rb
@@ -52,8 +52,9 @@ class Links
         sul_restricted?
     end
 
+    # Popular database links may not have any associated Solr documents
     def libraries
-      @libraries ||= document.fetch(:holdings_library_code_ssim, [])
+      @libraries ||= document&.fetch(:holdings_library_code_ssim, []) || []
     end
 
     def ezproxied_hosts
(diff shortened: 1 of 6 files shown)


app/models/links/ezproxy.rb
```diff
@@ -52,8 +52,9 @@ def apply_sul_proxy_prefix?
         sul_restricted?
     end
 
+    # Popular database links may not have any associated Solr documents
```
jcoyne, line 55: 👍

jcoyne (approved):

Issue #5597, opened 2025-07-14
For record view, don't duplicate display of Finding aid links


saseestone:
In record view, we're currently displaying Finding aid links in the Availability panel and under Contents/Summary. Example record: https://searchworks-uat.stanford.edu/view/in00000217727

We shouldn't duplicate the links in the Contents/Summary section in the content panel. We should only show them in the Availability panel. 

<img width="927" height="528" alt="Image" src="https://github.com/user-attachments/assets/98b89ed8-4ea3-4315-b5b4-bc7ffb9a834e" />

PR #5598, opened 2025-07-14
Don't duplicate display of finding aids

jcoyne:


jcoyne:
@cbeer I looked at 5733700 and it seems to also being duplicated.  Do we all understand the [ticket](https://github.com/sul-dlss/SearchWorks/issues/5597) in the same way?
Screenshot 2025-07-15 at 9 48 05 AM

PR #5600, opened 2025-07-14
Toggle between pages when changing search modes from the home page

cbeer:
https://github.com/user-attachments/assets/592fa625-e505-457d-9fa5-954c761f7b10


dbranchini:
Yes, let's try it.

cbeer:
I guess let's get the page navigation behavior and tackle query persisting later 🤷‍♂️

Issue #5548, opened 2025-07-11
Donor page with masthead and gallery view


dbranchini:
### Context

- we have specialized donor pages that a user can get to via a digital bookplate link from a record view 
  - [see example](url) 
  - click on the donor link 
  - and you'll go to [this page](url)
- users might also get to this page when a link to the page is sent via email 

### Designs with specifications

- [Donor page with masthead and gallery view](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=3122-154996&t=SFu8RukxxJKoPJ3Q-4)

### Design notes

- Image is constrained to a max width of 200 px
- Link "support teaching, learning and research at Stanford Libraries" goes to https://library.stanford.edu/support-stanford-libraries.

### Related  

**NOT ticketed yet** -- similar masthead needed for collections on results page

- [Collection results page with masthead (and regular results list view)](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=2308-225004&t=SFu8RukxxJKoPJ3Q-4)

**Several tickets on browse nearby aka related items aka call number browse

taylor-steve:
Example from current prod: https://searchworks.stanford.edu/catalog?f%5Bfund_facet%5D%5B%5D=tf882hn2198&sort=new-to-libs&view=gallery

PR #5604, opened 2025-07-14
Move and restyle donor masthead

taylor-steve:
Ref #5548 

Makes the requested masthead change (moved above the contraints in a Slack conversation) and pulls in the gallery styling. Widening the gallery view at XXL size and bringing back the view mode selection will be in subsequent PRs.

Screenshot 2025-07-14 at 3 21 22 PM


jcoyne:
Can we perhaps merge Global alerts first https://github.com/sul-dlss/SearchWorks/pull/5585 and then see if this works well with that?

Issue #4979, opened 2025-05-20
Browse related scrolls independently of the preview


jcoyne:
<img width="1224" alt="Image" src="https://github.com/user-attachments/assets/08018f7a-d964-418d-a702-a95ad1f6fd05" />

The user can get lost as what they are previewing is no longer pointing up at the right thing.

PR #5610, opened 2025-07-15
Get the arrow to travel with the element it points at.

cbeer:


jcoyne:
I believe this is addressing https://github.com/sul-dlss/SearchWorks/issues/4979

PR #5611, opened 2025-07-15
Persist the q parameter when toggling between different search modes (but don't persist it after searching)

### LINKED ISSUE #5616: Toggle b/n Catalog and Articles+ on home and expected behavior
When a user is on Catalog and selects Articles+, the page should change to the Articles+ home page. And the opposite is true too. When a user is on Articles+ and selects Catalog, the page should change to the Catalog home page. If a user has already entered a keyword search, that search term should persist.

cbeer:
Fixes #5616


cbeer:
Yes!

taylor-steve (approved):

PR #5619, opened 2025-07-15
Load the collection in the controller

jcoyne:
Interfacing with external services like solr ought to be done in the controller rather than in the view


marlo-longley:
Does this PR have implications for how other components are loading `page_location` from helpers? Such as 

https://github.com/sul-dlss/SearchWorks/blob/b834d0f7faf48b023b8d4a2a32e8dbbdb9bf5da9/app/components/searchworks4/digital_collections_facet_header_component.rb#L5

https://github.com/sul-dlss/SearchWorks/blob/b834d0f7faf48b023b8d4a2a32e8dbbdb9bf5da9/app/components/searchworks4/search_result/collection_info_component.rb#L14

jcoyne:
@marlo-longley no, I don't think anything is changed about it. Its still in the helper context.

PR #5642, opened 2025-07-15
Add gallery view type selection

taylor-steve:
Part of #5548 

* Clears the view mode when the fund facet is removed or the masthead search is used

* Switches to gap instead of margin to control gallery view. I think this makes it easier to get closer to the design.

* Switches to uses d-none/d-block switching instead of inline style for previews and initializes the preview with d-none. This helps avoid the case where the zero height "closed" previews influence the flex layout (e.g., due to margin)

* Bumps the width on the document so common text doesn't wrap (e.g., 'Off-campus collections')
Screenshot 2025-07-15 at 12 52 31 PM

### DIFF
--- a/app/views/catalog/_view_type_group.html.erb
+++ b/app/views/catalog/_view_type_group.html.erb
@@ -1,13 +1,5 @@
-<%= render(Blacklight::Response::ViewTypeComponent.new(
+<%= render(ViewTypeDropdownComponent.new(
       response: controller_name == 'browse' ? [true] : @response, # Allows browse controller to reneder without a @response
       views: document_index_view_controls,
       search_state: search_state,
-      selected: document_index_view_type)) do |component| %>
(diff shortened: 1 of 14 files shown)


app/views/catalog/_view_type_group.html.erb
```diff
@@ -1,5 +1,13 @@
-<%= render(ViewTypeDropdownComponent.new(
```
jcoyne, line 1: Can you remove the ViewTypeDropdownComponent and ViewTypeDropdownItemComponent in here too?

cbeer:
> Adds the view selection when in gallery view so users can get back to list:

Do we need to be this clever, or could we kick `view` out of the params we remember in the search bar (which.. I think is the only way to get into that state now?)

jcoyne (changes requested):

PR #5647, opened 2025-07-15
Use existing search_state method

jcoyne:
Create and use `PageLocation#dissertation_theses?`

### DIFF
--- a/app/views/catalog/_facet_header.html.erb
+++ b/app/views/catalog/_facet_header.html.erb
@@ -5,8 +5,6 @@
       <div class='loading-spinner'></div>
     </div>
   </div>
-<% end %>    
-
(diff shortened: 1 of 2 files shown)


marlo-longley:
It's great to put `dissertation_theses?` into the appropriate class, but I don't understand why we can't keep delegating `:search_state` to helpers.

jcoyne:
@marlo-longley we could delegate search_state, but you passed in `search_state` when you initialized it.  It's weird to do it both ways.

https://github.com/sul-dlss/SearchWorks/blob/31a60e38a25acd20bdf610ac941101c5f857dc1a/app/views/catalog/_facet_header.html.erb#L11

PR #5654, opened 2025-07-16
Disable the reset button + zoomout buttons if the stackmap won't change

### LINKED ISSUE #5422: Stacks maps - UI
### Context

- Style stacks maps modal to align with other modals in the app
- A user can get to stacks maps from record view or search results -- see screenshots below 

Image

Image

### Designs with specifications

- [Modal - one call number](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=2556-186686&t=oTubIcWyZd3wMlb4-4) 
- [Modal - more than one call number](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=2562-194193&t=oTubIcWyZd3wMlb4-4) 
- [Tabs](https://www.figma.com/design/K1syTsalaB1nlRftgOi0Pw/DLSS-Components?node-id=3243-5628&t=BxW6wlY6G8jar7US-4) -- Added to the Figma component library (I think for now, add it to SW, and then later we can add to the component library)

### Design notes

- User can tab b/n map and text directions 
- Tabs are based on bootstrap tabs
  - Note the default, hover and active states in the designs
- User can also zoom in, out and reset the size
  - Resetting only does something if the user zooms in/out, so it should be disabled if the user has not zoomed in or out
- When there's more than one call number, it will default to the first one and the user can select a different one

jcoyne:
Fixes #5422

### DIFF
--- a/app/components/stackmap/map_component.html.erb
+++ b/app/components/stackmap/map_component.html.erb
@@ -18,10 +18,10 @@
         <%= button_tag class: "zoom-in btn btn-link p-sm-1", data: { action: 'click->stackmap#zoomIn' } do %>
           <i class="bi bi-zoom-in"></i><span class="d-none d-sm-inline"> Zoom in</span>
         <% end %>
-        <%= button_tag class: "zoom-out btn btn-link p-sm-1", data: { action: 'click->stackmap#zoomOut' } do %>
+        <%= button_tag class: "zoom-out btn btn-link p-sm-1", data: { stackmap_target: 'zoomOutButton', action: 'click->stackmap#zoomOut' } do %>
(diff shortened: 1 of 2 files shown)


cbeer:
After someone clicks Reset size, shouldn't the button get disabled again?

dnoneill:
I tried this and it didn't work. The button isn't updating. I don't think zoomstart is the correct listener.

dnoneill:
Nevermind it was a rebase issue

taylor-steve (approved):

PR #5667, opened 2025-07-16
Design tweaks

cbeer:


spec/components/searchworks4/availability_component_spec.rb
```diff
@@ -36,7 +36,6 @@
 
       aggregate_failures do
         expect(page).to have_css('.availability-component', text: 'Green Library')
-        expect(page).to have_link 'Stacks', href: "/view/10678312/GRE-STACKS/stackmap"
```
jcoyne, line 39: Why is this test removed?

cbeer: Because Darcy doesn't want to show the location link at the top level.

app/components/searchworks4/physical_availability_component.html.erb
```diff
@@ -100,9 +99,7 @@
               <div class="d-flex flex-column flex-sm-row gap-2 gap-sm-5 p-2 suppress-item-level-request-button-if-there-is-a-location-level-request-button">
                 <div class="pt-2" data-availability-target="location">
                   <div><%= render LocationComponent.new(location: location, document: document, suppress_off_campus: false, classes: []) %></div>
-                  <% unless document.holdings.libraries.one? && document.holdings.libraries.first.locations.one? %>
```
jcoyne, line 103: Wasn't this conditional in place to guard against bad records? Why is it being removed?

spec/components/searchworks4/availability_component_spec.rb
jcoyne: Its hard to understand the changes without having a written ticket or explanation of what is changing and why.

cbeer: I'm not sure I understand the suggestion. `AccessPanels::LibraryLocationComponent` is used in the sidebar on the catalog#show page, not here.

app/components/searchworks4/physical_availability_component.html.erb
cbeer: I believe you're thinking about a similar condition that's at the component level (checking if we should go down the single-location path or not). This condition is either gratuitous or wrong (if we're down this branch, we've already established there are multiple locations)

spec/components/searchworks4/availability_component_spec.rb
jcoyne: Yeah, I'm a bit confused on this PR.  There's a lot here.

jcoyne (approved):

PR #5680, opened 2025-07-16
Take the the user to a new page (hybrid of the home page + search res…

### LINKED ISSUE #5615: Clear all (filters/selections)
Currently on stage and uat, when a user clicks the "Clear all" button next to "Your selections," the page goes back to the home page. 

Instead, it should clear the selections (filters and keyword search), but stay on search results. See the empty search results design below. It's essentially the same content on the home page, but the featured resources is simplified into a list without descriptions.

Image

[Design with specifications](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=3362-66584&t=Qc6Q9Wg38hrgYO25-4)

cbeer:
…ults) when they click clear all

Fixes #5615

🤷‍♂️

### DIFF
--- a/app/views/catalog/_new_search.html.erb
+++ b/app/views/catalog/_new_search.html.erb
@@ -1,0 +1,91 @@
+<%# locals: (content_classes: document_index_view_type == :gallery ? 'col-md-8 col-xxl-9 px-xxl-4 ps-lg-4 ps-2' : 'col-md-8 col-xxl-6 px-xxl-4 ps-lg-4 ps-2', hide_constraints: false, hide_facets: false) %>
+
+<h1 class="visually-hidden">New search</h1>
+
+<div class="row align-items-start mt-4">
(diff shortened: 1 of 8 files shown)


app/views/catalog/_new_search.html.erb
```diff
@@ -0,0 +1,91 @@
...
+            <div data-facet-list-target="hideFacets" id="hidden-facets">
+              <%= render Blacklight::FacetComponent.with_collection(blacklight_config.facet_fields_in_group(nil).sort_by(&:label), response: @response) %>
+            </div>
+          <% end %>
+        <% end %>
+      </div>
+    </div>
+  </div>
+  <div class="col-lg-6 col-xl-8"><section class="px-lg-3 py-1 tips">
+      <header class="d-flex flex-wrap align-items-center h3"><span class="text-nowrap">Search tips</span>
```
cbeer, line 38: This content is almost the same but different from the home page search tips. I'm not sure trying to be clever and reconciling them is worth the added complexity, but could be done if we really cared 🤷‍♂️

taylor-steve (approved):

PR #5697, opened 2025-07-17
Have the stand-alone feedback form close the window and display the toast in the opening window.

cbeer:


dnoneill:
@cbeer Not happening in safari or chrome but it is happening in firefox. Maybe a cache thing? Can you check Firefox on yours?

dnoneill (approved):

PR #5721, opened 2025-07-18
Fix a problem with javascript scopes

cbeer:
`this` isn't the `this` we want it to be


jcoyne:
The arrow function operator would have done the trick too.

PR #5723, opened 2025-07-18
advanced clause, negative, and inclusive facet constraints with prefix

hudajkhan:


taylor-steve:
When I make an advanced search: 
Screenshot 2025-07-18 at 2 18 53 PM

I see this on the results page, which from there I make a standard search with a different term using the masthead searchbar:
Screenshot 2025-07-18 at 2 19 38 PM

I see the results for the new search, but I see the constraint for the previous advanced search rather than my current search constraint:
Screenshot 2025-07-18 at 2 20 52 PM

taylor-steve:
Noting that we think there might be a bug in advanced search, we're going to wait until we address that to come back to this

cbeer:
I think that's as expected? We're not showing a pill for the query term, and keyword searches are "search within"?

taylor-steve:
@cbeer I'm not sure I'm following. What is the expected behavior if a user uses the masthead to search after they do an advanced search?

hudajkhan:
@taylor-steve @cbeer I updated the method that handled advanced clause constraints to also handle regular queries (which is what the super class did as well).  In Steve's example, it should still show "Mark" along with the advanced clause constraint

taylor-steve (approved):

PR #5782, opened 2025-07-23
fix tab zeroing out on click

dnoneill:
There is a weird turbo load bar even after the articles+ tab has loaded. This fixes that.


jcoyne:
@cbeer I think your analysis is correct. It's showing the cached page while the spinner is running, and after 4s or so, it finishes the request.  This is the default Turbo behavior.

PR #5784, opened 2025-07-23
Extract articles+ full text to a separate pop-up

### LINKED ISSUE #5776: Full text (HTML) when available in Artcles+
### Context

Some records have HTML full text -- [here's an example](https://searchworks-uat.stanford.edu/articles/aph__185480548). 

As is, it's not accessible because some text should be headings -- see screenshot below. 

Image

<br><br>
In production, we seem to strip out headings using an [EDS gem](https://github.com/ebsco/edsapi-ruby/blob/810cdd97452422b23f9b9c3416a6f10997cd203a/lib/ebsco/eds/record.rb#L323), which isn't a good solution either. (Also notice the treatment of tables b/n UAT and production.) Images seems to be stripped out in both UAT and production. 

This ticket outlines a way to render HTML full text in a separate tab, but noting that it needs to be accessible. If tables can be addressed by rendering them in an accessible way, that's even better.

### Design with specifications

- [New tab/window with full text](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=3786-161714&t=F68jsLDmanym8YhX-4) 
- [Record view where HTML full text is available](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=3420-85778&t=F68jsLDmanym8YhX-4)

### Design notes

- Title of the article will be an H1 and then headings will follow accordingly after that. In this example:
  - "1. Introduction" is an H2 
  - "2. Materials and Methods" is an H2
  - "2.1. Initial Case Summary and Sample Collection" is an H3

cbeer:
... and do a better job mapping their weird markup to HTML.

Fixes #5776

### DIFF
--- a/app/models/eds_document.rb
+++ b/app/models/eds_document.rb
@@ -67,7 +67,7 @@ class EdsDocument # rubocop:disable Metrics/ClassLength
   def eds_html_fulltext
     return unless eds_html_fulltext_available?
 
-    dig('FullText', 'Text', 'Value')
+    CGI.unescapeHTML(dig('FullText', 'Text', 'Value'))
   end
 
   def eds_html_fulltext_available?
--- a/app/models/eds_document.rb
+++ b/app/models/eds_document.rb
@@ -65,7 +65,7 @@ class EdsDocument # rubocop:disable Metrics/ClassLength
   end
 
   def eds_html_fulltext
-    return unless eds_html_fulltext_available?
+    return unless eds_html_fulltext_available? && dig('FullText', 'Text', 'Value').present?
 
     CGI.unescapeHTML(dig('FullText', 'Text', 'Value'))
   end
--- a/app/assets/stylesheets/searchworks4/record.css
+++ b/app/assets/stylesheets/searchworks4/record.css
@@ -383,3 +383,10 @@
 .bound-with .show-more-button {
   margin-bottom: 0.1rem;
 }
+
+.blacklight-articles-fulltext_html {
+  #topnav .navbar-nav { display: none; }
+  .access-panel-heading {
+    display: none;
+  }
+}
--- a/app/components/articles/article_component.html.erb
+++ b/app/components/articles/article_component.html.erb
@@ -2,28 +2,13 @@
   <div class="article-record-sections record-sections">
     <% if @document.html_fulltext? %>
       <div class="html-fulltext-container mb-5">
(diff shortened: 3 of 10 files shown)


app/models/eds_document.rb
```diff
@@ -67,7 +67,7 @@ def eds_fulltext_links
   def eds_html_fulltext
     return unless eds_html_fulltext_available?
 
-    dig('FullText', 'Text', 'Value')
+    CGI.unescapeHTML(dig('FullText', 'Text', 'Value'))
```
taylor-steve, line 70: For `/articles/aph__185480548/fulltext` locally, unauthenticated I'm seeing:
* `eds_html_fulltext_available?` is true
* `dig('FullText', 'Text', 'Value')` is nil, which `CGI.unescapeHTML` is unhappy about

taylor-steve: I'm not sure if we need that, I'm confused what's going on. That item has full text in prod, but it seems like it isn't coming across.

Here's the `dig('FullText')` 

```ruby
{"Links" => [{"Type" => "pdflink"}], "Text" => {"Availability" => "1"}, "CustomLinks" => [{"Url" => "http://library.stanford.edu/sfx?ID=doi:10.3390%2Fani15101416&genre=article&atitle=First Detection and Genetic Characterization of Felis catus Papillomavirus Type 11, the First Treisetapapillomavirus Type to Infect Domestic Cats.&title=Animals (2076-2615)&issn=20762615&isbn=&volume=15&issue=10&date=20250515&aulast=Munday, John S.&spage=1416&pages=1416-1430&sid=EBSCO:Academic%20Search%20Premier:185480548", "Name" => "SFX Link Resolver (all content) s4392798", "Category" => "fullText", "Text" => "Check SFX for full text", "Icon" => "http://library.stanford.edu/sfx/img/sfxmenu/SFXFinal.gif", "MouseOverText" => "Check SFX for full text availability"}]}
```

I'm comparing to the EDS gem but haven't figured out the missing piece yet.

cbeer: It's confusing, but I dug into it more this morning. It seems like `eds_html_fulltext_available?` is whether there is any at all, but if you're not authenticated there won't actually be any.

taylor-steve (approved):

PR #5798, opened 2025-07-24
remove padding on flash messages if there are no flash messages

dnoneill:
Before:
Screenshot 2025-07-24 at 12 30 15 PM

After:
Screenshot 2025-07-24 at 12 30 04 PM

### DIFF
--- a/app/components/main_container_component.html.erb
+++ b/app/components/main_container_component.html.erb
@@ -3,7 +3,7 @@
   <main id="main-container" data-controller="google-cover-image" class="mb-5">
 
     <% # The quick report javascript requires this node to be on the page %>
-    <div class="flash_messages mx-auto my-3" style="max-width: 1000px;">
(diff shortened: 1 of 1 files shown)


app/components/main_container_component.html.erb
```diff
@@ -3,7 +3,7 @@
   <main id="main-container" data-controller="google-cover-image" class="mb-5">
 
     <% # The quick report javascript requires this node to be on the page %>
-    <div class="flash_messages mx-auto my-3" style="max-width: 1000px;">
+    <div class="flash_messages<% if flash.any? %> mx-auto my-3<% end %>" style="max-width: 1000px;">
```
cbeer, line 6: Is there any reason to render the container at all?

Or, maybe the actual `AlertComponent` brings its own vertical margins, so they don't stack awkwardly?

dnoneill: @cbeer According to the comment we are supposed to for the quick report javascript request the node to be on the page

PR #5799, opened 2025-07-24
Advanced facet 5321

hudajkhan:


app/components/searchworks4/facet_search_component.html.erb
```diff
@@ -3,9 +3,12 @@
...
-    <%= render(Blacklight::Facets::InclusiveConstraintComponent.new(facet_field: @facet_field)) %>
-    <%= content_tag :ul, class: 'facet-values list-unstyled', role: @role do %>
-      <%= render facet_items %>
+    <% if inclusive_present? %>
+      <%= render(Searchworks4::InclusiveConstraintComponent.new(facet_field: @facet_field)) %>
+    <% else %>
+      <%= content_tag :ul, class: 'facet-values list-unstyled', role: @role do %>
+        <%= render facet_items %>
+      <% end %>
     <% end %>
```
taylor-steve, line 12: Could you help me understand this change? It's Friday afternoon, forgive me if this is obvious and I'm missing it. I'm confused about these conditionals. I would expect `Searchworks4::InclusiveConstraintComponent` to share the same interface/contract as `Blacklight::Facets::InclusiveConstraintComponent` and handle this decision in `render?`. Is there something available to `inclusive_present?` that `Searchworks4::InclusiveConstraintComponent` doesn't have access to? Thanks.

taylor-steve: @hudajkhan explained that we (in SearchWorks) want to avoid rendering this in the inclusive case:

```ruby
<%= content_tag :ul, class: 'facet-values list-unstyled', role: @role do %>
  <%= render facet_items %>
<% end %>
```

So we might be able to push the component's part of that conditional to `render?`, but ultimately we will still need a conditional in the template because we are deviating from the behavior of the Blacklight template.

app/components/searchworks4/inclusive_selected_qfacet_value_component.rb
```diff
@@ -0,0 +1,23 @@
...
+    def initialize(values:, field_name:, item:)
+      @values = values
+      super(field_name:, item:)
+    end
+
+    def remove_href
+      # Generate the new values for the remove link without the value for this item
+      # @values is a nested array e.g. [['A', 'B']], but to add values back
+      # we need a regular array, hence the flattening.
+      new_values = @values.map { |v| v - [item.qvalue] }.flatten
```
taylor-steve, line 14: For a case like: http://localhost:3000/?f%5Bformat_hsim%5D%5B%5D=Software%2FMultimedia&f_inclusive%5Bformat_hsim%5D%5B%5D=Archive%2FManuscript&f_inclusive%5Bformat_hsim%5D%5B%5D=Book

Which is
`{"f" => {"format_hsim" => ["Software/Multimedia"]}, "f_inclusive" => {"format_hsim" => ["Archive/Manuscript", "Book"]}}`

This is erroring with `undefined method '-' for an instance of String`

I'm not sure how to handle these mixed scenarios 🤔

hudajkhan: This situation resulted from the addition links not using the inclusive facet but the regular facet. These should now be changed so we won't end up in this situation again.

taylor-steve (approved):

PR #5809, opened 2025-07-24
uniq authors on eds header for modal

dnoneill:
Before:
Screenshot 2025-07-24 at 4 05 16 PM

After:
Screenshot 2025-07-24 at 4 05 03 PM

### DIFF
--- a/app/components/searchworks4/record_summary_component.rb
+++ b/app/components/searchworks4/record_summary_component.rb
@@ -10,7 +10,7 @@ module Searchworks4
     end
 
     def authors
-      return presenter.document.eds_authors if presenter.document.eds?
+      return presenter.document.eds_authors&.uniq if presenter.document.eds?
 
       sum = []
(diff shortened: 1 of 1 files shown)


app/components/searchworks4/record_summary_component.rb
```diff
@@ -10,7 +10,7 @@ def initialize(presenter:)
     end
 
     def authors
-      return presenter.document.eds_authors if presenter.document.eds?
+      return presenter.document.eds_authors.uniq if presenter.document.eds?
```
cbeer, line 13:
```suggestion
      return presenter.document.eds_authors&.uniq if presenter.document.eds?
```

It looks like eds_authors can be nil?
https://github.com/sul-dlss/SearchWorks/blob/e360df830f40ef289e96870442264c95b0670971/app/models/eds_document.rb#L138

PR #5831, opened 2025-07-25
Use friendly library names in page title

### LINKED ISSUE #5825: Translate library codes in the page title
We're showing the library code instead of the human-readable library name in the page title:

E.g. https://searchworks.stanford.edu/?f%5Blibrary%5D%5B%5D=HILA has the title:

Image

marlo-longley:
Fixes #5825

### DIFF
--- a/app/helpers/catalog_helper.rb
+++ b/app/helpers/catalog_helper.rb
@@ -10,6 +10,16 @@ module CatalogHelper
     document_index_view_type.to_s
   end
 
+  def format_constraints_page_title(search_state)
+    text = render_search_to_page_title(search_state)
+    return text unless text.present? && search_state.filter('library').present?
+
+    library_code = search_state.filter('library').values.first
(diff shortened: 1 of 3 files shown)


cbeer:
Is there no hope doing this upstream/when we're initially constructing the text?

marlo-longley:
https://github.com/projectblacklight/blacklight/blob/d241cbbd27d0b223b48285e9bb229886bffb6a5a/app/helpers/blacklight/catalog_helper_behavior.rb#L131 This is my read on where the text is being constructed.

Since BL doesn't know about our Stanford library codes I am not sure.

dnoneill:
So blacklight doesn't ignore helper methods. I think it might be a bug in one of our overrides that is causing the problem. I can't quick figure out the problem though. Might be something for the next wc

marlo-longley:
Thanks for looking into this Niqui -- happy to discuss more

PR #5835, opened 2025-07-25
Remove .no-js class

cbeer:
We're not un-setting it, and the only style rule I saw doesn't seem like it's applicable to us.

### DIFF
--- a/app/javascript/controllers/record_print_controller.js
+++ b/app/javascript/controllers/record_print_controller.js
@@ -1,0 +1,14 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="record-print"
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

app/javascript/controllers/record_print_controller.js
```diff
@@ -0,0 +1,14 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="record-print"
+export default class extends Controller {
```
jcoyne, line 4: Is this supposed to be part of this PR?

jcoyne (changes requested):

PR #5873, opened 2025-07-29
Remove extraneous line wrapping that splits HTML unnecessarily

jcoyne:


cbeer:
I'm so glad to see this. I spent all work-cycle wondering if this was the new, cool style of how to write HTML (and I hate it.).

PR #5876, opened 2025-07-29
Use .icon-link to provide the appropriate padding + alignment for links with icons

cbeer:

### DIFF
--- a/app/assets/stylesheets/searchworks4/bookmarks.css
+++ b/app/assets/stylesheets/searchworks4/bookmarks.css
@@ -32,11 +32,6 @@
         overflow: visible;
         fill: var(--bl-icon-color);
         vertical-align: sub;
-
-
-        &.bookmark-checked {
-          display: none;
-        }
       }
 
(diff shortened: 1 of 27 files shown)


cbeer:
@taylor-steve : I think I found it.  `display: block` and `display: inline` produce different padding. I rewrote the bookmark icon handling in f69c037 to clarify what's going on and make sure we're doing things consistently 🤷‍♂️

taylor-steve:
@cbeer yes looks like you've solved that. One or two more things to maybe check if you are seeing as well:

Looks like the show details chevron has drifted
Screenshot 2025-08-05 at 3 51 28 PM

This could be an asset issue on my end, but on the Digital collections featured resource page (/catalog?f%5Bcollection_type%5D%5B%5D=Digital+Collection) there's a real big arrow now:
Screenshot 2025-08-05 at 3 54 58 PM

cbeer:
Thanks. How about with 54630b0?

PR #5881, opened 2025-07-29
Track stack map text directions usage

taylor-steve:
We were tracking this prior to SW 4.0. Now that we are using Bootstrap tabs, we can track usage via those tab events.

Event name with dimensions:
Screenshot 2025-07-29 at 11 02 57 AM


app/components/stackmap/map_component.html.erb
```diff
@@ -1,13 +1,13 @@
-<div data-controller="stackmap" data-stackmap-api-url-value="<%= api_url %>" data-action="shown.bs.tab@window->stackmap#tabVisibility">
+<div data-controller="analytics stackmap" data-stackmap-api-url-value="<%= api_url %>" data-action="shown.bs.tab@window->stackmap#tabVisibility" data-analytics-category-value="stackmap">
   <p>Call number <span data-stackmap-target="callnumber" class="fw-bold"></span> is located in <span data-stackmap-target="library" class="fw-bold"></span>, <span data-stackmap-target="floorname" class="fw-bold"></span>.</p>
   <ul class="nav nav-tabs" role="tablist">
-    <li class="nav-item" role="presentation">
+    <li class="nav-item" role="presentation" data-action="shown.bs.tab->analytics#trackEvent">
```
jcoyne, line 4: Why are you binding to the bs event rather than just the the button action?

taylor-steve: Conceptually I want to track the event in which "the text instructions have been seen", which made me go down this path. I think either road leads to Rome in this case.

jcoyne: This way just seems to be more indirect and susceptible to break in a way we don't detect when bootstrap decides to change it's inner workings.

PR #5893, opened 2025-07-31
Use imageTargetConnected to batch cover image fetching instead of doi…

cbeer:
…ng it at connect and requiring async javascript to initialize the controller again.


jcoyne (approved):
Nice 👍

PR #5905, opened 2025-08-01
Consolidate citation format approaches

cbeer:


app/components/searchworks4/citations/citation_component.html.erb
```diff
@@ -1,30 +1,23 @@
...
-          <% end %>
-        </select>
-      </div>
+  <div class="row mb-3">
+    <div class="col-auto">
+      <label for="citation-format-select" class="col-form-label fw-bold">Select format</label>
     </div>
-  <% end %>
+    <div class="col-auto">
+      <select id="citation-format-select" data-action="change->citation-format#reveal" class="form-select" style="width: unset;" aria-controls="<%= citations.keys.map { |style| "citation-format-#{style}" }.join(' ') %>">
```
taylor-steve, line 11: Okay so you're getting bit by the lack of tests here (which I contributed to...). Designs have no select for preferred only. Maybe we don't show the format select if there's only the default?
Screenshot 2025-08-01 at 10 07 36 AM

cbeer: Thanks. I guess that's what the hidden field conditions were also doing.

PR #5908, opened 2025-08-01
Let the turbostream update the feedback form state.

cbeer:


marlo-longley:
Tested works well; needs a rebase.

PR #5950, opened 2025-08-06
Add the additional articles parameters everywhere but the search page.

### LINKED ISSUE #5949: Full text default not sticking in Articles+ when user starts a search from a record view page
Irina reported on 7/30 that we are losing the "Direct access to full text" search default when the user is viewing a record view page, and then starts a new search. When the results list displays,  the "Direct access to full text" default is not applied.

Image

cbeer:
Fixes #5949 🤞

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -24,6 +24,15 @@ class ArticlesController < ApplicationController
     raise exception
   end
 
+  # manipulate the search state params to set 'Direct access to full text' as the default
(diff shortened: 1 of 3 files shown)


taylor-steve:
This didn't change the behavior locally or on stage for me for what I understand to be the issue?

* Search `dogs` in Articles+
* Click first result
* Search `cats` from the navbar search

Is that right?

Also current behavior matches old prod behavior?

taylor-steve (approved):

PR #5996, opened 2025-08-12
Use json format in mini-bento URLs

taylor-steve:
`headers: { 'accept': 'application/json' }` doesn't appear to trigger `controller.params[:format] == 'json'` for the remote apps. Doing it this way does, letting non-VPN users get mini-bento results for our apps that use our current bot challenge gem config.

Better ideas? 🤷‍♂️

### DIFF
--- a/app/components/search_result/mini_bento/layout_component.html.erb
+++ b/app/components/search_result/mini_bento/layout_component.html.erb
@@ -53,14 +53,14 @@
           </div>
         </div>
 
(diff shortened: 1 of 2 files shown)


jcoyne:
Instead of looking at `params`, why not look at `request.format.json?`?

jcoyne:
🤦 Thanks Steve. My misunderstanding.

taylor-steve:
That did the trick, thank you both.

PR #6026, opened 2025-08-20
Stop using PathHierarchyTokenizer for call numbers at query time

taylor-steve:
This splits the analyzers into index and query. It performs the same normalization for both, but treats the user query as a single term rather than exploding it into all the path components.


cbeer:
🤞

Issue #5627, opened 2025-07-15
Focus rings + keyboard navigation


dbranchini:
### From Darcy's review on 8/20/25
- [x] Skip to - Skip to search doesn't work (on home page and other pages) -- not sure where it's landing, but I can't seem to interact with it and there's no focus ring. Question though -- should it land on the input or the catalog/articles switch or the all fields dropdown? I lean toward input as a starting point.

- [ ] Header - Stanford University (a tiny bit more space around it)
- [ ] Header - No focus ring for Stanford University Libraries (focus happens, but you can't see it)
- [ ] Header/Search - Search mode - lands on first element in radio button group and allows you to move to the other option, but once pressed, you move to another page and lose the focus on the radio button (probably because it's different page, but wondering if we can make that experience more seamless)

- [x] Home - Featured resources -- focus ring should hug heading (instead of expanding to full width of heading)

- [ ] Search results - show/hide details -- focus ring should hug heading (same as above)

- [x] Search results - mini-bento - side of focus ring gets cut off by count (number of matches)
- [ ] Search results - search for a facet value - (minor) can the whole input (with the search icon) be part of the focus ring -- see screenshot for how it looks today.
  
<img width="396" height="130" alt="Image" src="https://github.com/user-attachments/assets/c81815fe-bd52-403b-b143-7a758a5115be" />
  
- [ ] Record view - tools (save, cite, email, etc.) - same focus ring (preferably the one used for cite, email, etc. is applied to save)
- [ ] Record view - Related items -- Needs discussion. Currently, it allows me to enter the area, scroll and select an item, but I only see a focus ring around the title. Can it also show the hover effect for the item too? Also, when I enter the scroll-able area for related items, could it scroll down to that place on the page? I entered it with it slightly hanging off the page. Also, should the user need to enter it with return or space? Similar to an accordion. Currently, you enter it with a tab, but you kind of end up trapped there. 
- [ ] Record view - Availability Expand - top of focus ring is cut-off

- [ ] Footer - No focus ring on Stanford University Libraries
- [ ] Footer - (minor) match the focus ring style for the SUL footer, it has a tiny bit more padding, the Stanford footer links don't

### Important 
[Defined focus ring style in Figma](https://www.figma.com/design/K1syTsalaB1nlRftgOi0Pw/DLSS-Components?node-id=802-1231&t=J5xViUNgKTk6ZGhF-4)

[...]

PR #6109, opened 2025-09-05
Prevent link focus when long text is truncated

taylor-steve:
Part of #5627 

Avoids this situation:
Image


cbeer:
Hm. I think this might not work right in all cases. I thought the CSS line truncation didn't hide text from screen reader users (intentionally.. they have better controls for skipping text), so I think this will end up making those links inaccessible ?

Issue #6102, opened 2025-09-02
Stacks maps -- no text directions


dbranchini:
Some stacks maps (Green reference) don't have text directions available. In that case, display copy that says, "No text directions available for this location."

Example - https://searchworks.stanford.edu/view/4520586

cc: @saseestone (I believe Sarah was going to investigate why we don't have text directions and if this can be remedied in stacks maps.)

dnoneill:
Can this be closed? Seems like https://github.com/sul-dlss/SearchWorks/pull/6140 closed this.

dnoneill:
Gotcha just double checking there isn't anything on the dev side that needs to be done.

saseestone:
Nope it's all me :)

PR #6140, opened 2025-09-12
Hide the stackmap directions tab if there are no directions

taylor-steve:
Ref #6102 

Screenshot 2025-09-12 at 3 18 25 PM

### DIFF
--- a/app/javascript/controllers/stackmap_controller.js
+++ b/app/javascript/controllers/stackmap_controller.js
@@ -7,7 +7,7 @@ export default class extends Controller {
     apiUrl: String
   }
 
-  static targets = ["map", "directions", "callnumber", "library", "floorname", 'fitButton', 'zoomOutButton']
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/stackmap_controller.js
```diff
@@ -42,8 +42,13 @@ export default class extends Controller {
 
       this.libraryTarget.textContent = map.library
       this.floornameTarget.textContent = map.floorname
-      // Replace <li>- with <li>
-      this.directionsTarget.innerHTML = map.directions.replaceAll(/<li>-\s*/g, '<li>')
+
+      if (map.directions == "") {
+        this.directionsTabTarget.classList.add('d-none')
+      } else {
+        // Replace <li>- with <li>
```
cbeer, line 49:
```suggestion
        // Replace <li>- with <li>
        this.directionsTabTarget.classList.remove('d-none')
```

I think I've convinced myself we never reuse the rendered partial in any case where we'd have to undo this, right?

PR #6154, opened 2025-09-18
Analytics

### LINKED ISSUE #4737: Update display of Analytics with no relationship in Folio
We have display issues on Analytics records with no Bound with/Analytics relationship in FOLIO. We've heard from Metadata that it will likely take years for them to get the records to be linked properly. We should make some updates to the display to help the user while we wait for better Folio data.

I will update the Folio discovery display names for "Shelved Elsewhere" locations. Darcy and I are thinking of using   "See individual volumes below".  But that text may need to change based on what we implement. 

We are also recommending we:
- [ ] Link the location display name to Browse related items. We need to ensure we're linking to the correct call number button). 
- [ ] Suppress the "Unknown" item status

Something like this:

Image

**Analytics data is messy,** in large part because it is complicated and it didn't migrate well from Symphony. We're trying to make some small changes, on records that are mostly "good",  to help users. This is not a fix that will help all the time, but the hope is that it won't make things worse.

Here are example records identified as "good" by our Metadata colleagues:
- Merged holdings: https://searchworks.stanford.edu/view/2027166 
- Unmerged holdings: https://searchworks.stanford.edu/view/12417107 
- SAL3 location other than stacks: https://searchworks.stanford.edu/view/1676599 
- Linked bound with on partially analyzed: https://searchworks.stanford.edu/view/492755 
- Two different call numbers: https://searchworks.stanford.edu/view/1121234 

**We're striving to make these "Good" records more user-friendly.**

"Bad" examples include:
- Does not display latest received (data not in FOLIO): https://searchworks.stanford.edu/view/12100343 
- Partially analyzed: https://searchworks.stanford.edu/view/351255 
- V.6 transferred: https://searchworks.stanford.edu/view/9629981 
- Request button on Media: https://searchworks.stanford.edu/view/463434 
- Lots of breaks in volumes: https://searchworks.stanford.edu/view/351225 
- Placeholder on partial analytic: https://searchworks.stanford.edu/view/357638 
- Partially analyzed & serial bound with: https://searchworks.stanford.edu/view/481487 
- Unlinked bound with: https://searchworks.stanford.edu/view/498445 
- Holdings with no item: https://searchworks.stanford.edu/view/5570461 
- Journal with only a few analyzed: https://searchworks-morison.stanford.edu/view/361434 
- Disc numbers do not match accession numbers: https://searchworks.stanford.edu/view/6316702 
- Shelved by title: https://searchworks.stanford.edu/view/3941412 
- (Hopkins item should not be listed since classed separately)
- 2 different SAL3 locations and only one shelved elsewhere item: https://searchworks.stanford.edu/view/355842 

**We don't want to make these "Bad" examples any worse than they are now.**

hudajkhan:
Closes #4737 

What this PR does: 
- If the location code for an item is SAL3-SEE-OTHER, displays a "see volume" button in the availability panel on the right side of the screen and in the availability modal. The button opens up a volume modal.
- Sets up code in catalog controller that does a search for the call number, sorted by title, and then displays the results in the modal.  The "volumes_modal" view uses the volumes modal component to display the content.
- If there are more than 4 items returned, the volumes modal will show a search box that allows users to type in any portion of the title and callnumber of interest and the matching values are bolded in the modal.
- Note: Currently, the label shown above the see volumes button shows the location name for "SAL3-SEE-OTHER". @saseestone has updated the name in Folio to match what we want to display instead i.e . "Individual volumes catalogued separately" and those changes will be reflected next week probably in our systems.
Screenshots

Availability section of the page shows the "see volume" button 
Screenshot 2025-09-18 at 5 16 32 PM

Volumes modal

Screenshot 2025-09-18 at 5 16 41 PM

Volumes modal showing content being bolded based on what the user types in the search box
Screenshot 2025-09-18 at 5 16 47 PM

See volumes button within the availability modal
Screenshot 2025-09-18 at 5 17 00 PM

### DIFF
--- a/spec/components/searchworks4/availability_component_spec.rb
+++ b/spec/components/searchworks4/availability_component_spec.rb
@@ -171,7 +171,7 @@ RSpec.describe Searchworks4::AvailabilityComponent, type: :component do
 
     it 'displays the link for the volumes modal' do
       render_inline(described_class.new(document: document))
-      
+
       expect(page).to have_css('a[data-blacklight-modal="trigger"]', text: 'See volumes', visible: :all)
     end
   end


app/components/searchworks4/volumes_modal_component.rb
```diff
@@ -0,0 +1,30 @@
...
+      @response.dig(:response, :docs)
+    end
+
+    def call_number_display(solr_doc)
+      if solr_doc.key?('lc_assigned_callnum_ssim')
+        solr_doc['lc_assigned_callnum_ssim'].join(', ')
+      elsif solr_doc.key?('item_display_struct')
+        solr_doc['item_display_struct'][0]['callnumber']
+      else
+        ''
```
cbeer, line 22: I don't think this is the call number logic we want to use. Maybe there's existing code for browse nearby that does it correctly? (At the very least, using the LC assigned call number by default is giving us bad call number data when our local call numbers don't match (e.g. we're using Dewey or something)

hudajkhan: I can look into browse nearby. I was quite unsure about how call numbers are supposed to work here so was hoping for some feedback on this part.

Issue #13, opened 2014-05-02
Access Point Specific Facet Headings


jkeck:
We need to have the facility for access point specific facet headings (ideally using i18n).

The databases page will be a good start as that data exists in our index.

![access-point-specific-heading](https://cloud.githubusercontent.com/assets/96776/2867527/92555326-d237-11e3-9548-71c128b6f128.png)

PR #6156, opened 2025-09-19
handle ASCII characters in abstract

### LINKED ISSUE #6153: line break and whitespace character encoding visible in Abstracts
Reported via SW feedback (and passed around the PURL feedback too). [SW-4547](https://jirasul.stanford.edu/jira/browse/SW-4547)
(Note that Jeanette instructed Adan to fix the Abstract in Argo, and he has. Now it displays as a blob of text.)

In the abstract of SDR records, we're displaying the encoding for line breaks and whitespace characters in 4.0. I checked -morison, and we had a better display in 3.3.  

**Example record w/screenshot:** 
https://searchworks.stanford.edu/view/zs304tj5371 

Image

vs
https://searchworks-morison.stanford.edu/view/zs304tj5371 (same record in SW3.3)

Image

**Another example:**
https://searchworks.stanford.edu/view/cz789ms7413 
https://searchworks-morison.stanford.edu/view/cz789ms7413 (same record in SW3.3)

This seems to be happening when the depositor has copy and pasted from a PDF into the Abstract field. "Real" line breaks are displaying fine, but we're seeing encoding if there is markup that came along in that copy/paste action.

SDR folks will try to educate depositors that they should be careful of copy/paste, but it's likely it will continue to happen.  I'm hopeful that we have code already 🤞 to remove the encodings from displaying.

hudajkhan:
Closes #6153 .

What this PR does:
Unescapes ASCII characters like $#13; (which is carriage return). 
The mods display gem has additional handling around dealing with HTML encoding, and the unescape method is called before the text is passed on to the Mods Display component(see https://github.com/sul-dlss/mods_display/blob/main/app/helpers/mods_display/record_helper.rb#L103 ).

### DIFF
--- a/app/components/record/mods_document_component.html.erb
+++ b/app/components/record/mods_document_component.html.erb
@@ -90,7 +90,7 @@
     <%= render Searchworks4::DocumentSectionLayout.new(title: 'Abstract/Contents', dl_classes: 'dl-horizontal dl-text') do %>
       <% document.mods.abstract.each do |abstract| %>
         <div class="my-2">
-          <%= truncated_mods_record_field(abstract, value_transformer: ->(v) { tag.div v, data: { 'long-text-target' => 'text' } } ) %>
+          <%= truncated_mods_record_field(abstract, value_transformer: ->(v) { tag.div CGI.unescapeHTML(v), data: { 'long-text-target' => 'text' } } ) %>
         </div>
       <% end %>
 
--- a/spec/components/record/mods_document_component_spec.rb
+++ b/spec/components/record/mods_document_component_spec.rb
@@ -136,4 +136,13 @@ RSpec.describe Record::ModsDocumentComponent, type: :component do
       expect(page).to have_css("dd", text: "amazing")
     end
   end
+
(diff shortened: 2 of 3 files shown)


app/components/record/mods_document_component.html.erb
```diff
@@ -90,7 +90,7 @@
     <%= render Searchworks4::DocumentSectionLayout.new(title: 'Abstract/Contents', dl_classes: 'dl-horizontal dl-text') do %>
       <% document.mods.abstract.each do |abstract| %>
         <div class="my-2">
-          <%= truncated_mods_record_field(abstract, value_transformer: ->(v) { tag.div v, data: { 'long-text-target' => 'text' } } ) %>
+          <%= truncated_mods_record_field(abstract, value_transformer: ->(v) { tag.div CGI.unescapeHTML(v), data: { 'long-text-target' => 'text' } } ) %>
```
dnoneill: Is there an instance where the abstract is nil? This is my one concern.

hudajkhan: I can look at that function. Thanks!

Issue #5952, opened 2025-08-06
Cleanup - Availability sidebar on record view


dbranchini:
- [x] Expand should be red - [see designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=950-58828&t=XJ37KtF1P4xue7h4-4)
- ~~[ ] Analyzed serials~~ MOVING to a separate ticket
- [ ] Variations on Location, Summary of items (holdings statement) and request button (location-level) - [see designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-?node-id=950-59213&t=uzvtDhdi80UMerR9-4)
  - Location + Summary of items
  - Location + Summary of items + Request
  - Location + Request
  - NOTE: Variations on the same sidebar for short and long location names as well as simply Request button and Request via Finding Aid button
- [x] Truncate access conditions/use and reproduction on SDR records to 5 or 6 lines with a show more/less behavior - [see Figma design](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=2181-162834&t=XJ37KtF1P4xue7h4-4) and [examples where it's needed](https://searchworks.stanford.edu/?search_field=search&q=%22README-COPYRIGHT+file%22). 
- [x] Making "Report a connection problem" more prominent - [see Figma design](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=4125-51284&t=XJ37KtF1P4xue7h4-4)

Issue #6136, opened 2025-09-12
Update how links are displayed in Articles+


dbranchini:
### Context

We received feedback on links displayed in Articles+. It's logged in the [SW 4.0 Gryphon Search feedback spreadsheet](https://docs.google.com/spreadsheets/d/1Oucn2XTbaDb84mrc6XPQxa4dN_kvNJtlOEDxtGtrwrU/edit?gid=0#gid=0). This addresses that feedback as well as some issues reported via Slack and/or seen by our development team.

### Issues and/or Designs

**Record view**

- [x] Change the label from "Best source" to "Online." Online should be displayed in green as it is in Catalog.
- [x] List should have the same spacing and lines as online links are displayed in Catalog.
- [x] PDF icon is [filetype-pdf](https://icons.getbootstrap.com/icons/filetype-pdf/) and should be displayed in Digital blue (#006CB8) as part of the link. 

-- See [Articles+ designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=4703-55320&t=dx8NHHr7TQH5ffRI-4) and [Catalog designs](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=950-59886&t=dx8NHHr7TQH5ffRI-4) 

**Results view**

- [x] List will appear with one "Available online" label and then a list of possible links. 
 
-- See [Catalog design](https://www.figma.com/design/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?node-id=3598-160623&t=dx8NHHr7TQH5ffRI-4) for reference.

**Other features/issues to address**

- [x] Remove PDF icon from items that are not a PDF ([ePub example](https://searchworks.stanford.edu/articles/nlebk__1335757))

PR #6171, opened 2025-09-24
updating and cleaning up articles plus display

hudajkhan:
Addresses #5952 

What this PR does:

Address many of the cleanup issues listed in the PR.

- Update the PDF icon and only show for PDF items

Screenshot example: new PDF icon which is linked
Screenshot 2025-09-24 at 3 32 35 PM

Screenshot example: non-PDF item does not get PDF icon
Screenshot 2025-09-24 at 3 32 48 PM

- Use the text 'online' instead of 'best source'. Also have the styling on the article page for availability match the catalog. 

Screenshot: showing one item with 'best source'
Screenshot 2025-09-24 at 3 34 42 PM

- Having multiple items under availability show "availability" only once and then list the items (hard to get a screenshot after not showing 'view content' links if other links are available)

- Not showing 'view content' link if other full text links are visible
Screenshot 2025-09-24 at 3 36 32 PM

### DIFF
--- a/app/components/articles/link_component.rb
+++ b/app/components/articles/link_component.rb
@@ -13,7 +13,7 @@ module Articles
     end
 
     def pdf_icon
-      tag.i class:'bi bi-filetype-pdf link-icon', aria: {label: 'PDF'}
+      tag.i class: 'bi bi-filetype-pdf link-icon', aria: { label: 'PDF' }
     end
 
     def href


app/models/concerns/eds_links.rb
```diff
@@ -14,7 +14,26 @@ def eds_fulltext_links_as_searchworks_links
...
+    fulltext_text = fulltext_links.filter_map(&:link_text).uniq
+
+    # If there are multiple fulltext links, and one is 'View on content provider's site',
+    # we do NOT want to use the 'View on content provider's site' link.
+    # Settung the 'fulltext' prope
+    if fulltext_text.size > 1 && fulltext_text.include?('View on content provider\'s site')
+      mapped_links.each do |link|
+        link.fulltext = false if link.link_text == 'View on content provider\'s site'
+      end
+    end
```
taylor-steve, line 34: The inner `FulltextLink` class already seems responsible for the knowledge of the `'View on content provider\'s site'` string. It might be worth checking if we can keep that knowledge in once place, maybe via something like a `provider_link?` method or similar.

taylor-steve (commented):
I think this might be for https://github.com/sul-dlss/SearchWorks/issues/6136?

app/components/searchworks4/link_component.html.erb
```diff
@@ -1,5 +1,4 @@
-<%= icon %>
-<%= link_to link_text, href, **link_attr %>
+<%= link_to display_link, href, **link_attr %>
```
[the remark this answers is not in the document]
hudajkhan: I could perhaps change it to link_text_for_display. Agreed - this is the text and icon included in the link, not the link itself.

app/models/concerns/eds_links.rb
hudajkhan: I can look into that some more.  What makes this a little challenging is it is not enough to know the characteristics of one link.  We want to hide the 'view content' link only if there are other full text links that can be displayed and that are not 'view content'.

hudajkhan:
after conversation with @taylor-steve , splitting out the 'view content' logic into a separate branch

app/components/articles/link_component.rb
```diff
@@ -7,9 +7,13 @@ def render?
...
     def icon
-      return unless link
+      return unless link && %w[ebook-pdf pdflink pdf].include?(link.type)
 
-      image_tag(image_url('pdf-icon.svg'), height: '20px', alt: 'PDF') unless link.type == 'other'
+      pdf_icon
+    end
+
+    def pdf_icon
+      sanitize "<i class='bi bi-filetype-pdf link-icon' aria-label='PDF'></i>"
```
cbeer, line 16: Instead of sanitizing the HTML after the fact, could we build safe HTML? E.g. something like:

```suggestion
      tag.i class: 'bi bi-filetype-pdf link-icon', aria: { label: 'PDF' }
```

hudajkhan: Changed per suggestion, thanks

app/components/articles/link_component.rb
```diff
@@ -39,6 +43,10 @@ def link_attr
       { data: { turbo: false } }
     end
 
+    def link_text_for_display
+      icon.present? ? (sanitize "#{icon} #{link_text}") : link_text
```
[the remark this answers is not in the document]
hudajkhan: also changed, thanks!

PR #6193, opened 2025-09-29
Deprioritize the microform format when selecting an icon.

cbeer:


jcoyne (commented):
Can you explain why we want this change?

cbeer:
This restores the icon behavior to before we added the microform type.

taylor-steve (approved):

PR #6219, opened 2025-10-03
Avoid using deprecated solr_field: pub_date

jcoyne:

### DIFF
--- a/lib/holdings/spine.rb
+++ b/lib/holdings/spine.rb
@@ -21,12 +21,12 @@ class Holdings
 
     # create sorting key for spine
     # shelfkey asc, then by sorting title asc, then by pub date desc
-    # note: pub_date must be inverted for descending sort
+    # note: pub_year_ss must be inverted for descending sort
     def sort_key
-      sort_pub_date = if document[:pub_date].blank?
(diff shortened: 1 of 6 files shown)


lib/holdings/spine.rb
```diff
@@ -21,12 +21,12 @@ def item
 
     # create sorting key for spine
     # shelfkey asc, then by sorting title asc, then by pub date desc
-    # note: pub_date must be inverted for descending sort
+    # note: pub_year_ss must be inverted for descending sort
```
cbeer, line 24: I thought `pub_year_ss` possibly contains non-numeric data, like date ranges and stuff. I don't think it's a drop-in replacement?

jcoyne: `pub_date` contains non-numeric data.
Screenshot 2025-10-03 at 3 46 21 PM

https://sul-solr-prod-a.stanford.edu/solr/#/searchworks-prod/query?q=pub_year_tisim:*&q.op=OR&indent=true&fl=pub_date_search,%20pub_date,%20pub_year_isi,%20pub_year_tisim,%20pub_year_ss

PR #6220, opened 2025-10-03
Put all database names in config

jcoyne:
For consistency, only have a single way of showing the database name on the databases page


cbeer:
I think it's a feature that we prefer the MARC record title and let librarians maintain the data.

jcoyne:
Why not for Web of Science though?

PR #6230, opened 2025-10-07
update facet drawer height

### LINKED ISSUE #6229: Change mobile drawer height
Currently, we use drawers for filters on the search results screen for XS viewports. The drawer height for filters is only 20% ish, let's change it to 80% so that it's easier to use.

| **As designed** | **In production** |
| ------------- | ------------- |
| Image | Image |

[Interactive prototype](https://www.figma.com/proto/rphRL8X4aDjBf4OaOFZCYl/SearchWorks-4.0--WC2-3-?page-id=2%3A3884&node-id=2380-167011&viewport=-5516%2C-4049%2C0.29&t=HeRQD9k69pFw7NB8-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2380%3A167011&show-proto-sidebar=1
)

dnoneill:
closes #6229

Screenshot 2026-01-12 at 4 18 42 PM
Screenshot 2026-01-12 at 4 18 30 PM

### DIFF
--- a/app/assets/stylesheets/searchworks4/facets.css
+++ b/app/assets/stylesheets/searchworks4/facets.css
@@ -170,3 +170,7 @@ label[for="facet_option_stanford_only"] > a {
 .blacklight-catalog:not(:has(.constraints .constraint)) #facet-pub_year_tisim .chart-wrapper {
   display: none !important; /* important needed to override the inline style that chartjs inserts */
 }
+
+#offcanvas-facets {
+  --bs-offcanvas-height: 80%;
+}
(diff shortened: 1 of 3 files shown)


taylor-steve:
Is medium the full size side drawer? That seems fine to me but Darcy would the one to say for sure.

dnoneill:
@taylor-steve I went with a h1 because the top filters and other filters headings are h2 so it seemed weird to make it the same level heading when semantically it isn't.

dbranchini:
H2's is what we use on modals too, it should be an H2. (@alundgard and I discussed this a while back, it's a gray area for both drawers and modals, but we decided on H2s.)

PR #6359, opened 2025-12-12
Remove extraneous initializer

jcoyne:
The config is set up in config/initializers/config.rb


jcoyne:
Yes. I mixed up the two files.

PR #6387, opened 2026-01-05
Interactive elements, like checkbox, are not permitted inside button …

jcoyne:
…elements


taylor-steve:
It looks like the connection between the state of the save all control and the individual bookmarks was lost.

Screenshot 2026-01-05 at 8 57 00 AM

jcoyne:
@taylor-steve good catch. It was just a style issue, that I've resolved.

PR #6390, opened 2026-01-05
Avoid using CGI.parse

jcoyne:
It is going to be removed in Ruby 4.0

### DIFF
--- a/app/models/links/link.rb
+++ b/app/models/links/link.rb
@@ -88,8 +88,8 @@ class Links
       uri = URI.parse(Addressable::URI.encode(href.strip))
       host = uri.host
       if host =~ Links::PROXY_REGEX && uri.query
-        query = CGI.parse(uri.query)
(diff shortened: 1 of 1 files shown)


app/models/links/link.rb
```diff
@@ -88,8 +88,8 @@ def link_host
       uri = URI.parse(Addressable::URI.encode(href.strip))
       host = uri.host
       if host =~ Links::PROXY_REGEX && uri.query
-        query = CGI.parse(uri.query)
-        host = URI.parse(query['url'].first).host if query['url'].present?
+        query = Rack::Utils.parse_nested_query(uri.query)
+        host = URI.parse(query['url']).host if query['url']
```
dnoneill: I think present? is required. I think the url parameter can be = ''. I need to double check though.

dnoneill (approved):

PR #6494, opened 2026-02-19
Display cocina metadata on the librarian view

### LINKED ISSUE #6489: Switch librarian view to show the Cocina instead of MODS

jcoyne:
Fixes #6489


thatbudakguy (commented):
Tried this out and it looks good; I like the dynamic import strategy for highlighting so we can get rid of a dependency.

Do we want to separate out the removal of the MODS librarian view, so that it's not non-functional until we do a full reindex? Merging this will result in it being non-functional for nearly all SDR items until we do that.

PR #6633, opened 2026-05-11
Update dependencies

sul-devops-team:


app/controllers/concerns/all_caps_params.rb
```diff
@@ -14,7 +14,7 @@ module AllCapsParams
   def downcase_all_caps_params
     downcase_all_caps_param params, :q
 
-    params[:clause]&.each_value do |clause_hash|
+    params.expect(:clause)&.each_value do |clause_hash|
```
cbeer: I think this might be wrong.. `params.expect` raises an error if `clause` is missing (but it seems like we're probably ok with that?)

dnoneill: That was my assumption when I approved other requests. I wouldn't think we would want an non-erroring page when the param was missing but that might have been a wrong assumption.

cbeer: It's too early for double-negatives. I don't know if we agree that optional parameters shouldn't blow up, or if you're saying this parameter shouldn't be optional and should blow up the page?

taylor-steve: I'm inclined to disable the rule globally or at each site for now. Autocorrect over-applies for what we're trying to do in most places and I don't know if I want to think hard about each call this morning 😅

taylor-steve (approved):

PR #6707, opened 2026-07-08
Add error pattern for Cloudflare Turnstile

jcoyne:
This prevents a bunch of Cloudflare frontend errors (~450/hr), that are not actionable to us, from being reported. Presumably these are bots.
See https://app.honeybadger.io/projects/50022/faults/132138684


jcoyne:
This should already be prevented by `/TurnstileError/`

PR #6729, opened 2026-07-20
Guard preview controller against missing frame target

taylor-steve:
Another variation on #6704 

https://app.honeybadger.io/projects/50022/faults/122423098

### DIFF
--- a/app/javascript/controllers/preview_controller.js
+++ b/app/javascript/controllers/preview_controller.js
@@ -34,13 +34,15 @@ export default class extends Controller {
   }
 
   appendFrame() {
+    if (this.hasFrameTarget) return
+
     const frame = document.createElement("turbo-frame")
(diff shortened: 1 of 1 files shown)


jcoyne (changes requested):
Can you explain why this error might be happening?

app/javascript/controllers/preview_controller.js
```diff
@@ -34,13 +34,15 @@ export default class extends Controller {
...
     const frame = document.createElement("turbo-frame")
     frame.setAttribute("data-preview-target", "frame")
-    frame.setAttribute("data-preview-target", "frame")
     this.element.appendChild(frame)
   }
 
   reset() {
+    if (!this.hasFrameTarget) return
     this.frameTarget.innerHTML = ""
   }
```
jcoyne, line 47: can we put the appendFrame() call in reset?

PR #6767, opened 2026-08-12
Also apply ezproxy link text check for law, per SW-4689

cbeer:


app/models/links/ezproxy.rb
```diff
@@ -34,14 +34,15 @@ def ezproxy_host
...
-        Settings.libraries['LANE'].ezproxy_host
+        Settings.libraries.LANE.ezproxy_host
       elsif apply_sul_proxy_prefix?
         Settings.libraries.default.ezproxy_host
       end
     end
 
     def apply_law_proxy_prefix?
-      libraries.include?('LAW') &&
+      (link.link_text.downcase.include?('law library') || libraries.include?('LAW')) &&
```
jcoyne, line 44: Do we need `libraries.include?('LAW')` twice?  It must always be true due to the second instance, right?

jcoyne (changes requested):

app/models/links/ezproxy.rb
jcoyne: Rather, it doesn't seem that `link.link_text.downcase.include?('law library')` will have any effect.

app/models/links/ezproxy.rb
```diff
@@ -34,14 +34,15 @@ def ezproxy_host
...
+        libraries.include?('LAW') &&
```
cbeer:
```suggestion
```

jcoyne (approved):

PR #6769, opened 2026-08-12
Enable new rubocops

jcoyne:


dnoneill:
@jcoyne could you please put a one line comment why we are enabling these?

jcoyne:
@dnoneill we enable all of the rubocop rules.  These are just new ones that were released.

See https://github.com/sul-dlss/SearchWorks/pulls?q=is%3Apr+rubocop+is%3Aclosed

dnoneill:
I was confused because the numbers on these rules are 1.89 and the previous rubocops are 3.10. I did not realize until now rubocops are marked weird.

dnoneill (approved):

PR #6798, opened 2026-08-25
Fixed race condition in fetch error handler

jcoyne:
The problem was eager turbo frames were loading before the error handler were loaded. This resolves that order so that handlers are in place first.


jcoyne:
Superseded by #6823

PR #6812, opened 2026-09-02
Don't show lookbook in prod.

jcoyne:


cbeer:
... but I think we do want to show it in e.g. searchworks-stage.stanford.edu for e.g. Darcy to look at. I'm not sure using the Rails environment is the right key.

jcoyne:
Updated to use a feature flag.

PR #6815, opened 2026-09-02
Update solr configs for modern conventions.

jcoyne:
handleSelect is obsolete.

requestHandlers names should be urls


jcoyne:
This was part of #6814, and I was hoping we could do the two changes separately, but this shows the solr config and the code changes are dependent on one another.

PR #6829, opened 2026-09-04
Track mini-mini-bento navigation in Google Analytics

### LINKED ISSUE #6803: Add GA event to mini-mini-bento in Catalog and Articles
We want to better track if and how users move b/n Catalog and Articles+. Currently, there are three ways for users to do this -- the radio buttons in search, mini-bento and what we call mini-mini-bento. We already have an event on mini-bento and [we're adding an event to the radio buttons in search](https://github.com/sul-dlss/SearchWorks/pull/6800). This ticket addresses mini-mini-bento. 

Mini-mini-bento is seen in red on these screenshots:

Image
Image

taylor-steve:
Closes #6803

### DIFF
--- a/app/components/searchworks4/mini_mini_bento_component.html.erb
+++ b/app/components/searchworks4/mini_mini_bento_component.html.erb
@@ -1,6 +1,6 @@
-<div class="d-flex mt-2 mb-3 justify-content-center mini-mini-bento d-none" data-controller="mini-mini-bento">
+<div class="d-flex mt-2 mb-3 justify-content-center mini-mini-bento d-none" data-controller="mini-mini-bento analytics" data-analytics-category-value="mini-mini-bento">
   <div class="border py-2 px-3 position-relative d-inline-flex align-items-center gap-2">
-    <%= link_to url, class: 'stretched-link text-secondary link-decoration-none' do %>
+    <%= link_to url, class: 'stretched-link text-secondary link-decoration-none', data: { action: 'click->analytics#trackLink' } do %>
       <span class="fw-normal">SearchWorks</span>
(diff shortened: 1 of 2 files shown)


app/components/searchworks4/mini_mini_bento_component.html.erb
```diff
@@ -1,6 +1,6 @@
-<div class="d-flex mt-2 mb-3 justify-content-center mini-mini-bento d-none" data-controller="mini-mini-bento">
+<div class="d-flex mt-2 mb-3 justify-content-center mini-mini-bento d-none" data-controller="mini-mini-bento analytics" data-analytics-category-value="mini-mini-bento">
   <div class="border py-2 px-3 position-relative d-inline-flex align-items-center gap-2">
-    <%= link_to url, class: 'stretched-link text-secondary link-decoration-none' do %>
+    <%= link_to url, class: 'stretched-link text-secondary link-decoration-none', data: { action: 'click->analytics#trackLink' } do %>
```
jcoyne, line 3: Won't this condition exclude them from being tracked? https://github.com/sul-dlss/SearchWorks/blob/main/app/javascript/controllers/analytics_controller.js#L52

taylor-steve: No, I don't think so. That's the early exit for external links. The links we're trying to track in mini-mini bento are internal, which GA doesn't track by default. This is the pattern we use for other internal tracked links in SearchWorks and we are currently receiving data in GA.

jcoyne: Okay, I was thinking of mini-bento instead.

jcoyne (approved):

PR #6835, opened 2026-09-09
Create a Rails 8.1 style Dockerfile suitable for deployment

jcoyne:
I was asked by Julian to create a Docker image for Searchworks.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -256,6 +256,7 @@ GEM
       faraday (>= 1, < 3)
     faraday-net_http (3.4.4)
       net-http (~> 0.5)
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -671,6 +676,7 @@ GEM
     zeitwerk (2.8.3)
 
 PLATFORMS
+  aarch64-linux
```
corylown, line 679: should this be `aarch64-linux-gnu`?

corylown (approved):

Gemfile.lock
jcoyne: I think as written is correct. Bundler intentionally abstracts away the specific C standard library.

PR #6854, opened 2026-09-17
Upgrade newrelic to v10

### LINKED ISSUE #6424: New Relic v10 has breaking changes
https://github.com/newrelic/newrelic-ruby-agent/blob/dev/CHANGELOG.md

Capistrano recipes are removed, which is fine. The rest need to be checked.

jcoyne:
Fixes #6424

Should also fix Huawei browser errors in newrelic frontend code: https://app.honeybadger.io/projects/50022/faults/134327305


config/deploy/prod.rb
```diff
@@ -13,5 +13,3 @@
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, 'production'
-
-after 'deploy:updated', 'newrelic:notice_deployment'
```
jcoyne, line 17: This was already a no-op because it runs on the deployer machine, but depends on config/newrelic.yml, which was removed in 66203d53c

